PR #4, opened 2018-01-04
🛠  components: add flow and prettier

cy:


alex-wilmer:
What do you think about trying babel 7? I'm using it in the other modules and it seems to be working. not sure if it will interfere with storybook or is a bad idea for some other reason. I think it's going to get out of beta soonish

modules/components/.eslintrc.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  extends: ["prettier", "prettier/flowtype", "plugin:flowtype/recommended", "react-app"],
+  plugins: [
+    "flowtype",
+    "prettier"
+  ],
+  rules: {
+    "prettier/prettier": [1, { trailingComma: "all", singleQuote: true }],
```
Jephuff: Could you add 
```
"prettier": {
    "singleQuote": true,
    "trailingComma": "all"
  }
```
to the root package json as well?

PR #9, opened 2018-01-08
basic grid

Jephuff:


Jephuff:
For this page It might make more sense to still just use flex, but this is what I had to done to play with the grid.

PR #47, opened 2018-01-26
♻️ table theme code cleanup

alex-wilmer:
@hlminh2000 make sure I didn't break anything


modules/components/src/ThemeSwitcher/StyleProvider.js
```diff
@@ -1,40 +1,39 @@
 import React from 'react';
 
 export default class StyleProvider extends React.Component {
-
   state = {
```
hlminh2000: props.page needs to be fed into state, we'll need a constructor here

alex-wilmer: hmm not sure I understand. `props.page` isn't being used anywhere here?

hlminh2000: we could just hardcode it to be the 1, but the react-table does pass it in

hlminh2000: It could be a custom prop from ReactTable being passed in. Regardless, we need to initialize `state.page` to a number.

hlminh2000:
@alex-wilmer I pushed the change to CustomPagination for the props. 
I tried just using state `props.page` in render but ReactTable doesn't seem to be passing the change back...

PR #63, opened 2018-01-29
Integrates a treeview component and adds Beagle styling

hlminh2000:


modules/components/src/NestedTreeView/NestedTreeView.js
```diff
@@ -0,0 +1,26 @@
...
+  dataSource.map(
+    (node, i) =>
+      node.children ? (
+        <ReactTreeView
+          key={i}
+          nodeLabel={node.title}
+          defaultCollapsed={true}
+          itemClassName={node.isHeader && 'header'}
+        >
+          {node.children ? <NestedTreeView dataSource={node.children} /> : null}
```
alex-wilmer: if `node.children` is falsey this line will never be reached anyways from the check above

modules/components/src/NestedTreeView/NestedTreeView.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import ReactTreeView from './ReactTreeView';
+import './style.css';
+
+const Leaf = ({ title }) => (
```
alex-wilmer: is this component really adding much? why not inline it

alex-wilmer: also you're just renaming `children` to `title` here, probably just use children?

modules/components/src/NestedTreeView/NestedTreeView.js
hlminh2000: Good point... :P

modules/components/src/NestedTreeView/NestedTreeView.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import ReactTreeView from './ReactTreeView';
+import './style.css';
+
+const NestedTreeView = ({ dataSource }) =>
+  dataSource.map(
+    (node, i) =>
+      node.children ? (
+        <ReactTreeView
+          key={i}
```
Jephuff: Can you make the key something like `node.key` or `node.key || node.title` if we don't want to force a key field.

modules/components/src/NestedTreeView/style.css
```diff
@@ -0,0 +1,44 @@
...
+
+.tree-view_children-collapsed {
+  height: 0px;
+}
+
+.tree-view_arrow {
+  cursor: pointer;
+  margin-right: 6px;
+  display: inline-block;
+  -webkit-user-select: none;
```
Jephuff: We should have an autoprefixer so you can leave the prefixes off in any css module.

hlminh2000: I guess this is also a good reason to bring the theme css files into the build process rather than having them hanging out as static assets... I'll add an issue for adding autoprefixers, for now I'll just take out the prefixes.

modules/components/src/NestedTreeView/NestedTreeView.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import ReactTreeView from './ReactTreeView';
+import './style.css';
+
+const NestedTreeView = ({ dataSource, depth = 0 }) =>
+  dataSource.map(
+    ({ title, id, children, isHeader }, i) =>
+      children ? (
+        <ReactTreeView
+          key={id || i}
```
Jephuff: Instead of `i` could you use `title` or ``${title}-${i}``. will avoid weird bugs down the line :)

modules/components/stories/Portal.js
```diff
@@ -158,9 +158,6 @@ storiesOf('Portal', module).add('Exploration', () => (
                             .map(agg => ({
                               ...agg,
                               ...data[index].aggregations[agg.field],
-                              ...data[index].extended.find(
```
Jephuff: is this on purpose?

modules/components/src/Aggs/AggsQuery.js
```diff
@@ -8,13 +8,9 @@ export default ({ index = '', aggs = [], ...props }) =>
   ) : (
     <Query
       name={`${capitalize(index)}AggregationsQuery`}
-      variables={{
```
Jephuff: is this on purpose?

modules/components/stories/Treeview.js
```diff
@@ -0,0 +1,65 @@
...
+          },
+        ],
+      },
+    ],
+  },
+];
+
+storiesOf('Treeview', module).add('Treeview', () => (
+  <>
+    <ThemeSwitcher availableThemes={AVAILABLE_THEMES} />
```
Jephuff: Could you use the theme decorator here? It recently got merged, you can look in `SQONView.js` for an example

modules/components/stories/mappingToTreeData.js
```diff
@@ -0,0 +1,59 @@
...
+  },
+};
+const elasticMappingToDisplayTreeData = elasticMapping => {
+  const mappingKeys = Object.keys(elasticMapping);
+  return mappingKeys.map(key => {
+    const fieldProps = elasticMapping[key];
+    return {
+      title: key,
+      id: key,
+      [fieldProps.properties && 'children']: fieldProps.properties
```
Jephuff: Double check on fieldProps.properties, this should probably just be children always. 
if you need children to not exist if `fieldProps.properties` is falsy, you can conditionally spread. 
```
...(fieldProps.properties ? {children: elasticMappingToDisplayTreeData(fieldProps.properties) }) : {}
```

modules/mapping-utils/src/createConnectionResolvers.js
```diff
@@ -13,11 +13,6 @@ let createConnectionResolvers: TcreateConnectionResolvers = ({ type }) => ({
       // TODO: stitch extended mapping
       return type.mapping;
     },
-    extended: (obj, { fields }) => {
```
Jephuff: is this on purpose?

modules/mapping-utils/src/createConnectionTypeDefs.js
```diff
@@ -4,8 +4,6 @@ export default ({ type, fields = '' }) => `
   type ${type.name} {
     mapping: JSON
 
-    extended(fields: [String]): JSON
```
Jephuff: Is this on purpose?

modules/server/src/projects/spinup.js
```diff
@@ -69,14 +69,6 @@ export default ({ app }) => async (req, res) => {
         let fields = await es.search({
           index: `arranger-projects-${id}-${type.index}`,
           type: `arranger-projects-${id}-${type.index}`,
-          size: 0,
```
Jephuff: Is this on purpose?

modules/mapping-utils/src/mappingToDisplayTreeData.js
```diff
@@ -0,0 +1,59 @@
...
+    type: 'text',
+    fields: {
+      keyword: {
+        type: 'keyword',
+        ignore_above: 256,
+      },
+    },
+  },
+};
+const elasticMappingToDisplayTreeData = elasticMapping => {
```
Jephuff: This function exists twice,  can you use the same one in both spots?

modules/components/src/Aggs/AggsQuery.js
hlminh2000: This is part of the normalization commit

modules/components/src/NestedTreeView/NestedTreeView.js
hlminh2000: definitely!

modules/components/stories/Portal.js
hlminh2000: This is part of the normalization commit

modules/components/stories/mappingToTreeData.js
hlminh2000: I see, yes that makes more sense

modules/components/stories/Treeview.js
```diff
@@ -0,0 +1,80 @@
...
+          },
+        ],
+      },
+    ],
+  },
+];
+
+storiesOf('Treeview', module)
+  .addDecorator(themeDecorator)
+  .addDecorator(story => (
```
Jephuff: oh sorry, you don't need this part, just the `.addDecorator(themeDecorator)` part.

hlminh2000: gotcha!

modules/mapping-utils/src/mappingToDisplayTreeData.js
hlminh2000: I added it here so it can be shared through the `mapping-utils` module, but I think we need to publish a new version of the module to have access to it in `components.

modules/mapping-utils/src/createConnectionResolvers.js
hlminh2000: This is part of the normalization commit

modules/mapping-utils/src/createConnectionTypeDefs.js
hlminh2000: This is part of the normalization commit

modules/server/src/projects/spinup.js
hlminh2000: This is part of the normalization commit

PR #68, opened 2018-01-30
:sparkles: Collapsible TermAggs

hlminh2000:
Adds a layer above the TermAggs to toggle collapsing, with style


hlminh2000:
Should have just made a new branch instead of recycling the old branch... will just squash

PR #97, opened 2018-02-01
:sparkles: added delete type

jgnieuwhof:
for #36 , also fixed an issue where following a project link wasn't setting it as the activeProject

### DIFF
--- a/modules/components/src/Admin/ProjectsTable.js
+++ b/modules/components/src/Admin/ProjectsTable.js
@@ -4,6 +4,7 @@ import DataTable from '../DataTable';
 
 export default ({
   newProjectName,
+  setActiveProject,
   setNewProjectName,
   addProject,
   projectsTotal,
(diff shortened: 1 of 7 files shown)


modules/components/src/Admin/ProjectsTable.js
```diff
@@ -25,7 +26,14 @@ export default ({
     allowTSVExport={false}
     customTypes={{
       entity: props => {
-        return <Link to={`/projects/${props.value}`}>{props.value}</Link>;
+        return (
+          <Link
+            onClick={() => setActiveProject({ activeProject: props.value })}
+            to={`/projects/${props.value}`}
```
jgnieuwhof: @alex-wilmer - I thought there should be some better way of tying a route change to state, but wasn't sure, is this how you'd do it?

Jephuff: You can derive the state from the url so you don't have to maintain it.
That will likely require a bit of refactoring of the routes which is probably out of scope for this ticket.

alex-wilmer: yeah I had `activeProject` before I added router at all. we can eventually just use route params for that, agreed that its out of scope here

PR #99, opened 2018-02-01
💄styles the Show More button with some minor refactoring

hlminh2000:


modules/components/src/Aggs/TermAgg.js
```diff
@@ -10,6 +10,7 @@ const TermAggs = ({
   handleValueClick = () => {},
   isActive = () => {},
   Content = 'div',
+  defaultMaxEntries = 5,
```
alex-wilmer: maybe this props should be called `maxTerms`. the `5` is the default part, the `prop` itself isn't default

alex-wilmer:
@hlminh2000 added a story to test it -- looks good

PR #105, opened 2018-02-02
✨ refactor state management

Jephuff:


modules/components/src/Arranger/Arranger.js
```diff
@@ -111,17 +128,7 @@ class Arranger extends React.Component {
...
-                      projectId,
-                      index,
-                      sqon,
-                      streamData: options =>
-                        streamData({ ...options, type: index }),
-                      fetchData,
-                      onSQONChange: this.onSQONChange,
-                    })
-                  : 'default'}
+                {render ? render({}) : 'default'}
```
alex-wilmer: what's this part?

alex-wilmer: oh I think I get it.. looks weird though

Jephuff: Just the render callback but passing it no values. The default was because I was initially going to make a default portal, but I think I should remove that. 

We could do something like react-router and accept `render`, `Component` or `children` and error if multiple are passed.

PR #112, opened 2018-02-05
:sparkles: a range agg that supports unit conversions

jgnieuwhof:


modules/components/src/Admin/Dashboard.js
```diff
@@ -600,7 +602,90 @@ class Dashboard extends React.Component {
...
-                                  {typeof val === 'string' && (
+                                  {key === 'unit' ? (
+                                    <State
+                                      initial={{
+                                        val,
+                                        measure: val
+                                          ? convert().describe(val).measure
+                                          : '',
+                                      }}
+                                      val={val}
```
alex-wilmer: this isn't being used, right?

PR #115, opened 2018-02-05
✨ text search / filtering

jgnieuwhof:


modules/components/src/Arranger/Table.js
```diff
@@ -16,6 +18,32 @@ const Table = ({ sqon, projectId, index, streamData, fetchData, setSQON }) => {
             onSelectionChange={console.log('selection changed')}
             streamData={streamData(index)}
             fetchData={fetchData(projectId)}
+            onColumnsChange={columnState.toggle}
+            onFilterChange={value =>
```
jgnieuwhof: this seems out of place, but wasn't sure where to put it

jgnieuwhof:
also, a note on the searching. I went with a `phrase_prefix` search in ES. this means that it will match incomplete words as a user types. there are two issues with this:
1) keyword fields will only work for a full match (won't respond to just prefix)
2) text fields need their `fielddata` prop turned on
3) it's not fuzzy (but I don't _think_ there's a way to do both prefix and fuzzy)

spoke with junjun about it, he had a suggestion to use `copy_to` for each field that we want searchable, then search on the aggregate virtual field... we'd have to talk about a mapping convention here so that the UI and / or middleware knows what to do.

for now this works, but we'll need to discuss & resolve the bigger picture mapping stuff at some point

modules/components/src/DataTable/TableToolbar/TableToolbar.js
```diff
@@ -16,6 +17,7 @@ const TableToolbar = ({
   onColumnsChange,
   onSQONChange,
   onFilterChange = () => {},
+  donFilterChange = debounce(onFilterChange, 300),
```
Jephuff: This is going to recall debounce on every render, might be better to use recomposes [withPropsOnChange](https://github.com/acdlite/recompose/blob/master/docs/API.md#withpropsonchange) instead

PR #118, opened 2018-02-05
🐛 don't overwrite state when starting server

Jephuff:


Jephuff:
closing and replacing with https://github.com/overture-stack/arranger/pull/124

PR #119, opened 2018-02-05
✨ Feat/advanced facet view

hlminh2000:


modules/components/src/AdvancedFacetView/FacetViewNode.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import { esToAggTypeMap } from '@arranger/mapping-utils';
+import AggWrapper from './Facets';
+
+const esTypeToAggType = esType => esToAggTypeMap[esType];
+
+const FacetViewNode = ({ title, id, children, path, type, aggregations }) => {
+  const esType = type;
```
alex-wilmer: why make this copy?

modules/components/src/AdvancedFacetView/FacetView.js
```diff
@@ -0,0 +1,108 @@
...
+    isAnimating: false,
+  };
+  componentWillReceiveProps({
+    selectedMapping,
+    path: selectedPath,
+    aggregations,
+    disPlayTreeData,
+  }) {
+    if (selectedPath) {
+      this.scrollToPath(selectedPath);
```
Jephuff: This probably shouldn't be on props change, we only want to scroll on user interaction so it can go directly in the onClick handler.

modules/components/src/AdvancedFacetView/FacetView.js
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }
+
+  scrollToPath = path => {
+    const targetElementId = serializeToDomId(path);
+    const targetElement = $(this.root).find(`#${targetElementId}`);
+    if (targetElement) {
+      this.setState({ isAnimating: true });
+      $(this.root)
```
Jephuff: No need for jquery, something like this should be good
```
const el = document.querySelector(`#${targetElementId}`);
if(el) {
  el.scrollIntoView();
}
```

modules/components/src/AdvancedFacetView/FacetView.js
```diff
@@ -0,0 +1,108 @@
...
+        );
+    }
+  };
+
+  stopScrollAnimation = () => {
+    $(this.root).stop();
+    this.setState({ isAnimating: false });
+  };
+
+  onScroll = debounce(e => {
```
Jephuff: What do you think of using [react-waypoint ](https://github.com/brigade/react-waypoint) instead of listening on scroll? might simplify it a bit.

hlminh2000: I will opt for the refactoring we talked about yesterday to remove the extra feature so I don't we'll need to worry about this :)

modules/components/src/AdvancedFacetView/index.js
```diff
@@ -0,0 +1,278 @@
...
+    };
+  };
+  return rawDisplayData.map(replaceNodeDisplayName);
+};
+
+const filterOutNonValue = ({ aggregations, displayTreeData }) => {
+  const aggregationsWithValue = keys(aggregations).reduce(
+    (a, key) => ({
+      ...a,
+      ...(() => {
```
Jephuff: This closure shouldn't be required, can move that logic into the reduce scope.

modules/components/src/AdvancedFacetView/index.js
```diff
@@ -0,0 +1,278 @@
...
+    const displayTreeDataWithValue = applyFilterToDisplayNodeCollection(
+      displayTreeData,
+    );
+    return { displayTreeDataWithValue, aggregationsWithValue };
+  } else {
+    return { aggregationsWithValue };
+  }
+};
+
+const emptySQON = {
```
Jephuff: Should be able to remove this, and we can handle null sqon values

modules/components/src/AdvancedFacetView/index.js
```diff
@@ -0,0 +1,278 @@
...
+};
+
+const serializeToDomId = path => path.split('.').join('__');
+const serializeDomIdToPath = path => path.split('.').join('__');
+
+export default class AdvancedFacetView extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      localSQON: props.sqon || emptySQON,
```
Jephuff: Do you still need localSQON? looks like the value passed in should be accurate

modules/components/stories/Aggs.js
```diff
@@ -109,27 +109,30 @@ storiesOf('Aggs', module)
...
-                            ],
-                          },
-                          sqon,
-                        ),
-                      })
-                    }
-                  />
-                )}
+                Content={({ content, ...props }) => {
+                  console.log(props);
```
Jephuff: Could you revert this change? (unless it's not just debugging code)

modules/components/src/NestedTreeView/NestedTreeView.js
```diff
@@ -2,26 +2,44 @@ import React from 'react';
...
+  selectedPath = '',
+}) =>
+  dataSource.map(({ title, id, children, path }, i) => {
+    const selectedPathArray = selectedPath?.split('.') || [];
+    return children ? (
+      <ReactTreeView
+        key={path}
+        nodeLabel={title}
+        defaultCollapsed={true}
+        itemClassName={depth == 0 && 'header'}
```
Jephuff: triple equals

modules/components/src/NestedTreeView/NestedTreeView.js
```diff
@@ -2,26 +2,44 @@ import React from 'react';
...
+          dataSource={children}
+          depth={depth + 1}
+        />
+      </ReactTreeView>
+    ) : (
+      <div
+        onClick={() => onLeafSelect(id || title)}
+        key={path}
+        className={`tree-view_children leaf
+          ${depth == 0 ? 'header' : ''}
```
Jephuff: triple equals

modules/components/src/NestedTreeView/NestedTreeView.js
```diff
@@ -2,26 +2,44 @@ import React from 'react';
...
+        nodeLabel={title}
+        defaultCollapsed={true}
+        itemClassName={depth == 0 && 'header'}
+      >
+        <NestedTreeView
+          onLeafSelect={selectedPath => {
+            onLeafSelect(`${id || title}.${selectedPath}`);
+          }}
+          selectedPath={selectedPathArray
+            .slice(1, selectedPathArray.length)
```
Jephuff: You can just do `.slice(1)` that will take the rest of the array.

modules/components/src/NestedTreeView/NestedTreeView.js
```diff
@@ -2,26 +2,44 @@ import React from 'react';
...
-      ),
-  );
+const NestedTreeView = ({
+  dataSource,
+  depth = 0,
+  onLeafSelect = () => {},
+  selectedPath = '',
+}) =>
+  dataSource.map(({ title, id, children, path }, i) => {
+    const selectedPathArray = selectedPath?.split('.') || [];
```
Jephuff: It won't make a different right now, but  this will result in `['']` if selectedPath is an empty string. Might cause some unexpected behaviour down the line so you should filter them out. `(selectedPath?.split('.') || []).filter(Boolean)`

modules/components/src/AdvancedFacetView/FacetView.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import FacetViewNode from './FacetViewNode';
+import $ from 'jquery';
+import { debounce } from 'lodash';
+
+const serializeToDomId = path => path.split('.').join('__');
+const serializeDomIdToPath = path => path.split('.').join('__');
```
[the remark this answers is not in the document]
hlminh2000: Refactored a little bit for better separation of concern, so now serializeToDomId is used :)

modules/components/src/AdvancedFacetView/FacetView.js
```diff
@@ -0,0 +1,36 @@
...
+export default class FacetView extends React.Component {
+  render() {
+    const {
+      aggregations,
+      displayTreeData,
+      onValueChange,
+      sqon = {},
+    } = this.props;
+    return (
+      <div className="facetView" ref={el => (this.root = el)}>
```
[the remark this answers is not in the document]
hlminh2000: now properly used within the component for scrolling :)

modules/components/src/AdvancedFacetView/LiveAdvancedFacetView.js
```diff
@@ -0,0 +1,142 @@
...
+};
+
+export default class LiveAdvancedFacetView extends React.Component {
+  constructor(props) {
+    super(props);
+    const { sqon } = props;
+    this.state = {
+      mapping: {},
+      extended: {},
+      sqon: sqon || {
```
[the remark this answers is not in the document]
hlminh2000: tested out null sqon and it was fine, so I'm just setting default to null to prevent `undefined`

modules/components/src/AdvancedFacetView/FacetView.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import FacetViewNode from './FacetViewNode';
+import $ from 'jquery';
```
Jephuff: jQuery isn't used any more so you can remove it :D

modules/components/src/AdvancedFacetView/LiveAdvancedFacetView.js
```diff
@@ -0,0 +1,141 @@
...
+const fetchMapping = async fetchConfig =>
+  fetchGraphqlQuery({
+    query: `{
+      ${fetchConfig.ES_INDEX} {
+        mapping,
+      }
+    }`,
+    ...fetchConfig,
+  }).then(data => {
+    return Promise.resolve(data[fetchConfig.ES_INDEX]);
```
Jephuff: A couple more promises you don't need (one below as well)

modules/components/src/AdvancedFacetView/LiveAdvancedFacetView.js
```diff
@@ -0,0 +1,141 @@
...
+      fetchAggregationDataFromExtendedMapping({
+        extended,
+        ...fetchConfig,
+      }).then(({ aggregations }) =>
+        this.setState({ mapping, extended, aggregations }),
+      ),
+    );
+  }
+  componentWillReceiveProps({ sqon }) {
+    if (sqon) {
```
Jephuff: It's possible an update to the sqon would make it `null` which this would block, so you should check for if the value has changed instead (you can use the lodash funtion `isEqual`)

PR #127, opened 2018-02-09
🐛fixes how the nextSQON is generated for the values

hlminh2000:
It was not set up to consider every kind of `op`, particularly was breaking for range types


Jephuff:
I think we fixed the problem, so this PR could be closed?

alex-wilmer:
reopen if needed

PR #131, opened 2018-02-13
🎨simplified some apis to help with other projects (capsid ;)

jgnieuwhof:


Jephuff:
From our discussion, just changing it so everything takes 1 handleChange callback and exposes a getNextSQON utility function

jgnieuwhof:
@Jephuff - good call on the big nope to multiple handlers, should be resolved now

PR #132, opened 2018-02-13
✨ accept more props and rename selection prop

Jephuff:


Jephuff:
Good point. It's "selected table row ids" so we could make it something like `selectedTableRows`?

alex-wilmer:
yeah that's totally fine

Jephuff:
@alex-wilmer updated 👍

PR #145, opened 2018-02-14
🐛handles null aggregation in advanced facet view for demo

hlminh2000:
This is just to temporarily handle null case for the demo. We need to look into the underlying problem later


modules/components/stories/AdvancedFacetView.js
```diff
@@ -9,12 +9,22 @@ import { replaceSQON, toggleSQON } from '../src/SQONView/utils';
...
 
-const PROJECT_ID = 'testing1';
-const ES_INDEX = 'testing1';
+const PROJECT_ID = 'feb-14';
+const ES_INDEX = 'file';
 const API_HOST = 'http://localhost:5050';
 const ES_HOST = 'http://localhost:9200';
 // const ES_HOST = 'http://142.1.177.54:9200';
 
+// REACT_APP_ES_HOST=http://es1.kfdrc.org:9200
```
Jephuff: Can you remove the comments?

PR #153, opened 2018-02-15
🐛 match dotfield with __ for aggs

alex-wilmer:


alex-wilmer:
image

alex-wilmer:
instead of a bunch of Unnamed Field

PR #155, opened 2018-02-15
🐛 fix saveTSV row joining

Jephuff:


modules/components/src/DataTable/TableToolbar/saveTSV.js
```diff
@@ -59,22 +59,24 @@ function getRows(args) {
     entities = [],
   } = args;
   if (pathIndex >= paths.length - 1) {
-    return columns.map(column => {
-      const entity = entities
-        .slice()
-        .reverse()
-        .find(entity => column.field.indexOf(entity.field) === 0);
+    return [
+      columns.map(column => {
```
alex-wilmer: returning to nowhere?

alex-wilmer: or sorry, 2d array? i guess that makes sense 💤

Jephuff: yep, normalizing the return so I cal always call flatten on the return value of this function :)

PR #156, opened 2018-02-15
🐛forces value to be array for Aggregations type in advanced facet view 

hlminh2000:


modules/components/src/AdvancedFacetView/index.js
```diff
@@ -124,7 +124,7 @@ export default class AdvancedFacetView extends React.Component {
                     op: 'in',
                     content: {
                       field: path,
-                      value: value,
+                      value: typeof value === 'string' ? [value] : value,
```
alex-wilmer: not sure I get this? numbers should not be forced into arrays?

hlminh2000: I guess we should check for not arrays

modules/components/src/AdvancedFacetView/index.js
```diff
@@ -124,7 +124,7 @@ export default class AdvancedFacetView extends React.Component {
                     op: 'in',
                     content: {
                       field: path,
-                      value: typeof value === 'string' ? [value] : value,
+                      value: !Array.isArray(value) ? [value] : value,
```
alex-wilmer: for some reason I'm tempted to suggest `[].concat(value)` which would work.. but that now I think that's a dumb js thing

hlminh2000: hum... yeah that's interesting but I think I would be confused seeing that

Jephuff: one difference is you can do `[].concat(value || [])` which means if `value === null` the result is `[]` instead of `[null]`
Positive? Negative? you decide!

alex-wilmer: or filter(Boolean)

alex-wilmer: /shrug

PR #158, opened 2018-02-15
🎨 separated text filter for external use

jgnieuwhof:
In using arranger for capsid we realized that we could use the arranger text filter (which has knowledge of how to modify sqon based on the text filter contents), but didn't need the suite of things that comes along with `TableToolbar`. This PR decouples `TextFilter` from `TableToolbar`


alex-wilmer:
can you

- add a description to this PR with your intended improvement
- reconsider the location of TextFilter.js. is it really coupled to the table toolbar?
- rebase master + add a unit test for generateNextSQON

thanks!

jgnieuwhof:
@alex-wilmer - comments above should be addressed, plz let me know if there's anything else!

jgnieuwhof:
@alex-wilmer - TextFilter is a top-level entity now, let me know if there's anything else :)

PR #165, opened 2018-02-16
✨ updates the aggregations given the new sqon

hlminh2000:
This only addresses updating the advanced facet view (afv), the same thing needs to be done in the actual portal for Kids First as well

### DIFF
--- a/modules/components/src/AdvancedFacetView/LiveAdvancedFacetView.js
+++ b/modules/components/src/AdvancedFacetView/LiveAdvancedFacetView.js
@@ -39,10 +39,9 @@ const fetchAggregationData = async ({
   extended,
   PROJECT_ID,
   ES_INDEX,
-  API_HOST,
   ES_HOST,
 }) => {
-  const fetchConfig = { PROJECT_ID, ES_INDEX, API_HOST, ES_HOST };
+  const fetchConfig = { PROJECT_ID, ES_INDEX, ES_HOST };
   const serializeToGraphQl = aggName => aggName.split('.').join('__');
   const serializeToPath = aggName => aggName.split('__').join('.');
   const allAggsNames = extended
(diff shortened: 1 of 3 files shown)


modules/components/src/AdvancedFacetView/LiveAdvancedFacetView.js
```diff
@@ -125,7 +132,15 @@ export default class LiveAdvancedFacetView extends React.Component {
...
     const { onSqonChange = () => {} } = this.props;
-    this.setState({ sqon }, () => onSqonChange({ sqon }));
+    fetchAggregationData({
+      ...this.props,
+      extended: this.state.extended.filter(
+        e => e.type !== 'object' && e.type !== 'nested',
+      ),
+      sqon,
+    }).then(({ aggregations }) =>
+      this.setState({ sqon, aggregations }, () => onSqonChange({ sqon })),
```
hlminh2000: This means all the values that are irrelevant for a given `sqon` will be filtered out as it will not be included in the aggregations returned. This feels like a nice feature in my opinion but good to note.

hlminh2000: So turned out, there is an argument for this in the `aggregations` query, but it's not working.
Tracking this as a separate issue here: https://github.com/overture-stack/arranger/issues/168

modules/components/src/AdvancedFacetView/LiveAdvancedFacetView.js
```diff
@@ -124,8 +124,22 @@ export default class LiveAdvancedFacetView extends React.Component {
     }
   }
   onSqonFieldChange = ({ sqon }) => {
-    const { onSqonChange = () => {} } = this.props;
-    this.setState({ sqon }, () => onSqonChange({ sqon }));
+    const {
+      onSqonChange = () => {},
+      projectId: PROJECT_ID,
```
Jephuff: can  you update the code to use `projectId` instead of renaming it here? (same for PROJECT_ID and ES_INDEX in the rest of the file)

PR #166, opened 2018-02-16
✨ AFV facet filter box

hlminh2000:


phuongmy:
Nice user experience!
A few comments:
1- Provide the ability to select value in dropdown with keyboard (not only mouse click)
2- I suggest that when the user selects the field in the dropdown, it displays the full label in the text box search? Right now, the focus in the right panel is good but it leaves in the texbox the few letters I typed, like "sta"
screen shot 2018-02-21 at 11 46 44 am
3- The font of the text in the dropdown seems big compared to the other components? It is possible to have the text smaller so it could potentially display more values in the dropdown? (currently 3 by default)

rosibaj:
feedback from slack:
1 - facet tree panel font needs to be updated 
2 - show 6 items, and the show more below that, rather than the show more taking up the 6th position
3 - Make expandable top-level category facet text clickable as well ( for example, Archive and both expand and be scrolled to.  Users should click the expand to expand, and the text to scroll)
4 - make sure ability to read multiple indices is correct (for example, Access is showing as a category, when really, it should be a facet under file)

alex-wilmer:
@rosibaj this was already merged, can you add a new issue?

hlminh2000:
@alex-wilmer @rosibaj I have added those points to the pull request over here, which is linked to the same issue: https://github.com/overture-stack/arranger/pull/179

PR #169, opened 2018-02-17
fixing export locations / test successful importing

alex-wilmer:
I'm kind of happy how this experiment turned out? I wanted to change the way arranger/components were being exported and I knew it was going to break stuff. 

this PR in its current state doesn't cover that entirely but I think this could be used as a decent model for integration tests moving forward.

### DIFF
--- a/modules/components/src/Admin/Dashboard.js
+++ b/modules/components/src/Admin/Dashboard.js
@@ -2,7 +2,7 @@ import React from 'react';
 import Component from 'react-component-component';
 import { debounce, startCase } from 'lodash';
 import io from 'socket.io-client';
-import { BrowserRouter, Route, Link, Redirect } from 'react-router-dom';
+import { Route, Link, Redirect } from 'react-router-dom';
 import convert from 'convert-units';
 import GraphiQL from 'graphiql';
 import 'graphiql/graphiql.css';
@@ -363,7 +363,7 @@ class Dashboard extends React.Component {
   render() {
     let headerHeight = 38;
     return (
-      <BrowserRouter>
+      <>
         <div
           className="app"
           css={`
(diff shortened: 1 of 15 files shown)


modules/components/src/Admin/Dashboard.js
```diff
@@ -519,7 +524,7 @@ class Dashboard extends React.Component {
           />
           <Route
             exact
-            path="/projects/:projectId/:index"
+            path={basePath + '/:projectId/:index'}
```
Jephuff: Can you avoid concating basePath and passing it around by making a router for the dashboard and setting `basename` on it?

alex-wilmer: I didn't think you could nest routers. I'll try it

alex-wilmer: this does seem to work. is this what you mean?

```
  render() {
    return (
      <BrowserRouter>
      <div>
        <Link to="/admin">go</Link>
        <Route path="/admin" render={({ match }) =>
          <BrowserRouter basename={match.url}>
            <div>
              <Route path="/" render={() => <div>sup</div>} />
            </div>
          </BrowserRouter>
        } />
      </div>
      </BrowserRouter>
    );
  }
```

alex-wilmer: this seems far better than what the official docs suggest for nested routing

Jephuff: yeah thats exactly what I was thinking :) it works? yay! 
whats the official method?

alex-wilmer: they just don't go that far to say nest routers. they're probably being conservative considering you rely on nesting context 

https://reacttraining.com/react-router/core/guides/philosophy/nested-routes

PR #172, opened 2018-02-18
next -> master

alex-wilmer:
We need to have a different branch than master for non-standard package releases (at least with lerna since it creates a commit + tag) but eventually we'll want this for actual "next" releases, so getting the ball rolling


alex-wilmer:
Also I never had a need to learn semver, so please excuse the, probably horribly incorrect random versioning 😊

Jephuff:
Is there a problem with tags on master for beta/alpha?
I also don't know that we will need to have a concept of `next` release for arranger, I think we only need that if we want to batch release features but we can release as things are ready and the user would update at their discretion.

Just my 2 cents! :) 
 
what does `v0.0.15-0.0.16.2` mean? :D

alex-wilmer:
well the conceptual problem is that if we want concurrent versions then those versions should be reflected in the package.json's etc, right? how would you go about having a 1.x and a 2.x release all being done through master?

> what does v0.0.15-0.0.16.2 mean? :D

I dunno, that was the default lerna suggested when i selected prerelease

Jephuff:
hmm good point. I was hoping we could avoid most long term support. But the server would definitely need security patches.

PR #175, opened 2018-02-20
build arranger api in Jenkins pipeline

jonathanganesh:


alex-wilmer:
@jonathanganesh can we please move this here https://github.com/kids-first/kf-arranger

PR #179, opened 2018-02-21
✨ improvements to the search box

hlminh2000:
Addressing the following points:

from @phuongmy 
- [x] 1- Provide the ability to select value in dropdown with keyboard (not only mouse click)
- [x] 2- I suggest that when the user selects the field in the dropdown, it displays the full label in the text box search? Right now, the focus in the right panel is good but it leaves in the texbox the few letters I typed, like "sta"
screen shot 2018-02-21 at 11 46 44 am

- [x] 3- The font of the text in the dropdown seems big compared to the other components? It is possible to have the text smaller so it could potentially display more values in the dropdown? (currently 3 by default)

from @kcullion 
- [x] 1- Highlight the matched portion of the search term, similar to: https://zpl.io/2Zxjy9J

from @rosibaj 
- [x] fix overflowing slider labels 
- [x] treeview font
- [x] Stacking styling for term aggs
screen shot 2018-02-23 at 9 28 37 am
- [x] Filter out IDs
- [x] show 6 items, and the show more below that, rather than the show more taking up the 6th position
- [x] Make expandable top-level category facet text clickable as well ( for example, Archive and both expand and be scrolled to. Users should click the expand to expand, and the text to scroll)
- [x] facet tree panel font needs to be updated


phuongmy:
Looks great!!

modules/components/src/AdvancedFacetView/SearchBox.js
```diff
@@ -1,32 +1,98 @@
 import React from 'react';
-import State from '../State';
-import TextInput from '../Input';
 import SearchIcon from 'react-icons/lib/fa/search';
+import $ from 'jquery';
```
alex-wilmer: is this being used anywhere?

modules/components/public/themeStyles/beagle/beagle.css
```diff
@@ -558,26 +558,33 @@ div {
...
-  height: 130px;
+  height: 140px;
   overflow-y: scroll;
+  border-radius: 5px;
+  box-shadow: 0 0 7.8px 0.2px #bbbbbb;
 }
 
 .advancedFacetViewWrapper .resultList.shown {
-  border: solid 1px #e7e8ed;
+  /* border: solid 1px #e7e8ed; */
```
alex-wilmer: can this be removed?

modules/components/public/themeStyles/beagle/beagle.css
```diff
@@ -558,26 +558,33 @@ div {
...
 }
 
 .advancedFacetViewWrapper .resultItem {
   padding: 10px;
   cursor: pointer;
-  border-bottom: solid 1px #e7e8ed;
+  font-size: 0.8em;
+  /* border-bottom: solid 1px #e7e8ed; */
```
alex-wilmer: same

modules/components/public/themeStyles/beagle/beagle.css
```diff
@@ -558,26 +558,33 @@ div {
...
 }
 
-.advancedFacetViewWrapper .resultItem:hover {
+.advancedFacetViewWrapper .resultItem.highlighted {
   background: #e7e8ed;
 }
 
 .advancedFacetViewWrapper .resultItem .title {
-  font-weight: bold;
+  /* font-weight: bold; */
```
alex-wilmer: same

modules/components/src/AdvancedFacetView/SearchBox.js
```diff
@@ -1,32 +1,98 @@
...
-    window.removeEventListener('click', this.onWindowClick);
-  }
-  render() {
+
+  handleKeyPress = e => {
+    const filteredList = this.getFilteredList();
+    const { highlightedField } = this.state;
+    const { extendedMapping, onFieldSelect = () => {} } = this.props;
+    const highlightedAgg = extendedMapping.find(
+      ({ field, displayName }) => field === highlightedField,
```
alex-wilmer: `displayName` not being used?

modules/components/src/AdvancedFacetView/SearchBox.js
```diff
@@ -1,32 +1,98 @@
...
+            behavior: 'smooth',
+            block: 'nearest',
+          }),
+      );
+    }
+
+    // down
+    if (e.keyCode === 40) {
+      const newHighlightedField =
+        filteredList?.[
```
alex-wilmer: is the only difference here `+1` vs `-1`. maybe make a helper function?

modules/components/src/AdvancedFacetView/SearchBox.js
```diff
@@ -50,20 +146,44 @@ export default class extends React.Component {
...
                 onClick={() => {
+                  this.setState({ currentValue: displayName });
                   onFieldSelect(field);
                 }}
               >
-                <span className="title">{displayName}</span>
-                <span className="field">{field}</span>
+                <span
+                  className="title"
+                  dangerouslySetInnerHTML={{
```
alex-wilmer: I'm 99% sure we shouldn't be doing this. Why not just pass a regular react component?

modules/components/src/AdvancedFacetView/Facets/TermAggregation.js
```diff
@@ -2,24 +2,31 @@ import React from 'react';
...
-      return inCurrentSQON({
-        value: d.value,
-        dotField: d.field,
-        currentSQON: sqon,
-      });
-    }}
-  />
-);
+export default ({ aggType, aggProps, title, onValueChange, sqon, path }) => {
+  const buckets = aggProps ? aggProps.buckets || [] : [];
```
alex-wilmer: can this be 

```
aggProps?.buckets || []
````

Comment (hlminh2000) [reply modules/components/src/AdvancedFacetView/Facets/TermAggregation.js]: oh yes... don't why I didn't think of that :P

File: modules/components/src/AdvancedFacetView/Facets/TermAggregation.js [hunk 9]
```diff
@@ -2,24 +2,31 @@ import React from 'react';
...
+  const hasLongValues = true && buckets.find(({ key }) => key.length > 50);
```
alex-wilmer: what `true` here? placeholder for something? can `50` come from props?

modules/components/src/AdvancedFacetView/Facets/TermAggregation.js
```diff
@@ -2,24 +2,31 @@ import React from 'react';
...
+    <div className={hasLongValues ? 'hasLongValues' : 'hasShortValues'}>
+      <TermAggs
+        key={title}
+        collapsible={false}
+        handleValueClick={({ bucket }) => {
+          onValueChange({
+            value: bucket.key,
+          });
+        }}
+        maxTerms={6}
```
alex-wilmer: same here, can this come from props?

modules/components/src/AdvancedFacetView/Facets/TermAggregation.js
hlminh2000: just to have it evaluated to a boolean. can be written as buckets.find(({ key }) => key.length > 50) != undefined

modules/components/src/AdvancedFacetView/Facets/TermAggregation.js
```diff
@@ -2,24 +2,41 @@ import React from 'react';
...
+  title,
+  onValueChange,
+  sqon,
+  path,
+  maxTerms = 6,
+  stackingLengthLimit = 50,
+}) => {
+  const buckets = aggProps ? aggProps.buckets || [] : [];
+  const hasLongValues =
+    buckets.find(({ key }) => key.length > stackingLengthLimit) !== undefined;
```
alex-wilmer: you don't need the undefined check

```
[1, 2, 3].find(x => x === 5)
// undefined
```

modules/components/src/AdvancedFacetView/LiveAdvancedFacetView.js
```diff
@@ -81,6 +81,39 @@ const fetchAggregationData = async ({ sqon, extended, projectId, index }) => {
   }));
 };
 
+const removeIds = ({ mapping, extended, parentField = null }) => {
```
alex-wilmer: this seems like a useful function to add to `mapping-utils`. can be out of scope for this, but if you can write a quick issue for that, i'd appreciate it

modules/components/src/AdvancedFacetView/Facets/TermAggregation.js
hlminh2000: I should print out a list of array HOF :P

modules/components/src/AdvancedFacetView/SearchBox.js
```diff
@@ -38,6 +105,46 @@ export default class extends React.Component {
             (currentValue?.length ? currentValue : null)?.toLowerCase(),
           ) > -1,
     );
+  };
+
+  getHighlightedHtmlTemplate = displayName => {
```
alex-wilmer: can this be a separate component? it's not specific to afv is it?

modules/components/src/AdvancedFacetView/SearchBox.js
```diff
@@ -50,20 +157,41 @@ export default class extends React.Component {
...
+                }`}
                 onClick={() => {
+                  this.setState({ currentValue: displayName });
                   onFieldSelect(field);
                 }}
               >
-                <span className="title">{displayName}</span>
-                <span className="field">{field}</span>
+                <span className="title">
+                  {this.getHighlightedHtmlTemplate(displayName)}
```
alex-wilmer: re above: seems more reasonable to have a component here than a class method

modules/components/stories/AdvancedFacetView.js
```diff
@@ -9,7 +9,7 @@ import { replaceSQON, toggleSQON } from '../src/SQONView/utils';
 import LiveAdvancedFacetView from '../src/AdvancedFacetView/LiveAdvancedFacetView';
 import { action } from '@storybook/addon-actions';
 
-const PROJECT_ID = 'feb-14';
+const PROJECT_ID = 'feb-23';
```
alex-wilmer: this is going to keep changing. env var?

modules/mapping-utils/src/flattenMapping.js
```diff
@@ -9,7 +9,7 @@ let flattenMapping = (properties, parent = '') => {
         !data.properties
           ? {
               field: joinWith()(parent) + field,
-              type: data.type,
+              type: field.indexOf('id') > -1 ? 'id' : data.type,
```
alex-wilmer: i think `.includes` is nicer than `indexOf` since it returns a boolean

modules/components/src/AdvancedFacetView/Facets/TermAggregation.js
alex-wilmer: here you go https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array

modules/components/src/AdvancedFacetView/LiveAdvancedFacetView.js
```diff
@@ -101,16 +135,16 @@ export default class LiveAdvancedFacetView extends React.Component {
     ]).then(([{ mapping }, { extended }]) =>
       fetchAggregationData({
         extended: extended.filter(
-          e => e.type !== 'object' && e.type !== 'nested',
+          e => !this.blackListedAggTypes.find(type => type === e.type),
```
alex-wilmer: comment here as to what this is doing would be fantastic

PR #180, opened 2018-02-22
🔥🐛removes stringify-object dependency

hlminh2000:
Needs to remove this dependency as integration into KF portal fails to minify during build


modules/components/src/Aggs/AggsQuery.js
```diff
@@ -56,3 +58,9 @@ export default ({ index = '', aggs = [], sqon = null, ...props }) =>
       {...props}
     />
   );
+
```
Jephuff: Could you remove this comment? 
Then this looks good 👍

hlminh2000: ah oops, sure thanks!

PR #184, opened 2018-02-26
make use of "active" field on projects

alex-wilmer:
- Add general endpoint for modifying project fields
- if ES_HOST is provided as env, find active projects and "spin them up"
- default maximum number of "live" projects (only so many event listeners can be active by default)
- [😢] add integration test (i tried)


alex-wilmer:
will add test after https://github.com/overture-stack/arranger/pull/205 gets merged

Jephuff:
something weird is going on with the package-locks some are deleted?

alex-wilmer:
okay I checked out the master version of lock files for packages i did not edit the package.json

alex-wilmer:
i think i did a bad `lerna add` somewhere long the way

PR #194, opened 2018-03-01
✨ add mock option

Jephuff:


modules/server/src/download/index.js
```diff
@@ -9,9 +9,10 @@ import getAllData from '../utils/getAllData';
 import dataToTSV from '../utils/dataToTSV';
 
 export default function({ projectId }) {
-  function makeTSV(args) {
+  function makeTSV(args, mock) {
```
alex-wilmer: What are "args" here? the actual object data?

Jephuff: yep, so if you make a request thats body contains `files: [file1]` args here would be file1.

Jephuff: I could add mock to the file options, that would allow individual mocking as well.

PR #195, opened 2018-03-01
✨ allow defining empty cell value

Jephuff:


alex-wilmer:
how hard is it to write a test for this?

alex-wilmer:
input object === output tsv

Jephuff:
not hard at all. I can make another PR today with some tests 👍

PR #201, opened 2018-03-05
✨ add withQuery

Jephuff:


modules/components/src/Query.js
```diff
@@ -34,7 +45,7 @@ export default class extends Component {
           loading: false,
         });
       } catch (error) {
-        this.setState({ error: error.message, loading: false });
+        this.setState({ data: null, error: error.message, loading: false });
```
alex-wilmer: no qualms, just making sure it makes sense to clear the last data if an error occurs. maybe this can be a flag, though I suspect this is probably a smart default

Jephuff: right now I don't think it really matters, if there is an error it renders that instead. I think that should be a option like `renderError` or something so components could handler their own errors. 
In that case we could pass all the values `this.props.render({data, error, loading})`
But that requires refactoring the portal too

alex-wilmer: makes sense

PR #202, opened 2018-03-05
🐛 transcription fix for global_aggs

alex-wilmer:
1) add ES_LOG variable to change server terminal output (thanks @jgnieuwhof )
2) fix "aggregations_should_filter_themselves" to handle agg panel values

### DIFF
--- a/modules/middleware/src/aggregations.js
+++ b/modules/middleware/src/aggregations.js
@@ -610,13 +610,15 @@ export default class AggregationProcessor {
           /*
           b = { k: innerValue for k, innerValue in v.items() if re.match(r"{}(:(stats|histogram))?".format(field), k) is not None}
           */
-          v = Object.entries(v).reduce((acc, [k, v]) => {
-            if (!k.match(new RegExp(`${field}(:(stats|histogram))?`)))
+          v = Object.entries(v).reduce((acc, [innerKey, innerValue]) => {
+            if (!innerKey.match(new RegExp(`${field}(:(stats|histogram))?`)))
               return acc;
 
-            acc[k] = v;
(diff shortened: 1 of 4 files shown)


modules/middleware/test/global_aggs.test.js
```diff
@@ -0,0 +1,66 @@
...
+              },
+            },
+          },
+        },
+      },
+    },
+  },
+};
+
+test('build_aggregations should handle nested aggregations', () => {
```
hlminh2000: maybe a description like `"aggregations_filter_themselves: false", should give global aggs`

modules/middleware/src/aggregations.js
```diff
@@ -604,17 +606,17 @@ export default class AggregationProcessor {
...
-          const innerValues = filteredEntries.map(item => item[1]);
-          v = {
-            [k]: innerValues,
-          };
-          break;
+          // original python code
+          /*
+          b = { k: innerValue for k, innerValue in v.items() if re.match(r"{}(:(stats|histogram))?".format(field), k) is not None}
+          */
+          v = Object.entries(v).reduce((acc, [k, v]) => {
```
hlminh2000: could we name the value in side the reduce something else? Was a little confusing as `v` is also used outside

alex-wilmer: confusing you say! yes this was horribly confusing and I kept it for some reason. will fix

modules/middleware/test/global_aggs.test.js
alex-wilmer: thanks nice catch

PR #203, opened 2018-03-05
💄 prefix dashboard styles

Jephuff:


alex-wilmer:
maybe the dynamic import doesn't make as much sense. but still, 99% of users are never going to go there may as well not import the css, right?

PR #204, opened 2018-03-05
✨ Adds ability to search values in the afv

hlminh2000:
screen shot 2018-03-05 at 3 23 34 pm

- [x] Include values in search results
- [x] Scroll to value when search result is selected

### DIFF
--- a/modules/components/src/State.js
+++ b/modules/components/src/State.js
@@ -6,17 +6,30 @@ class State extends React.Component {
     this.state = props.initial;
   }
   componentDidMount() {
-    if (this.props.async) {
-      Promise.resolve(this.props.async()).then(this.update);
-    }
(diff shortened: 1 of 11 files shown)


modules/components/src/State.js
```diff
@@ -6,17 +6,31 @@ class State extends React.Component {
     this.state = props.initial;
   }
   componentDidMount() {
-    if (this.props.async) {
-      Promise.resolve(this.props.async()).then(this.update);
-    }
+    const { async = () => {} } = this.props;
+    Promise.resolve(async).then(this.update);
```
Jephuff: should async be called here?

hlminh2000: oops, yes!

PR #205, opened 2018-03-05
✅ add elasticsearch to travis tests

alex-wilmer:


alex-wilmer:
🎉 this actually works.

PR #212, opened 2018-03-07
✨update state fields

Jephuff:


modules/server/src/utils/getIndexPrefix.js
```diff
@@ -0,0 +1,9 @@
+function getIndexPrefix({ projectId, type }) {
```
alex-wilmer: why take the object type here?

Jephuff: instead of just index?

Jephuff: I updated to just pass index 👍  initially I had thought I would have type available everywhere, but that was wrong. this makes more sense! :)

alex-wilmer: yea it looks like you're only using `index` here, and in some places above passing `type : { index }`. was just wondering if maybe we would use it later

PR #214, opened 2018-03-07
✨ use jsonpath for tsv

Jephuff:


modules/server/src/utils/__tests__/dataToTSV.js
```diff
@@ -0,0 +1,262 @@
...
+        },
+        {
+          Header: 'Test2',
+          field: 'test2',
+          accessor: 'test2',
+        },
+      ],
+    };
+
+    const expected = 'Test1	Test2\n1	txt1\n2	txt2';
```
alex-wilmer: can these be updated to show the tabs explicitly? `\t` this feels easily breakable in a way that might be hard to understand why

Jephuff: done :)

PR #216, opened 2018-03-07
🐛 types is 'entries' in resolvers

alex-wilmer:


alex-wilmer:
well then! https://github.com/overture-stack/arranger/pull/215

PR #217, opened 2018-03-07
✨ applies values length limit on afv

hlminh2000:


modules/components/src/Aggs/TermAgg.js
```diff
@@ -120,7 +121,11 @@ class TermAggs extends React.Component {
                             '-',
                           )}`}
                         />
-                        {bucket.name}
+                        {!valueCharacterLimit
```
Jephuff: You can use lodash truncate to avoid these checks. You could even do something like 
```
truncate(str, {length: valueCharacterLimit  || Infinity})
```
to default to the entire value :)

modules/components/src/SQONView/index.js
```diff
@@ -36,12 +36,17 @@ export const Op = ({ children, ...props }: { children?: mixed }) => (
...
   className = '',
+  valueCharacterLimit,
   ...props
 }: {
   children?: mixed,
 }) => (
   <Bubble className={`sqon-value ${className}`} {...props}>
-    {children}
+    {!valueCharacterLimit
+      ? children
```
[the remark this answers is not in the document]
hlminh2000: forgot that it the component was exposed from all the way up :P Thanks!

modules/components/src/AdvancedFacetView/index.js
```diff
@@ -129,11 +130,9 @@ export default class AdvancedFacetView extends React.Component {
                 }}
                 {...props}
               >
-                {!valueCharacterLimit
-                  ? value
-                  : `${value}`.length > valueCharacterLimit
-                    ? `${value.slice(0, valueCharacterLimit)}...`
-                    : `${value}`}
+                {truncate(`${value}`, {
```
Jephuff: Shouldn't need to cast the value, truncate will do that.

PR #220, opened 2018-03-07
💩 refactors to allow field types to exclude to be passed in

hlminh2000:
- refactors to allow field types to exclude to be passed in
- includes 'date' type in default

### DIFF
--- a/modules/components/src/AdvancedFacetView/LiveAdvancedFacetView.js
+++ b/modules/components/src/AdvancedFacetView/LiveAdvancedFacetView.js
@@ -154,7 +154,7 @@ export default class LiveAdvancedFacetView extends React.Component {
           mapping: removeFieldTypesFromMapping({
             mapping,
             extended,
-            fieldTypesToExclude: fieldTypesToExclude,
+            fieldTypesToExclude,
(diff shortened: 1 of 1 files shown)


modules/components/src/AdvancedFacetView/LiveAdvancedFacetView.js
```diff
@@ -147,15 +149,15 @@ export default class LiveAdvancedFacetView extends React.Component {
...
         ...fetchConfig,
       }).then(({ aggregations }) => {
-        const fieldsTypesToExclude = ['id', 'text'];
+        const { fieldTypesToExclude = defaultFieldTypesToExclude } = this.props;
         this.setState({
           mapping: removeFieldTypesFromMapping({
             mapping,
             extended,
-            fieldTypesToExclude: fieldsTypesToExclude,
+            fieldTypesToExclude: fieldTypesToExclude,
```
alex-wilmer: object shorthand?

hlminh2000: not sure why it wasn't :P

PR #234, opened 2018-03-08
🐛use same get value function

Jephuff:


Jephuff:
... not sure why it deleted the components package-lock.json... going to try updating lerna

PR #237, opened 2018-03-08
Feat/arrangable aggs

hlminh2000:


modules/components/src/Admin/previews/AggPreviews.js
```diff
@@ -0,0 +1,210 @@
...
+        ...acc,
+        [key]: {
+          height: termAgg.container.clientHeight,
+          width: termAgg.container.clientWidth,
+        },
+      }),
+      {},
+    );
+
+    const sorted = sortBy(newLayout, i => i.y);
```
hlminh2000: somehow lodash's `sortedIndexBy` was returning weird results

Jephuff:
hmm, can we just re organize the array based on the layout? I don't love saving the react-grid-layout stuff and sorting on the frontend.

PR #240, opened 2018-03-08
only one toast of each version

Jephuff:


alex-wilmer:
too much toast?!

Jephuff:
too much toast!

PR #253, opened 2018-03-09
✨lower chunk size

Jephuff:
It was taking too long until the first byte, this makes it more responsive.

### DIFF
--- a/modules/server/src/utils/getAllData.js
+++ b/modules/server/src/utils/getAllData.js
@@ -6,7 +6,7 @@ function getAllData({
   projectId,
   index,
   variables = null,
-  size = 10000,
+  chunkSize = 1000,
   fields = '',
   mock,
 }) {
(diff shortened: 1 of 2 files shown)


modules/server/src/utils/getAllData.js
```diff
@@ -6,7 +6,7 @@ function getAllData({
   projectId,
   index,
   variables = null,
-  size = 10000,
+  chunkSize = 1000,
```
alex-wilmer: env var?

alex-wilmer: or possibly even something configurable at the `arranger-projects` level? would be a lot easier to test chunk sizes that way

Jephuff: I was going to make it a param, so if you had a particular large one that was timing out you could set it from the point of use. but then I didn't. 
what do you think of that? otherwise I will just make an env for now :D

alex-wilmer: param is a great idea

Jephuff: made it a param 👍

PR #255, opened 2018-03-12
uses display name in afv SqonView and CurrentSQON

hlminh2000:


hlminh2000:
Going to merge for quick iteration

PR #257, opened 2018-03-12
🐛 fix handling of the  field in the extended mapping

Jephuff:
Array fields were coming back as `[object Object]` even if they were marked as `isArray`


Jephuff:
@alex-wilmer If this looks reasonable, I can clean it up a bit before merging.

PR #269, opened 2018-03-13
✨ add search_after param

Jephuff:


modules/mapping-utils/src/createConnectionResolvers.js
```diff
@@ -1,15 +1,16 @@
+// @flow
 import { get } from 'lodash';
 
 import resolveAggregations from './resolveAggregations';
 import resolveHits from './resolveHits';
 
-type TcreateConnectionResolversArgs = {
+type TCreateConnectionResolversArgs = {
```
alex-wilmer: not that this is the most important line of code in the entire application, but I kind of prefer it the other way. it's like:

```
someFn
```

```
TsomeFn // type information
```

capitalizing the `s` seems weird to me

modules/mapping-utils/src/resolveHits.js
```diff
@@ -83,14 +87,18 @@ export default type => async (
...
+
+    return {
+      searchAfter: JSON.stringify(x.sort || []),
+      node: { id: x._id, ...source, ...nested_nodes },
+    };
   });
 
   return {
-    hits: nodes,
+    edges,
```
alex-wilmer: is this not a fairly major / breaking change?

Jephuff: nope! that change along with https://github.com/overture-stack/arranger/pull/269/files#diff-d51be1b4aaffd8cba6ee072e7ff32159L53
makes the graphql response the same.

modules/mapping-utils/src/createConnectionResolvers.js
Jephuff: hmmm, yeah I suppose. I can revert :) just feels weird to not have camel case haha

modules/mapping-utils/src/resolveHits.js
alex-wilmer: ohh cool 👍

modules/mapping-utils/src/createConnectionResolvers.js
Jephuff: reverted :)

PR #275, opened 2018-03-14
🐛 a couple fixes from the capsid perspective

jgnieuwhof:
a few things here:
 - a slightly more robust null check for the range agg, so it doesn't try to render the slider if it doesn't have the appropriate data to do so
 - separated the current sqon extended mapping api fetching / state from the current sqon view component
 - removed a react warning due to passing `nextSQON` prop through to a DOM element

### DIFF
--- a/modules/components/src/Aggs/RangeAgg.js
+++ b/modules/components/src/Aggs/RangeAgg.js
@@ -1,6 +1,8 @@
 import React, { Component } from 'react';
 import InputRange from 'react-input-range';
 import convert from 'convert-units';
+import _ from 'lodash';
+
 import { replaceSQON } from '../SQONView/utils';
 
 import 'react-input-range/lib/css/index.css';
@@ -114,7 +116,7 @@ class RangeAgg extends Component {
                 <span className={`arrow ${isCollapsed && 'collapsed'}`} />
               )}
             </div>
-            {[!isCollapsed, min != null, max != null].every(Boolean) && (
+            {[!isCollapsed, !_.isNil(min), !_.isNil(max)].every(Boolean) && (
               <div className="range-wrapper">
                 <div className="unit-wrapper">
                   {unit &&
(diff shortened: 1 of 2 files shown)


modules/components/src/Aggs/RangeAgg.js
```diff
@@ -114,42 +114,40 @@ class RangeAgg extends Component {
...
-                    <InputRange
-                      minValue={min}
-                      maxValue={max}
-                      value={value}
-                      formatLabel={this.formatRangeLabel}
-                      onChange={x => this.setValue(x)}
-                      onChangeComplete={this.onChangeComplete}
-                    />
-                  </div>
+            {[!isCollapsed, min, max].every(Boolean) && (
```
Jephuff: I think 0 would be valid here as well?

jgnieuwhof: spot on, nice catch, will fix

modules/components/src/Aggs/RangeAgg.js
```diff
@@ -114,42 +114,40 @@ class RangeAgg extends Component {
...
-                    <InputRange
-                      minValue={min}
-                      maxValue={max}
-                      value={value}
-                      formatLabel={this.formatRangeLabel}
-                      onChange={x => this.setValue(x)}
-                      onChangeComplete={this.onChangeComplete}
-                    />
-                  </div>
+            {[!isCollapsed, min != null, max != null].every(Boolean) && (
```
Jephuff: I'm going to be annoying and as for no `!=`. I know it does what you need, but it's hard to know in the future if it's on purpose or not :D 
maybe use `isNil` from lodash instead?

jgnieuwhof: ja thought this might come up, agree it seems like something that will be "corrected" in the future, will do

Jephuff: thanks! :D

PR #278, opened 2018-03-14
✨ add custom column

Jephuff:


Jephuff:
not super pretty, but works. put it add the bottom because it will rarely be used.
It could also be a popup?
image

modules/components/src/DataTable/ColumnsState.js
```diff
@@ -126,6 +126,18 @@ export default class extends Component {
     this.setState({ temp }, () => this.save(temp));
   };
 
+  add = column => {
+    const { id } = column;
+    let existing = this.state.config.columns.find(x => x.id === id);
+    if (existing) return;
+    let temp = {
+      ...this.state.config,
+      columns: this.state.config.columns.concat({ ...column }),
```
hlminh2000: tempted to say `[... this.state.config.columns, { ...column }]` but not a big thing :P

PR #282, opened 2018-03-15
✨ middleware refactor - fixes aggregations

Jephuff:
I put adding tests in a separate PR https://github.com/overture-stack/arranger/pull/284

renames `exclude` to `not-in`
renames `excludeifany` to `some-not-in`

combine `is` and `not` into `missing` with either a `true` or `false` value
```
{op: 'missing', content: {field, value: true }}
```

a `not` operator has been added to allow this
```
{
  op: 'not',
  content: [
    ...filters
  ]
}
```


andricDu:
Looking good from a correctness point of view. At least on the KF portal.

PR #289, opened 2018-03-16
Boolean aggs

hlminh2000:
screen shot 2018-03-20 at 2 10 07 pm

### DIFF
--- a/modules/components/src/Aggs/AggsWrapper.js
+++ b/modules/components/src/Aggs/AggsWrapper.js
@@ -5,9 +5,9 @@ import './AggregationCard.css';
 export default ({ children, collapsible = true, displayName }) => (
   <Component initialState={{ isCollapsed: false, showingMore: false }}>
     {({ setState, state: { isCollapsed, showingMore } }) => (
-      <div className="test-term-aggregation aggregation-card">
+      <div className="aggregation-card">
         <div
-          className={`title-wrapper ${isCollapsed && 'collapsed'}`}
+          className={`title-wrapper ${isCollapsed ? 'collapsed' : ''}`}
           onClick={
             collapsible
               ? () => setState({ isCollapsed: !isCollapsed })
@@ -16,11 +16,11 @@ export default ({ children, collapsible = true, displayName }) => (
         >
           <span className="title">{displayName}</span>
           {collapsible && (
-            <span className={`arrow ${isCollapsed && 'collapsed'}`} />
+            <span className={`arrow ${isCollapsed ? 'collapsed' : ''}`} />
           )}
         </div>
         {!isCollapsed && (
-          <div className={`bucket ${isCollapsed && 'collapsed'}`}>
+          <div className={`bucket ${isCollapsed ? 'collapsed' : ''}`}>
             {children}
           </div>
         )}
--- a/modules/components/src/Aggs/AggsWrapper.js
+++ b/modules/components/src/Aggs/AggsWrapper.js
@@ -3,8 +3,8 @@ import Component from 'react-component-component';
 import './AggregationCard.css';
 
 export default ({ children, collapsible = true, displayName }) => (
-  <Component initialState={{ isCollapsed: false, showingMore: false }}>
-    {({ setState, state: { isCollapsed, showingMore } }) => (
+  <Component initialState={{ isCollapsed: false }}>
+    {({ setState, state: { isCollapsed } }) => (
       <div className="aggregation-card">
         <div
           className={`title-wrapper ${isCollapsed ? 'collapsed' : ''}`}
--- a/modules/components/src/Aggs/AggsWrapper.js
+++ b/modules/components/src/Aggs/AggsWrapper.js
@@ -1,0 +1,30 @@
+import React from 'react';
+import Component from 'react-component-component';
+import './AggregationCard.css';
+
+export default ({ children, collapsible = true, displayName }) => (
+  <Component initialState={{ isCollapsed: false, showingMore: false }}>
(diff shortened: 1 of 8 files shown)


modules/components/src/Aggs/BooleanAgg.js
```diff
@@ -0,0 +1,119 @@
...
+    evaluateInSqon: ({ sqon }) =>
+      inCurrentSQON({
+        dotField,
+        value: booleanValues.false,
+        currentSQON: sqon,
+      }),
+  });
+  const isNeitherActive = !isTrueActive && !isFalseActive;
+
+  const handleTrueClick = () =>
```
alex-wilmer: do we really need both of these functions, and not just one function that takes a boolean argument?

alex-wilmer:
Could you include a screenshot of this agg in this pull request please?

modules/components/src/Aggs/BooleanAgg.js
hlminh2000: not sure why I kept two... yes, refactored!

modules/components/src/Aggs/AggsWrapper.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import Component from 'react-component-component';
+import './AggregationCard.css';
+
+export default ({ children, collapsible = true, displayName }) => (
+  <Component initialState={{ isCollapsed: false, showingMore: false }}>
+    {({ setState, state: { isCollapsed, showingMore } }) => (
+      <div className="test-term-aggregation aggregation-card">
+        <div
+          className={`title-wrapper ${isCollapsed && 'collapsed'}`}
```
Jephuff: make this `className={`title-wrapper ${isCollapsed ? 'collapsed' : ''}`}` or you will end up with a `false` class :)

modules/components/src/Aggs/AggsWrapper.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import Component from 'react-component-component';
+import './AggregationCard.css';
+
+export default ({ children, collapsible = true, displayName }) => (
+  <Component initialState={{ isCollapsed: false, showingMore: false }}>
+    {({ setState, state: { isCollapsed, showingMore } }) => (
+      <div className="test-term-aggregation aggregation-card">
```
Jephuff: not just for term aggregations?

modules/components/src/Aggs/AggsWrapper.js
```diff
@@ -0,0 +1,30 @@
...
+              : () => {}
+          }
+        >
+          <span className="title">{displayName}</span>
+          {collapsible && (
+            <span className={`arrow ${isCollapsed && 'collapsed'}`} />
+          )}
+        </div>
+        {!isCollapsed && (
+          <div className={`bucket ${isCollapsed && 'collapsed'}`}>
```
Jephuff: false class here too

modules/components/src/Aggs/BooleanAgg.css
```diff
@@ -0,0 +1,38 @@
+.booleanFacetWrapper {
+  display: flex;
+  flex-direction: row;
+  height: 30px;
+  font-size: 0.9em;
+}
+.booleanFacetAny {
```
Jephuff: a lot of shared styles between these. Could be good to make one selector for the general stuff

modules/components/src/Aggs/BooleanAgg.js
```diff
@@ -0,0 +1,124 @@
...
+const booleanValues = {
+  true: 'true',
+  false: 'false',
+};
+
+export default ({
+  field,
+  buckets,
+  handleValueClick = () => {},
+  isTrueSelected = () => false,
```
Jephuff: can this match the `isActive` api in the term facet? seems like it's doing the same thing

modules/components/src/Aggs/BooleanAgg.js
```diff
@@ -0,0 +1,124 @@
...
+      generateNextSQON: sqon =>
+        replaceSQON(
+          {
+            op: 'and',
+            content: [
+              {
+                op: 'in',
+                content: {
+                  field: dotField,
+                  value: [isTrue ? booleanValues.true : booleanValues.false],
```
Jephuff: what happens if you don't convert this to a string?

modules/components/src/Aggs/AggsWrapper.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
```
Jephuff: Could you use this in the range and term agg? If it doesn't happen now it will never happen :D

hlminh2000: done! :D had that as a todo for the AFV rewrite actually, but done anyway :)

Jephuff: ahh fair :)

modules/components/src/Aggs/BooleanAgg.js
```diff
@@ -0,0 +1,122 @@
...
+
+  const createSqonValidator = isTrue => ({ sqon }) =>
+    inCurrentSQON({
+      dotField,
+      value: isTrue,
+      currentSQON: sqon,
+    });
+
+  const isTrueActive = isActive({
+    defaultEvaluator: createSqonValidator(true),
```
Jephuff: I don't think you need `defualtEvaluator` or `createSqonValidator`. the term and boolean agg do the same thing for `isActive` so just let the parent handle this.

PR #290, opened 2018-03-16
✨ support all regex

Jephuff:
right now you can add a `*` to any value in a `in` or `not-in` filter which is replaced with a `.*` regular expression. 

this changes it so you have to explicitly say the filter is a regular expression and will allow all regular expression features.

before:
```
{
    op: 'in',
    content: {
        field: 'field',
        value: ['bbb', 'aaa*']
    }
}
```

after:
```
{
    op: 'or',
    content: [
        {
            op: 'in',
            content: {
                isRegex: true,
                field: 'field',
                value: ['aaa.*']
            }
        },
        {
            op: 'in',
            content: {
                field: 'field',
                value: ['bbb']
            }
        }
    ]
}
```

### DIFF
--- a/modules/middleware/src/buildQuery/index.js
+++ b/modules/middleware/src/buildQuery/index.js
@@ -52,7 +52,7 @@ function getRegexFilter({ nestedFields, filter }) {
   const esFilter = wrapFilter({
     filter,
     nestedFields,
-    esFilter: { regexp: { [field]: value.replace('*', '.*') } },
+    esFilter: { regexp: { [field]: value } },
     isNot: NOT_IN_OP === op,
   });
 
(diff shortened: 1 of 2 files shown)


Jephuff:
instead of `isRegex` could make it a different op like
```
{
    op: 'or',
    content: [
        {
            op: 'in-regex',
            content: {
                field: 'field',
                value: ['aaa.*']
            }
        },
        {
            op: 'in',
            content: {
                field: 'field',
                value: ['bbb']
            }
        }
    ]
}
```

hlminh2000:
not sure about the specific implication, but the flag seems to be more flexible, in case the op and the content need to be set by different components

PR #292, opened 2018-03-19
Feat/date agg

hlminh2000:


modules/components/package.json
```diff
@@ -72,8 +72,10 @@
     "js-cookie": "^2.2.0",
     "jsonpath": "^1.0.0",
     "lodash": "^4.17.4",
+    "moment": "^2.21.0",
```
Jephuff: can you use `date-fns` instead of moment.

Jephuff: booo react-dates requires moment :(

modules/components/src/Aggs/DatesAgg.js
```diff
@@ -0,0 +1,295 @@
...
+          (startFromSqon && bucketDateToMoment(endFromSqon)) ||
+          maxBucket?.moment,
+      };
+    };
+
+    const initialRange = getInitialRange();
+
+    return (
+      <AggsWrapper {...{ displayName, collapsible }}>
+        <Component
```
Jephuff: Can you refactor some of this logic to be outside the render? it's getting a bit busy.
This state can be moved to the class, and the functions using it can be moved to class methods :)

modules/components/src/Aggs/DatesAgg.js
```diff
@@ -0,0 +1,293 @@
...
+      })
+      ?.find(({ op }) => op === (dateToGet === START_DATE_INPUT ? '>=' : '<='))
+      ?.content.value;
+  };
+
+  handleInputValueChange = ({ value, input }) => {
+    const newMoment = inputDateToMoment(value);
+    const isValidInputDateString =
+      newMoment.isValid() &&
+      sumBy(value.split('/'), str => str.length === 2) === 3;
```
Jephuff: took me a second to realize what this line was doing. Might be a bit clearer to use a regex or something. I think it's checking for `value.match(/^\d\d\/\d\d\/\d\d$/)`?

modules/components/src/Aggs/DatesAgg.js
```diff
@@ -0,0 +1,293 @@
...
+    const {
+      displayName = 'Unnamed Field',
+      collapsible = true,
+      datePickerPosition = 'BOTTOM_LEFT',
+      numberOfMonths = 2,
+    } = this.props;
+
+    const { localRange, inputRangeValues, focusedInput } = this.state;
+
+    const initialRange = this.getInitialRange(this.props);
```
Jephuff: This value is already stored in state isn't it?

modules/components/src/Aggs/DatesAgg.css
```diff
@@ -0,0 +1,68 @@
+.dateInput {
```
Jephuff: Could you use emotion for the styles? or prefix them, these class names are just a little too generic.

modules/components/stories/Aggs.js
```diff
@@ -175,6 +177,44 @@ storiesOf('Aggs', module)
...
+            <DatesAgg
+              field="disease_type"
+              displayName="Disease Type"
+              buckets={require('./dummyData/datesBucketsSample.json')}
+              startDateFromSqon={({ getDateFromSqon }) => getDateFromSqon(sqon)}
+              endDateFromSqon={({ getDateFromSqon }) => getDateFromSqon(sqon)}
+              handleClearClick={({ generateNextSQON }) => {
+                setState({ sqon: generateNextSQON(sqon) });
+              }}
+              handleDateChange={({ generateNextSQON = () => {} } = {}) =>
```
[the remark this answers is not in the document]
hlminh2000: :P `handleClearClick` isn't actually used, just forgot to take it off the story

modules/components/src/Aggs/DatesAgg.css
hlminh2000: opted for CSS so it can be overridden if needed :)

PR #295, opened 2018-03-21
🐛 normalize / denormalize data in / out of range agg

jgnieuwhof:
Apparently the input range slider component we're using only handles integers, not doubles. In an effort to not swap out / rebuild the component altogether I'd like to suggest we add the ability to normalize / denormalize the numbers we give to & get from the range agg. Ex. normalize percentages to be between 0 & 100, rather than 0 & 1 (`normalize={x => x * 100}`, `denormalize={x => x / 100}`)

Also I've enabled the 'draggableTrack' prop, which lets you drag the previously selected range around. Ex. select 5->10, drag range to 4->9, 3->8, etc...

### DIFF
--- a/modules/components/src/Aggs/RangeAgg.js
+++ b/modules/components/src/Aggs/RangeAgg.js
@@ -4,6 +4,7 @@ import convert from 'convert-units';
 import _ from 'lodash';
 
 import { replaceSQON } from '../SQONView/utils';
+import mapObjectValues from '../utils/mapObjectValues';
 import AggsWrapper from './AggsWrapper';
 
 import 'react-input-range/lib/css/index.css';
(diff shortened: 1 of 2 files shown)


Jephuff:
Could the normalizing just happen around the input? 
```
<InputRange
  minValue={normalize(min)}
  onChange={({min, max}) => onChange({ min: denormalize(min), max: denormalize(max)}}
  ...
/>
```

Maybe even just take a multiplier instead of functions? `minValue={min * multiplier}`

Just some thoughts!

Jephuff:
maybe a smaller change, did you try passing `step={0.1}`?

jgnieuwhof:
@Jephuff - (facepalm)... step works, will do

jgnieuwhof:
@Jephuff - updated, now using step, thanks for the nudge in the right direction!

PR #301, opened 2018-03-22
Reimplements side bar in AFV

hlminh2000:


modules/components/src/AdvancedFacetView/index.js
```diff
@@ -13,12 +13,36 @@ import Input from '../Input';
...
+      ...(children
+        ? {
+            children: orderDisplayTreeData(children),
+          }
+        : {}),
+    }),
+  );
+  return [
+    ...orderBy(
+      setWithChildrenOrdered.filter(({ children }) => !children),
```
Jephuff: You don't need setWithChildrenOrdered the children here, because there are none 
That also means you can pre filter displayTreeData and remove the check for children in the map.

modules/components/src/AdvancedFacetView/SearchBox.js
```diff
@@ -224,60 +225,17 @@ export default class extends React.Component {
               },
               () => {
                 const newFilteredList = getCombinedFilteredList();
-                this.setState({
```
Jephuff: Do we need these callbacks? could we refactor it to not need them?

hlminh2000: yes, this will be cleaned up after as part of reimplementation of the searchbox (a separate issue :) )

PR #303, opened 2018-03-23
reimplements afv facet view

hlminh2000:


modules/components/src/AdvancedFacetView/FacetView.js
```diff
@@ -1,46 +1,106 @@
...
+              extendedMapping.find(({ field }) => field === path)?.displayName,
           );
+          const agg = aggregations[path];
+          return aggComponents[type]?.({
+            ...agg,
+            key: path,
+            field: path,
+            onValueChange: ({ sqon }) => {
+              onValueChange({ sqon });
+              const updateSubs = this.$updated.subscribe(() => {
```
Jephuff: Is this to get the scroll to happen after update? could you just setstate with the path and check if it changed in componentDidUpdate?

modules/components/src/AdvancedFacetView/FacetView.js
```diff
@@ -1,46 +1,106 @@
...
+                <div className={`header`}>
+                  <div className={`title`}>
+                    <TextHighlight
+                      content={title}
+                      highlightText={searchString}
+                    />
+                  </div>
+                  <div className={`breadscrumb`}>
+                    {pathDisplayNames
+                      .slice(0, pathDisplayNames.length - 1)
```
Jephuff: you can use -1 if you want to just remove the last item `.slice(0, -1)`

modules/components/src/AdvancedFacetView/aggComponents.js
```diff
@@ -0,0 +1,80 @@
...
+      onValueChange({ sqon: generateNextSQON(sqon) });
+    }}
+    isActive={d =>
+      inCurrentSQON({
+        value: d.value,
+        dotField: d.field,
+        currentSQON: sqon,
+      })
+    }
+    {...{ ...rest }}
```
Jephuff: You only need one spread :)

modules/components/src/Aggs/BooleanAgg.js
```diff
@@ -8,24 +8,29 @@ import {
...
+  searchString,
   ...rest
 }) => {
   const trueBucket = buckets.find(({ key }) => key === '1');
   const falseBucket = buckets.find(({ key }) => key === '0');
   const dotField = field.replace(/__/g, '.');
 
   const isTrueActive = isActive({
-    value: true,
+    value: 'true',
```
Jephuff: what happens if these are booleans?

modules/components/src/AdvancedFacetView/aggComponents.js
hlminh2000: lol :P copy-past byproduct :) will clean up

modules/components/src/AdvancedFacetView/FacetView.js
hlminh2000: yes, it's doing exactly that :) though using state seems to be triggering too many rendering cycles that it's not scrolling properly :/

modules/components/src/Aggs/BooleanAgg.js
hlminh2000: seems there's a truthy check somewhere that would break it... I think using strings make it more strict

Jephuff: might be work finding those checks and fixing them. It could break on other values as well :)

hlminh2000: so the server currently doesn't handle `false` value in sqon, so we're going to stick to using strings to abstract the values as ES is ok with querrying `false` with `"false"` :)

modules/components/src/AdvancedFacetView/index.js
```diff
@@ -1,26 +1,48 @@
...
+  <Spinner
+    fadeIn="none"
+    name="circle"
+    color="#a9adc0"
+    style={{
+      width: 30,
+      height: 30,
+    }}
+  />
+);
```
jgnieuwhof: Is this spinner applicable to other parts of arranger? If so it might be worth giving it its own file so it's obvious for re-use.

modules/components/src/AdvancedFacetView/index.js
```diff
@@ -1,26 +1,48 @@
...
 
 export default class AdvancedFacetView extends React.Component {
   constructor(props) {
     super(props);
+    const { elasticMapping, extendedMapping } = props;
```
jgnieuwhof: Unused vars?

modules/components/src/AdvancedFacetView/index.js
```diff
@@ -39,6 +61,66 @@ export default class AdvancedFacetView extends React.Component {
...
+  componentDidUpdate(prevProps, prevState, { shouldEndLoading }) {
+    const shouldRecomputeDisplayTree = !isEqual(
+      {
+        elasticMapping: this.props.elasticMapping,
+        extendedMapping: this.props.extendedMapping,
+      },
+      {
+        elasticMapping: prevProps.elasticMapping,
+        extendedMapping: prevProps.extendedMapping,
+      },
```
jgnieuwhof: not really important, but could use lodash `pick` for this type of object projection

modules/components/src/AdvancedFacetView/index.js
```diff
@@ -47,178 +129,188 @@ export default class AdvancedFacetView extends React.Component {
...
+                              });
+                            }}
+                          />
+                        }
+                        className="filterInput"
+                        type="text"
+                        placeholder="Filter"
+                        value={value || ''}
+                        showClearButton={true}
+                        onChange={e => {
```
jgnieuwhof: could destructure `e` directly, rather than declaring the intermittent var

modules/components/src/AdvancedFacetView/index.js
```diff
@@ -47,178 +129,188 @@ export default class AdvancedFacetView extends React.Component {
...
+              position: absolute;
+              left: 0px;
+              right: 0px;
+              top: 0px;
+              bottom: 0px;
+              display: flex;
+              justify-content: center;
+              align-items: center;
+            `}`}
+          >
```
jgnieuwhof: not sure, but would the loading screen be worth componentizing for re-use as well?

modules/components/src/Aggs/BooleanAgg.js
```diff
@@ -58,13 +75,12 @@ export default ({
     });
 
   return (
-    <AggsWrapper {...{ buckets, ...rest }}>
+    <AggsWrapper {...{ displayName, WrapperComponent, collapsible }}>
       <div className={`booleanFacetWrapper`}>
         <div
```
jgnieuwhof: If you feel like it I suppose this could be the second use of `ToggleButton`

modules/components/src/Aggs/TermAgg.js
```diff
@@ -62,6 +64,24 @@ class TermAggs extends React.Component {
 
   state = { showingMore: false, isExclude: false };
 
+  componentWillReceiveProps(nextProps) {
```
jgnieuwhof: I kind of feel like the term agg search filter should be along the lines of `showExcludeOption`, basically a prop that can toggle on/off whether or not the term agg should show the text filter, and then it can pass the text input box into the AggsWrapper `filters` prop. thoughts?

modules/components/src/Aggs/BooleanAgg.js
hlminh2000: updated BooleanAgg to render using the ToggleButton :) thanks!

PR #321, opened 2018-04-05
✨ adds alwaysSendHeaders to download params

hlminh2000:


modules/components/src/utils/download.js
```diff
@@ -99,7 +100,7 @@ function download({ url, params, method = 'GET' }) {
     ? Math.abs(hashString(JSON.stringify(params) + downloadToken)).toString(16)
     : null;
 
-  let fields = toHtml('params', params);
+  let fields = toHtml('params', { ...params, ...getAlwaysAddHeaders() });
```
jgnieuwhof: should this be `{ ...getAlwaysAddHeaders(), ...params }` so that params take precedence over the headers?

PR #322, opened 2018-04-06
reimplements headers with better separation of concern

hlminh2000:


modules/components/src/utils/api.js
```diff
@@ -27,6 +28,7 @@ export const fetchExtendedMapping = ({ graphqlField, projectId }) =>
 
 export const addHeaders = headers => {
   alwaysSendHeaders = { ...alwaysSendHeaders, ...headers };
+  addDownloadHttpHeaders(headers);
```
alex-wilmer: this is kind of messy in my opinion. Do we need to make these imperative calls? can't we just pass in headers when we actually make calls vs "setting" them and using the set ones later on?

hlminh2000: Added this issue so we can come back to this later :) : https://github.com/overture-stack/arranger/issues/323

modules/components/src/utils/download.js
```diff
@@ -93,21 +94,31 @@ function createIFrame({ method, url, fields }) {
   return iFrame;
 }
 
-function download({ url, params, method = 'GET' }) {
+function download({ url, params, method = 'GET', customBody = {} }) {
```
alex-wilmer: can you name this just `body`?

hlminh2000: sure :)

PR #324, opened 2018-04-06
🐛 small grammar tweak

hlminh2000:


alex-wilmer:
this pr doesn't add the plural either, so the grammar error will remain

hlminh2000:
@alex-wilmer it was mentioned during the meeting that it should be "Showing 1 - 20 of 300 files", which I feel makes more sense https://github.com/kids-first/kf-portal-ui/issues/263

alex-wilmer:
hmm this got merged but the plural mistake will remain

jgnieuwhof:
@hlminh2000 - I thought you'd mentioned that it was now showing `files` rather than `file`?

hlminh2000:
ah I was looking at the admin dashboard story and it looked fine, seems the component is used differently in the portal. I'll take a look

PR #325, opened 2018-04-10
 ✨ add ability to save sets

jgnieuwhof:
This adds an endpoint to save sets.

Some TODOs:
- set fetching
- set searching
- configurable set ID fetching via aggs rather than scroll ES api

### DIFF
--- a/modules/mapping-utils/src/resolveSets.js
+++ b/modules/mapping-utils/src/resolveSets.js
@@ -10,7 +10,7 @@ const retrieveSetIds = async ({
   path,
   BULK_SIZE = 1000,
 }) => {
-  const search = async ({ searchAfter }) => {
+  const search = async ({ searchAfter } = {}) => {
     const response = await es.search({
       index,
       type,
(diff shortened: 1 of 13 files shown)


Jephuff:
From what I had read, scroll search shouldn't be used for real time user requests because it uses a lot of memory. Not sure if it will affect arranger, but might be worth looking into `search_after` instead (unless you already have, just thought I would mention it! )

jgnieuwhof:
@Jephuff - you're right, learn somethin' new each day... I'll go ahead and make that change

modules/mapping-utils/src/resolveSets.js
```diff
@@ -2,40 +2,36 @@ import { get, isEmpty, uniq } from 'lodash';
...
   BULK_SIZE = 1000,
 }) => {
-  const handleResult = async ({ scrollId, total, ids = [] }) => {
-    if (ids.length === total) return uniq(ids);
-    const response = await es.scroll({ scroll: SCROLL_TIME, scrollId });
-    return handleResult({
-      total,
-      scrollId: response._scroll_id,
-      ids: [...ids, ...mapIds({ response, path })],
+  const search = async ({ searchAfter }) => {
```
alex-wilmer: does it make more sense to add the default here instead of passing an empty object at the call spot?

```
async ({ searchAfter } = {}) =>
```

jgnieuwhof: yup

PR #326, opened 2018-04-11
✨ add support for set parsing in sqon

jgnieuwhof:
adds set_id parsing to sqon -> es middleware.

depends on https://github.com/overture-stack/arranger/pull/325


modules/middleware/src/buildQuery/index.js
```diff
@@ -181,13 +181,40 @@ function getGroupFilter({ nestedFields, filter: { content, op } }) {
   return wrappers[op](esFilters);
 }
 
+function getSetFilter({ nestedFields, filter }) {
+  const { content } = filter;
+
+  const setId = (Array.isArray(content.value)
```
alex-wilmer: functional style yo

```
const [setId] = flatMap([content.value])
```

alex-wilmer: or

```
const [,setId] = flatMap([content.value]).split('set_id:')
```

jgnieuwhof:
@alex-wilmer - this should be g2g pending review

jgnieuwhof:
also I should note that this is code shamelessly stolen from @Jephuff's PR that was a little too stale to deal w/ the conflicts.

PR #327, opened 2018-04-11
🐛 adds s for plural

hlminh2000:
Didn't want to do a big refactor to lift state up just to add a character, so I went with an observable...

### DIFF
--- a/modules/components/package-lock.json
+++ b/modules/components/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@arranger/components",
-  "version": "0.3.10",
+  "version": "0.3.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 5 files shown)


jgnieuwhof:
Ok so I bounced this idea off @alex-wilmer - he suggested we move the `Showing...` markup into being its own component, that fetches its own data count (using `Query`). This would decouple the `Showing` component from the table that's fetching the actual data, without the need for observables.

Another note - we should probably use something like `pluralize` to do the pluralization, it's not always the case that adding an `s` to a word is the right thing to do to make it plural.

Thoughts?

hlminh2000:
@jgnieuwhof I reverted and tried reimplementing it with that idea :)

Jephuff:
Shouldn't the tool bar already have all the information it needs to add the s? 
```
total={total}
page={page}
pageSize={pageSize}
```

something like `const isPlural = total > 1 && pageSize > 1 && (Math.ceil(total / pageSize) !== page || total % pageSize > 1)`

Maybe thats too much.

sorry, I will leave :P

hlminh2000:
@Jephuff that in fact is a very good point... probably much cleaner to make all the other `type` fed in singular instead.

hlminh2000:
@jgnieuwhof @Jephuff interestingly, `pluralize` knows to not add `s` when it's already a plural word

alex-wilmer:
> sorry, I will leave :P

don't leave us!

PR #331, opened 2018-04-13
✨ add quick-search component

jgnieuwhof:
This adds a quick search component, here's an overview:
- added 'quick search enabled' and 'primary key' fields to extended mapping. 'quick search enabled' means that we'll search the field in quick search. 'primary key' is the field that gets 'pinned' when an item is selected
- added a quick search component, can be placed anywhere but will typically be placed above aggregations

minor extras:
- fixed a bug in `withQuery` where it was ignoring the fact that it actually shouldn't fetch
- enhanced our text highlighter to accept a configurable highlight style


jgnieuwhof:
going to modify this PR to separate the query from the quick search component (for re-use in the matchbox), and to handle multiple identifiers

jgnieuwhof:
@alex-wilmer - should be good to go now, separation of query component + handling multiple identifiers is done.

modules/components/src/Arranger/QuickSearch/QuickSearch.js
```diff
@@ -0,0 +1,174 @@
...
+            icon={loading ? LoadingIcon : Icon}
+            type="text"
+            value={value}
+            componentRef={inputRef}
+            placeholder={placeholder}
+            onChange={({ target: { value } }) => setValue(value || '')}
+          />
+
+          <div
+            className={css`
```
alex-wilmer: css prop

modules/components/src/Arranger/QuickSearch/QuickSearch.js
```diff
@@ -0,0 +1,174 @@
...
+              </div>
+            ))}
+          </div>
+          <TextInput
+            icon={loading ? LoadingIcon : Icon}
+            type="text"
+            value={value}
+            componentRef={inputRef}
+            placeholder={placeholder}
+            onChange={({ target: { value } }) => setValue(value || '')}
```
alex-wilmer: default assignment? `value = ''` i think its nicer

PR #336, opened 2018-04-19
TableToolbar: allow custom filename for export tsv

cy:
for https://github.com/nci-hcmi-catalog/portal/issues/20


modules/components/src/DataTable/TableToolbar/TableToolbar.js
```diff
@@ -41,6 +41,7 @@ const TableToolbar = ({
   style,
   columnDropdownText = 'Show columns',
   exportTSVText = 'Export TSV',
+  exportTSVFilename = `${type}-table.tsv`,
```
hlminh2000: can you actually reference `type` here...? didn't know js allowed that

hlminh2000: apparently it works in Chrome! :D

PR #337, opened 2018-04-19
💄 More accurate afv style according to design

hlminh2000:
style ref: https://zpl.io/bP4BdYD

### DIFF
--- a/modules/components/src/AdvancedFacetView/AdvancedFacetView.css
+++ b/modules/components/src/AdvancedFacetView/AdvancedFacetView.css
@@ -99,20 +99,22 @@
   flex-wrap: wrap;
 }
 
-.facetViewWrapper .bucket .bucket-item:nth-child(4n + 1),
-.facetViewWrapper .bucket .bucket-item:nth-child(4n + 2),
-.facetViewWrapper .bucket .bucket-item:nth-child(4n + 3) {
-  border-bottom: solid 1px #e0e1e6;
+.facetViewWrapper .bucket .bucket-item:nth-child(4n + 1):not(:last-child),
+.facetViewWrapper .bucket .bucket-item:nth-child(4n + 2):not(:last-child),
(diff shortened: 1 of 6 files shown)


modules/components/src/AdvancedFacetView/FacetView.js
```diff
@@ -77,6 +77,28 @@ export default class FacetView extends React.Component {
...
+                      bucket,
+                      showingBuckets,
+                      i,
+                      showingMore,
+                    }) =>
+                      `row_${Math.floor(i / 4)} col_${i % 4} ${
+                        Math.floor(i / 4) ===
+                        Math.floor((showingBuckets.length - 1) / 4)
+                          ? 'last_row'
+                          : ''
```
jgnieuwhof: Just checking - advanced facet view is currently hardcoded for 4 columns?

modules/components/src/AdvancedFacetView/AdvancedFacetView.css
```diff
@@ -98,20 +99,22 @@
...
-.facetViewWrapper .bucket .bucket-item:nth-child(3),
-.facetViewWrapper .bucket .bucket-item:nth-child(4) {
-  border-bottom: solid 1px #e0e1e6;
+
+.facetViewWrapper .bucket .bucket-item.only_row {
+  border-bottom: none;
+}
+
+.facetViewWrapper .bucket .bucket-item:last-child:not(.col_3):not(.only_row) {
+  border-right: solid 1px #e0e1e6;
```
jgnieuwhof: It seems like there might be some colors / styles here that are beagle specific rather than arranger specific, is now a good time to move 'em over or should we hold off?

hlminh2000: yeah this probably should be done in beagle :) I'll update

modules/components/src/AdvancedFacetView/FacetView.js
hlminh2000: yes, it is for now :)

PR #340, opened 2018-04-23
🐛 fixed sqon set parsing (to handle > 1 set at a time)

jgnieuwhof:
a sqon set is a 'special' filter, to be separated & "or'd"


modules/middleware/src/buildQuery/normalizeFilters.js
```diff
@@ -21,7 +21,7 @@ function groupingOptimizer({ op, content }) {
 }
 
 function isSpecialFilter(value) {
-  return `${value}`.includes('*');
+  return `${value}`.includes('*') || `${value}`.includes('set_id:');
```
alex-wilmer: in case this gets long you could do

```
return ['*', 'set_id'].some(x => `${value}.includes(x)`)
```

PR #342, opened 2018-04-24
:sparkles: resolve sets

cy:
Instead of custom sets resolver, concat sets into types array being passed to `generateTypeDefs` and `generateResolvers`. Works but let me know if it's a terrible idea.


cy:
@jgnieuwhof updated hopefully didnt break anything else :x

PR #351, opened 2018-04-27
🐛 date range picker should popout like a modal

jgnieuwhof:
to avoid the case where it's in a sidebar and is cutoff

also included a fix to display 'YYYY-MM-DD' in the SQON / UI, and ensure that in our middleware we translate it into the longer format that ES will accept


jgnieuwhof:
@alex-wilmer - converted middleware `moment` to `date-fns`, I started converting the UI but realized that `react-dates` requires `moment` so I left it as is

modules/components/src/Aggs/DatesAgg.js
```diff
@@ -1,145 +1,47 @@
...
     super(props);
     this.state = {
-      localRange: this.getInitialRange(props), // the moment objects passed into DayPickerRangeController
-      inputRangeValues: {
-        // the strings to render in the inputs
-        startDate: null,
-        endDate: null,
-      },
+      ...this.getInitialRange(props),
       focusedInput: null, // must be set with this.setInputFocus to bind with dom state
```
hlminh2000: `this.setInputFocus` is not a thing anymore so we probably don't need the comment. In fact, do we still need this state? I think `DateRangePicker` can manage its own input state, unless we're controlling it with props :P

hlminh2000: @jgnieuwhof confirmed from doc that DateRangePicker needs the state managed outside: https://github.com/airbnb/react-dates#daterangepicker 

so... 🥇

PR #352, opened 2018-04-30
saveSet: allow specify sort and default id field to _id

cy:
Figured out why sets of size >1000 would not create for HCMI data:

1) there's no id field in the hcmi data source, so `path: "id"` fails on `get(x, '_source.${path.split('__').join('.')}')`, so default to `_id`
2) if `path: name` and sort: [`_id`], `search_after` doesnt work properly, it keeps returning pages of 1000 and never reaches the end resulting in an infinite loop. Allow `sort` to be passed in.

### DIFF
--- a/modules/components/src/utils/saveSet.js
+++ b/modules/components/src/utils/saveSet.js
@@ -3,7 +3,7 @@ import { graphql } from './api';
 export default ({ type, path, userId, sqon = {}, returnIds = false, api }) =>
   (api || graphql)({
     query: `
-      mutation saveSet($type: String! $userId: String! $sqon: JSON! $path: String!) {
+      mutation saveSet($type: String! $userId: String $sqon: JSON! $path: String!) {
         saveSet(type: $type, userId: $userId, sqon: $sqon, path: $path) {
           setId
           createdAt
--- a/modules/components/src/utils/saveSet.js
+++ b/modules/components/src/utils/saveSet.js
@@ -1,10 +1,18 @@
 import { graphql } from './api';
 
-export default ({ type, path, userId, sqon = {}, returnIds = false, api }) =>
+export default ({
+  type,
+  path,
+  userId,
+  sqon = {},
+  returnIds = false,
+  api,
+  sort = [],
+}) =>
   (api || graphql)({
     query: `
-      mutation saveSet($type: String! $userId: String $sqon: JSON! $path: String!) {
-        saveSet(type: $type, userId: $userId, sqon: $sqon, path: $path) {
+      mutation saveSet($type: String! $userId: String $sqon: JSON! $path: String!, $sort: [Sort]) {
+        saveSet(type: $type, userId: $userId, sqon: $sqon, path: $path, sort: $sort) {
           setId
           createdAt
           path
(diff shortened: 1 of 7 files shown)


modules/components/src/utils/saveSet.js
```diff
@@ -3,7 +3,7 @@ import { graphql } from './api';
 export default ({ type, path, userId, sqon = {}, returnIds = false, api }) =>
```
jgnieuwhof: what about adding `sort` to this function / query, which defaults to an empty array, and in the resolver below checks for both `sort` and `sort.length` before defaulting to `['_id']` (or another method of defaulting to _id). Just might be nice to keep this helper function inline with the resolver.

cy:
@jgnieuwhof this is ready again

To allow for keeping table order and set order the same, `searchAfter` needs to be passed values from a unique field, and all fields in `sorted` needs to have values appear in `searchAfter`. So, added an 'alwaysSorted' prop to DataTable (different from `defaultSorted` in that `defaultSorted` gets removed on header click), which should be passed a unique field. The savedSets resolver is then passed the DataTable's current sorted array concat with `alwaysSorted`.

modules/mapping-utils/src/resolveSets.js
```diff
@@ -8,24 +8,40 @@ const retrieveSetIds = async ({
...
       ...(!isEmpty(query) && { query }),
       ...(searchAfter && { search_after: searchAfter }),
     };
+
     const response = await es.search({
       index,
       type,
-      sort: ['_id'],
+      sort: sort.map(
+        ({ field, order }) => `${field}${order ? `:${order}` : ''}`,
```
jgnieuwhof: should we be explicit here about order?
```
 `${field}:${order || 'asc'}`
```

modules/mapping-utils/src/resolveSets.js
```diff
@@ -8,24 +8,40 @@ const retrieveSetIds = async ({
...
       body,
     });
     const ids = response.hits.hits.map(x =>
-      get(x, `_source.${path.split('__').join('.')}`),
+      get(x, `_source.${path.split('__').join('.')}`, x._id || ''),
+    );
+
+    const sortColValues = sort.map(({ field }) =>
+      response.hits.hits.map(x => x._source[field] || x[field]),
     );
```
jgnieuwhof: I _think_ this map + the reduce below is equivalent to:
```
  const searchAfter = sort
    .map(({ field }) => response.hits.hits.map(x => x._source[field] || x[field]))
    .map(x => x.slice(-1))
```
& just use `searchAfter` in the args below... if you feel like it lol

cy: can't do the ` .map(x => x.slice(-1))` part cuz it'll come back with a nested array [['stuff'],['moreStuff']], need a flat map but didn't want to import anything so just used reduce

jgnieuwhof: ohhh right you are, nvm ma bad

cy:
thanks @jgnieuwhof :D updated

PR #363, opened 2018-05-07
🐛 keep term agg in view on collapse

jgnieuwhof:
2 things here:
 - changes to columnstate broke some stuff in KF / arranger admin, so put in some safety checks for that
 - when clicking 'less' on the term agg they want the term agg to stay in agg view (in a 'scrolling' aggregation sidebar context)

### DIFF
--- a/modules/components/src/Admin/Tabs/Aggregations/TableTab.js
+++ b/modules/components/src/Admin/Tabs/Aggregations/TableTab.js
@@ -13,30 +13,36 @@ export default ({ projectId, graphqlField }) => (
           flexDirection: 'row',
         }}
       >
-        <div style={{ flex: 1 }}>
-          <EditColumns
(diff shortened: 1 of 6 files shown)


modules/components/src/Admin/Tabs/Aggregations/TableTab.js
```diff
@@ -13,30 +13,36 @@ export default ({ projectId, graphqlField }) => (
...
-              ),
-              projectId,
-              graphqlField,
-              onOrderChange: newItemList =>
-                columnsState.saveOrder(newItemList.map(({ field }) => field)),
-            }}
-          />
-        </div>
+        {!columnsState.loading && (
+          <React.Fragment>
```
alex-wilmer: `<>` ?

PR #367, opened 2018-05-09
🐛 new date picker that can handle year changes

jgnieuwhof:
ditching react-dates, this one seems to work well

### DIFF
--- a/modules/components/package.json
+++ b/modules/components/package.json
@@ -7,12 +7,15 @@
     "access": "public"
   },
   "scripts": {
-    "prepare": "rm -rf dist && mkdir dist && babel src --out-dir dist --copy-files",
+    "prepare":
+      "rm -rf dist && mkdir dist && babel src --out-dir dist --copy-files",
     "watch": "babel src --out-dir dist --copy-files --watch",
     "build-storybook": "build-storybook -s ./public",
     "storybook": "start-storybook -p 8080 -c .storybook -s ./public",
-    "dashboard": "STORYBOOK_DASHBOARD=true start-storybook -p 6060 -c .storybook -s ./public",
-    "portal": "STORYBOOK_PORTAL=true start-storybook -p 7070 -c .storybook -s ./public",
+    "dashboard":
+      "STORYBOOK_DASHBOARD=true start-storybook -p 6060 -c .storybook -s ./public",
+    "portal":
+      "STORYBOOK_PORTAL=true start-storybook -p 7070 -c .storybook -s ./public",
     "test": "jest",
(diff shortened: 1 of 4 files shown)


modules/components/package.json
```diff
@@ -74,6 +74,7 @@
     "moment": "^2.21.0",
     "pluralize": "^7.0.0",
     "react-component-component": "^1.0.1",
+    "react-datepicker": "^1.4.1",
     "react-dates": "^16.3.6",
```
hlminh2000: just a dependency cleanup if no longer used :P also clear package-lock

modules/components/src/Aggs/DatesAgg.js
```diff
@@ -16,55 +16,61 @@ const momentToBucketDate = moment => moment?.format(BUCKET_DATE_FORMAT);
...
-                  content: {
-                    field,
-                    value: momentToBucketDate(endDate.endOf('day')),
-                  },
-                },
-              ],
-            },
-            sqon,
+            content.length ? { op: 'and', content } : null,
+            removeSQON(field, sqon),
```
hlminh2000: Just so I understand it fully, is this to handle the change above that's checking each field separately? :)

modules/components/src/Aggs/DatesAgg.js
```diff
@@ -16,55 +16,61 @@ const momentToBucketDate = moment => moment?.format(BUCKET_DATE_FORMAT);
...
   }
 
-  getInitialRange = ({ buckets = [], getActiveValue = () => ({}) }) => {
+  initializeState = ({ buckets = [], getActiveValue = () => ({}) }) => {
     const { field } = this.props;
     const bucketMoments = buckets.map(x => bucketDateToMoment(x.key_as_string));
-    const minDate = minBy(bucketMoments, moment => moment.valueOf());
-    const maxDate = maxBy(bucketMoments, moment => moment.valueOf());
+    const minDate = minBy(bucketMoments, x => x.valueOf()).subtract(1, 'days');
+    const maxDate = maxBy(bucketMoments, x => x.valueOf()).add(1, 'days');
```
hlminh2000: a null check here? `__missing__` as a bucket would cause a breakage on the chained Moment method call.

modules/components/src/Aggs/DatesAgg.js
jgnieuwhof: yeah, basically remove both >= and <=, then add them in one-by-one

jgnieuwhof:
@hlminh2000 - should be addressed

PR #368, opened 2018-05-09
💄 various requested changes

jgnieuwhof:
3 changes:
- term agg to show 5 terms by default
- missing bucket to show up last in terms list
- missing bucket to display as 'No Data', rather than 'Missing'


hlminh2000:
LGTM! 👍 😄

PR #375, opened 2018-05-14
Propagate the selected facet/agg value up to the highest parent component

bdolly:
When a user interacts with a facet/aggregation value propagate that call and it's value up to the highest parent component to allow for hooks into these events

### DIFF
--- a/modules/components/src/Aggs/RangeAgg.js
+++ b/modules/components/src/Aggs/RangeAgg.js
@@ -1,18 +1,18 @@
-import React, { Component } from 'react';
-import InputRange from 'react-input-range';
-import convert from 'convert-units';
-import _ from 'lodash';
+import React, { Component } from "react";
+import InputRange from "react-input-range";
+import convert from "convert-units";
+import _ from "lodash";
 
-import { replaceSQON } from '../SQONView/utils';
-import AggsWrapper from './AggsWrapper';
+import { replaceSQON } from "../SQONView/utils";
+import AggsWrapper from "./AggsWrapper";
 
-import 'react-input-range/lib/css/index.css';
-import './AggregationCard.css';
-import './RangeAgg.css';
+import "react-input-range/lib/css/index.css";
+import "./AggregationCard.css";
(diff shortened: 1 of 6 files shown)


alex-wilmer:
looks like your editor is ignoring the prettier config options. do you have eslint installed in your editor?

```
  'prettier/prettier': [1, { trailingComma: 'all', singleQuote: true }],
```

modules/components/src/Arranger/Aggregations.js
```diff
@@ -1,22 +1,23 @@
-import React from 'react';
+import React from "react";
```
alex-wilmer: prettier here too?

alex-wilmer:
I don't see a problem with this. @jgnieuwhof you wrote most of these components, what do you think?

hlminh2000:
my only thought is `field` should all be available from the wrapping components in `modules/components/src/Aggs/aggComponentsMap.js` and fed down, so it doesn't need to be propagated all the way up from the individual agg components

modules/components/src/Aggs/RangeAgg.js
```diff
@@ -1,18 +1,18 @@
-import React, { Component } from 'react';
-import InputRange from 'react-input-range';
-import convert from 'convert-units';
-import _ from 'lodash';
+import React, { Component } from "react";
```
alex-wilmer: oh more prettier stuff

alex-wilmer:
I think this change just adds that information on events. It's true could probably grab the same info inside the function body of the event where it's called, but this way it's just there?

hlminh2000:
yeah it doesn't make any difference for the data collection, would just be less code :P 
I'm ok with having it coming from the event too if that's too much hassle to change now :)

bdolly:
I like having it in the events because it provides "hooks" for the app that it's nested inside of since Arranger is typically used in a sub-app context. 

The main use case right now is hooking in analytics events to the various user interactions

hlminh2000:
ok that's cool for me then :)

PR #377, opened 2018-05-14
onTableExport prop to TableComponent

bdolly:
implements a onTSVExported callback function once the file has been downloaded

### DIFF
--- a/modules/components/src/utils/download.js
+++ b/modules/components/src/utils/download.js
@@ -55,9 +55,8 @@ function download({
 
   const resolveOnDownload = new Promise((resolve, reject) => {
     io.on(`server::download::${downloadKey}`, () => {
-      callback({ downloadKey });
       io.off(`server::download::${downloadKey}`);
-      resolve();
+      resolve({ downloadKey });
     });
   });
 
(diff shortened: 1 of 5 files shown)


modules/components/src/utils/download.js
```diff
@@ -43,12 +43,19 @@ function createIFrame({ method, url, fields }) {
   return iFrame;
 }
 
-function download({ url, params, method = 'GET', body = {} }) {
+function download({
+  url,
+  params,
+  method = 'GET',
+  body = {},
+  callback = () => {},
```
alex-wilmer: the name `callback` here could be a bit confusing. `onDownloadStart` or end? or whatever

Jephuff: Probably don't need the callback option here, saveTSV returns download() which returns the promise so it could just be `saveTSV({...}).then((downloadKey) => onTSVExported({ downloadKey, files }))` would just need to resolve with `downloadKey`

alex-wilmer: hehe i should've scrolled down. the diff shows the promise and I kinda figured but didn't look

PR #379, opened 2018-05-14
:bug: null in TableTab causing it to crash

cy:
table tab wasn't loading for me anymore. This fixes.

### DIFF
--- a/modules/components/src/Admin/Tabs/Aggregations/TableTab.js
+++ b/modules/components/src/Admin/Tabs/Aggregations/TableTab.js
@@ -26,7 +26,7 @@ export default ({ projectId, graphqlField }) => (
               <strong>Columns order</strong>
               <DraggableListOrderer
                 {...{
-                  itemsList: columnsState.state.columns.map(
+                  itemsList: columnsState.state?.columns?.map(
                     ({ show, ...rest }) => ({
                       active: show,
                       ...rest,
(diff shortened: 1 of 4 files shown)


modules/components/src/Admin/Tabs/Aggregations/TableTab.js
```diff
@@ -26,7 +26,7 @@ export default ({ projectId, graphqlField }) => (
               <strong>Columns order</strong>
               <DraggableListOrderer
                 {...{
-                  itemsList: columnsState.state.columns.map(
+                  itemsList: (columnsState.state || {}).columns.map(
```
alex-wilmer: what about `?.`

modules/components/src/Admin/Tabs/Aggregations/TableTab.js
```diff
@@ -26,7 +26,7 @@ export default ({ projectId, graphqlField }) => (
               <strong>Columns order</strong>
               <DraggableListOrderer
                 {...{
-                  itemsList: columnsState.state.columns.map(
+                  itemsList: columnsState.state?.columns.map(
```
alex-wilmer: can you add before map as well just in case?

```
columnsState.state?.columns?.map()
```

modules/components/src/Admin/Tabs/Aggregations/TableTab.js
cy: thanks, updated

modules/components/src/Admin/Tabs/Aggregations/TableTab.js
cy: ok

PR #380, opened 2018-05-14
add field value to onTermSelected prop

bdolly:
just adds dotField value into response for more context


bdolly:
this PR is not needed the field value is already included in the response

PR #383, opened 2018-05-15
Advanced Facet View onFacetNavigation prop

bdolly:


modules/components/src/Arranger/Aggregations.js
```diff
@@ -1,7 +1,7 @@
-import React from 'react';
+import React from "react";
 
-import { AggsState, AggsQuery } from '../Aggs';
-import aggComponents from '../Aggs/aggComponentsMap.js';
+import { AggsState, AggsQuery } from "../Aggs";
+import aggComponents from "../Aggs/aggComponentsMap.js";
```
jgnieuwhof: prettier config issue?

modules/components/src/AdvancedFacetView/FacetView.js
```diff
@@ -98,13 +98,14 @@ export default class FacetView extends React.Component {
               : {}),
             key: path,
             field: path,
-            onValueChange: ({ sqon }) => {
+            onValueChange: ({ sqon, value }) => {
+              console.log(value)
```
jgnieuwhof: this still WIP?

modules/components/src/AdvancedFacetView/FacetView.js
```diff
@@ -98,13 +98,13 @@ export default class FacetView extends React.Component {
...
             field: path,
-            onValueChange: ({ sqon }) => {
+            onValueChange: ({ sqon, value }) => {
               this.setState(
                 {
                   focusedPath: path,
                 },
                 () => {
-                  onValueChange({ sqon });
+                  onValueChange({ sqon, value });
```
jgnieuwhof: this should probably just wrap altogether, aka
```
onValueChange: x => this.setState({ focusedPath: path }, () => onValueChange(x))
```

jgnieuwhof: that way this type of maintenance doesn't need to happen again

modules/components/src/AdvancedFacetView/index.js
```diff
@@ -89,13 +89,13 @@ export default class AdvancedFacetView extends React.Component {
...
 
-  setSearchTerm = debounce(
-    value =>
-      this.setState({
-        searchTerm: value,
-      }),
-    500,
-  );
+  setSearchTerm = debounce(value => {
+    const { onFilterChange = () => {} } = this.props;
```
jgnieuwhof: we have conditional chaining in arranger, so could do `onFilterChange?.(value)` without having to set the default

modules/components/src/AdvancedFacetView/index.js
```diff
@@ -115,6 +116,7 @@ export default class AdvancedFacetView extends React.Component {
     } = this.state;
     const scrollFacetViewToPath = path => {
       this.facetView.scrollToPath({ path });
+      onFacetNavigation(path);
```
jgnieuwhof: same here w/ conditional chaining, if ya feel like it

PR #385, opened 2018-05-15
🐛fixes the pagination page index

hlminh2000:
This just rounds the page index down to an integer


alex-wilmer:
maybe we could extract some of this pagination math into its own function and write some unit tests for it. that makes me think this math must exist somewhere already

alex-wilmer:
maybe https://www.npmjs.com/package/ultimate-pagination

PR #387, opened 2018-05-15
🐛 fixes bad reduce

hlminh2000:


alex-wilmer:
I don't understand this change.. are you just trying to make sure the values is `null` instead of an empty array? if so I think doing that downstream is better

alex-wilmer:
or just passing `null` as the default argument to `reduce` is exactly the same I think?

hlminh2000:
so I totally thought it would give `undefined`, but apparently reduce just wouldn't execute on an empty array without an initial value. And yup I realized just now that you can pass null and it'll be fine... weird js quirk

screen shot 2018-05-15 at 5 39 38 pm

alex-wilmer:
not sure I've ever seen that runtime error before!

hlminh2000:
@alex-wilmer @jgnieuwhof, tested with kf portal, it worked.

PR #388, opened 2018-05-15
Table: allow checkbox off

cy:
need this for hcmi variant tables


alex-wilmer:
`in`?!?! I was so confused reading this at first.

alex-wilmer:
@Jephuff where do you stand on this?

cy:
is the `in` part strange, I could use `defaultProps` instead? Trying not to break existing kidsfirst stuff

Jephuff:
I would go with defaultProps :D 
Though... long term a better default might be no checkboxes. But thats not worth it now haha

cy:
Thanks @Jephuff ! @alex-wilmer - i've updated with `defaultProps`, which i forgot about for a sec :sweat_smile:

PR #389, opened 2018-05-16
sanitize ES type names on input

bdolly:
simply converts `-` in type names to `_` because graphQL schemas don’t like hyphens in type names

### DIFF
--- a/modules/components/src/Admin/Dashboard.js
+++ b/modules/components/src/Admin/Dashboard.js
@@ -371,7 +371,7 @@ class Dashboard extends React.Component {
       body: {
         eshost: this.state.eshost,
         index: this.state.newTypeIndex,
-        name: this.state.newTypeName,
+        name: this.state.newTypeAlias || this.state.newTypeName,
       },
     });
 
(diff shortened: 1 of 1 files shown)


hlminh2000:
I feel it's cleaner to just apply the transformation directly on `newTypeName` rather than bringing in a new state which brings more noise to this already gigantic file :P `newTypeName` would not be persisted anywhere otherwise anyway. That said, we should take some time to refactor this file... I'll have that on my bucket list. What do you think @alex-wilmer

alex-wilmer:
I think I got confused by `alias` in the variable name. Is that field just for storing the string.replaced value? 

`alias` is already such an "es" based term we should keep that reserved to mean es alias

alex-wilmer:
As far as "refactoring" is concerned. If you have as solid plan that you can execute quickly then by all means. It can be just "cleaned up" pretty quickly by moving large sections into their own files.

hlminh2000:
@alex-wilmer Yeah looks like the `newTypeAlias ` is just for holding the string.replaced value, which is what actually gets submitted to server. Visually it's still showing `newTypeName` so there's kind of a mismatch between visual rendering and what's going on underneath... hence my recommendation to just apply the string replace to `newTypeName`.

As for "refactoring", yeah I pretty much just mean breaking it up :P Which I've already done a little bit earlier for the table and aggs ordering stuff

PR #391, opened 2018-05-16
makes match box vertically responsive

hlminh2000:
this is needed for https://github.com/kids-first/kf-portal-ui/pull/386

### DIFF
--- a/modules/components/public/themeStyles/beagle/beagle.css
+++ b/modules/components/public/themeStyles/beagle/beagle.css
@@ -430,15 +430,37 @@ div {
 /**************************************/
 /******* styling for match box  *******/
 /**************************************/
+.match-box {
+  display: flex;
+  flex-direction: column;
+}
 .match-box-id-form > * {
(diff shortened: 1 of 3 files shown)


modules/components/public/themeStyles/beagle/beagle.css
```diff
@@ -430,16 +430,38 @@ div {
...
+.match-box .tabs .tabs-content {
+  display: flex;
+  flex-direction: column;
+}
+.match-box .tabs .tabs-titles {
+  display: block;
+}
+.match-box .tabs .tabs-titles .tabs-title {
+  float: left;
+}
```
jgnieuwhof: aside from the margin change I'm not sure if these changes are beagle specific, kind of seems like a global match-box layout thing. But I could be wrong (maybe it is specific), what do you think?

hlminh2000: yeah we can move this into matchbox itself

PR #392, opened 2018-05-16
🐛 JS can't handle ES's max long special val, give them a str ver

jgnieuwhof:
explanation here https://github.com/elastic/elasticsearch-js/issues/662

this is a stopgap until ES-JS sol'n exists

### DIFF
--- a/modules/mapping-utils/src/resolveHits.js
+++ b/modules/mapping-utils/src/resolveHits.js
@@ -1,5 +1,5 @@
 import getFields from 'graphql-fields';
-import { buildQuery } from '@arranger/middleware';
+import { buildQuery, CONSTANTS as ES_CONSTANTS } from '@arranger/middleware';
 
 let joinParent = (parent, field) => (parent ? `${parent}.${field}` : field);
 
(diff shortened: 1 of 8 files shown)


modules/mapping-utils/src/resolveHits.js
```diff
@@ -98,7 +98,13 @@ export default type => async (
     let source = x._source;
     let nested_nodes = resolveNested({ node: source, nestedFields });
     return {
-      searchAfter: x.sort || [],
+      searchAfter:
+        x.sort?.map(
+          x =>
+            Number.isInteger(x) && !Number.isSafeInteger(x)
+              ? `-9223372036854775808`
```
hlminh2000: it could be worth making a constant `ES_MAX_INT`, partly so we can reference it again if needed but also just to help future consumers of the code with the semantic :P

jgnieuwhof: yeah good call... I was just lazy and didn't want to have to deal w/ the cross-dep between server & middleware to get those constants here lol

Jephuff: might also be good to add a `// TODO: ` with a link to that issue :)

jgnieuwhof: 👋  yeahhh good call

PR #395, opened 2018-05-17
✨ added ability to pass in gql middleware

jgnieuwhof:
adds ability to pass in gql middleware from parent app


jgnieuwhof:
@alex-wilmer - want to get your opinion on this. netlify was having real issues with babel version inconsistencies, so I locked all module's babel versions to beta 40, which solved the problem, but now I'm worried about so many package-lock changes. I don't know, is it something to worry about? Is there a different sol'n or is this alright?

modules/schema/src/index.js
```diff
@@ -71,5 +74,5 @@ export default ({
     });
   }
 
-  return schema;
+  return middleware?.length ? applyMiddleware(schema, ...middleware) : schema;
```
hlminh2000: tiny thing but does it fail with an empty array? if not we can probably remove this check, `middleware` is defaulted to an array anyway

jgnieuwhof: lol you're right, just tested w/ empty array, will update

jgnieuwhof:
@hlminh2000 - addressed :)

PR #396, opened 2018-05-18
🐛 Bug/sets and sqon

hlminh2000:
basically just resolving sets into actual ids in sqon before feeding it in to build query

### DIFF
--- a/modules/mapping-utils/src/resolveAggregations.js
+++ b/modules/mapping-utils/src/resolveAggregations.js
@@ -1,5 +1,5 @@
 import getFields from 'graphql-fields';
-import { flattenDeep } from 'lodash';
+import { flattenDeep, isArray } from 'lodash';
 
 import {
   buildQuery,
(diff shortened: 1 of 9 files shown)


modules/mapping-utils/src/resolveAggregations.js
```diff
@@ -7,6 +10,69 @@ import {
...
+
+const resolveSetsInSqon = ({ sqon, es }) => {
+  const setIds = getSetIdsFromSqon(sqon || {});
+  return setIds.length
+    ? Promise.all(setIds.map(resolveSetIdsFromEs(es))).then(searchResult => {
+        const setIdsToValueMap = zipObject(
+          setIds.map(id => `set_id:${id}`),
+          searchResult,
+        );
+        console.log('setIdsToValueMap: ', setIdsToValueMap);
```
jgnieuwhof: :fire:

PR #397, opened 2018-05-18
✨ need graphql context options as well

jgnieuwhof:
so my last middleware PR wasn't enough, we need to be able to populate graphql context as well, in order to have access to the user information in the resolver middleware.

I've tested this locally, am 90% sure this gives us what we'll need to properly protect resolvers from KF-arranger

### DIFF
--- a/modules/server/src/projects/index.js
+++ b/modules/server/src/projects/index.js
@@ -13,7 +13,7 @@ import addProject from './addProject';
 import getProjects from './getProjects';
 import updateField from './updateField';
 
-export default ({ io, graphqlMiddleware }) => {
+export default ({ io, graphqlOptions }) => {
   const router = express.Router();
   // create es client
   router.use('/', async (req, res, next) => {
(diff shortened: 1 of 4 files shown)


alex-wilmer:
how would this extra context be used at the actual resolver level?

jgnieuwhof:
good q. the user's ego token is in the request headers, so the consumer could do something like this:
```
app.use('/graphql', graphqlHTTP(async (request, response, graphQLParams) => ({
  schema: MyGraphQLSchema,
  rootValue: await someFunctionToGetRootValue(request),
  graphiql: true
})));
```

jgnieuwhof:
good q. in this case it's to support resolver level authentication in kf-arranger. we need to populate the graphql context with some information about the user's permission levels that's found in the ego token in the request headers. so something like the following:
```
  Arranger({
    io,
    graphqlOptions: {
      middleware: [...], // something that would use what's placed in context
      context: async request => {
         return { ... };
      }
    }
app.use('/graphql', graphqlHTTP(async (request, response, graphQLParams) => ({
  schema: MyGraphQLSchema,
  rootValue: await someFunctionToGetRootValue(request),
  graphiql: true
})));

**Comment by jgnieuwhof** [comment]
Oh yeah good point, just realized there's nothing in terms of documentation here. There's an example on the `graphql-middleware` page here: https://github.com/prismagraphql/graphql-middleware.

```
const myMiddleware = {
  Mutation: async (resolve, parent, args, context, info) => {
    if (userInvalid(...)) {
      throw new Error(...)
    }
    resolve(parent, args, context, info);
  },
}
```

alex-wilmer:
ohhhh that's awesome! I didn't realize we were using that. resolver level before after hooks sounds incredible

jgnieuwhof:
yeah! so this resolver protection, and then the changes here: https://github.com/overture-stack/ego-token-middleware/pull/10/files ensure that we can completely seal off the arranger admin api (rest + graphql) from non-admin's

PR #399, opened 2018-05-22
🐛 initializes query loading state based on shouldFetch

hlminh2000:


modules/components/src/Query.js
```diff
@@ -5,7 +5,11 @@ import defaultApi from './utils/api';
 import { defaultProps } from 'recompose';
 
 class Query extends Component {
-  state = { data: null, error: null, loading: false };
+  constructor(props) {
+    super(props);
+    const { shouldFetch } = props;
+    this.state = { data: null, error: null, loading: shouldFetch && true };
+  }
```
jgnieuwhof: since `shouldFetch` is a bool I don't think we need the `&&`. also I think `this.props` is available as follows:
```
class Query extends Component {
  state = { data: null, error: null, loading: this.props.shouldFetch };
  ...
}
```

hlminh2000:
@jgnieuwhof  WUUTTTT??!!! I swear I tried that state initialization syntax once upon a time and it didn't work, but now it does! My world just flipped...

Anyway, addressed! :P

PR #402, opened 2018-05-25
💄 move layout agg bar width to KF portal

jgnieuwhof:
moved this into the KF repo


hlminh2000:
does this break any of the stories?

jgnieuwhof:
@hlminh2000 - not unless you like half your portal to be the agg sidebar ;) fixed

PR #408, opened 2018-05-29
✨ if you just want the next sqon...

jgnieuwhof:
give the parent app a little more flexibility for what to do with the sqon result

### DIFF
--- a/modules/components/src/Arranger/MatchBox.js
+++ b/modules/components/src/Arranger/MatchBox.js
@@ -245,28 +245,23 @@ const MatchBox = ({
                       }),
                       dataPath,
                     );
-                    if (setSQON) {
-                      setSQON(
-                        toggleSQON(
+                    const nextSQON = toggleSQON(
+                      {
+                        op: 'and',
+                        content: [
                           {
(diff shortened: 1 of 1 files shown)


modules/components/src/Arranger/MatchBox.js
```diff
@@ -245,28 +245,23 @@ const MatchBox = ({
...
-                                  value: [].concat(
-                                    `set_id:${data.setId}` || [],
-                                  ),
-                                },
-                              },
-                            ],
+                            op: 'in',
+                            content: {
+                              field: primaryKeyField?.field,
+                              value: [].concat(`set_id:${data.setId}` || []),
```
hlminh2000: would ``set_id:${data.setId}` || []` ever evaluate to `[]`? if `data.setId` is the null check then maybe `data.setId ? `set_id:${data.setId}` : []` is a safer check

hlminh2000: `[...(data.setId ? ['set_id:${data.setId}'] : [])]` I guess

hlminh2000:
👍

PR #409, opened 2018-05-31
✨import/export arranger config

jgnieuwhof:
Adds the ability to export (download) an arranger project config, and import (upload) config when adding a type.

Export comes in a tarball, with one subdir per type. Each type includes the `extended`, `aggs-state`, `columns-state`, and `matchbox-state` json config files.

When adding a type you can optionally upload config files, which are stored with the type in the arranger version meta. The spinup flow hasn't changed - extended fields are initialized in `addType`, the other states are initialized on spinup.

Import treats the generated mapping as the source of truth, and just overwrites fields that have a config found in one of the uploaded config files, if they exist.

### DIFF
--- a/modules/components/src/utils/parseInputFiles.js
+++ b/modules/components/src/utils/parseInputFiles.js
@@ -1,0 +1,13 @@
+export default ({ files }) =>
+  Promise.all(
+    [...files].map(
+      f =>
+        new Promise((resolve, reject) => {
+          const reader = new FileReader();
+          reader.onload = () =>
+            resolve({ name: f.name, content: reader.result });
+          reader.onerror = e => reject(e);
+          reader.readAsText(f);
+        }),
+    ),
+  );
--- a/modules/server/src/projects/addType.js
+++ b/modules/server/src/projects/addType.js
@@ -1,13 +1,13 @@
-import { flattenDeep } from 'lodash';
 import { extendMapping } from '@arranger/mapping-utils';
 import { fetchMappings } from '../utils/fetchMappings';
 import mapHits from '../utils/mapHits';
 import getIndexPrefix from '../utils/getIndexPrefix';
+import initializeExtendedFields from '../utils/initializeExtendedFields';
 
 export default async (req, res) => {
   let { es } = req.context;
   let { id } = req.params;
-  let { index, name } = req.body;
+  let { index, name, config: rawConfig = {} } = req.body;
 
   const esType =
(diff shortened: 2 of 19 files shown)


alex-wilmer:
This is really cool. Haven't finished going through it all, but excellent work

modules/components/src/utils/parseInputFiles.js
```diff
@@ -0,0 +1,13 @@
+export default ({ files }) =>
+  Promise.all(
+    [...files].map(
```
hlminh2000: if `[...files]` is meant to catch nulls, maybe `(files || [])`... `[...null]` would throw an error anyway

modules/server/src/projects/addType.js
```diff
@@ -82,21 +91,9 @@ export default async (req, res) => {
...
-            },
-          },
-          JSON.stringify(x),
-        ]),
-      );
-
-      await es.bulk({ body });
+      await initializeExtendedFields({ indexPrefix, config, fields, es });
     } catch (error) {
+      console.log(error);
```
hlminh2000: prefer `console.error` or 🔥

modules/server/src/utils/replaceBy.js
```diff
@@ -0,0 +1,5 @@
+export default (arr1, arr2, operator) => {
+  const cArr1 = arr1 || [];
+  const cArr2 = arr2 || [];
+  return cArr1.map(x => cArr2.find(y => operator(x, y)) || x);
+};
```
hlminh2000: 👍 this would make a good lodash function eh :P

modules/components/src/utils/parseInputFiles.js
jgnieuwhof: this is an interesting one, `files` isn't actually an array, it's an enumerable: https://developer.mozilla.org/en-US/docs/Web/API/FileList

enumerables are spreadable, so spreading into an array turns it into an array, and then you can map :)

modules/server/src/projects/addType.js
jgnieuwhof: 👍 good call

modules/components/src/utils/parseInputFiles.js
hlminh2000: oh I see! yeah totally makes sense then :)

jgnieuwhof:
@alex-wilmer - any further thoughts on this? might be nice to get this into the real world

PR #411, opened 2018-06-01
✨ adds ability to feed custom component props in Aggregations component

hlminh2000:


modules/components/src/Arranger/Aggregations.js
```diff
@@ -18,6 +18,12 @@ const Aggregations = ({
   api,
   Wrapper = BaseWrapper,
   containerRef,
+  componentProps = {
+    getTermAggProps: () => ({}),
+    getRangeAggProps: () => ({}),
+    getBooleanAggProps: () => ({}),
+    getDatesAggProps: () => ({}),
+  },
```
hlminh2000: just made these functions so in case we need to switch props logically based on different fields or something, then we can pass it in from within `aggComponentsMap`

modules/components/stories/Portal.js
```diff
@@ -131,7 +131,15 @@ const ChooseProject = ({ index, projectId, update, projects }) => {
...
   return (
     <div style={{ display: 'flex', ...style }}>
-      <Aggregations style={{ width: 300 }} {...props} />
+      <Aggregations
+        style={{ width: 300 }}
+        componentProps={{
+          getTermAggProps: () => ({
+            maxTerms: 3,
+          }),
+        }}
```
hlminh2000: example usage :)

PR #412, opened 2018-06-04
💄 updates filter icon

hlminh2000:
just changing the default icon for all filter inputs


jgnieuwhof:
@alex-wilmer - yeah we're overriding the input element in KF anyway. just thought this was a better default :P

hlminh2000:
yeah this pr isn't really necessary anymore but it's still just a better default anyway :P

PR #416, opened 2018-06-14
Quicksearch handle copy_to, allow autocomplete

cy:
HCMI quicksearch wants the quicksearch input to autocomplete on multiple fields, to do this we changed the mapping to 'copy_to' desired fields to a field with elasticsearch autocomplete analysers.

This is still kind of WIP but I've changed the code quite a bit and I don't want to break kidsfirst :sweat_smile:  With copy_to fields, the same field https://github.com/nci-hcmi-catalog/portal/pull/103/files#diff-992d63ffc3841f5e20a65aab011f1f18R5 `autocomplete` can be use to search model.name, variants.genes and variants.name, but the `autocomplete` field does not appear in _source unlike normal fields, so arranger couldn't return any data although elasticsearch found the nodes. I changed columnsState to save a query to the source fields if it has been copied to, and in the QuickSearchQuery, recursively find the values matching searchTextParts instead of jp.query for the field.

### DIFF
--- a/modules/components/src/Arranger/QuickSearch/QuickSearch.js
+++ b/modules/components/src/Arranger/QuickSearch/QuickSearch.js
@@ -43,12 +43,12 @@ const QuickSearch = ({
   value,
   setValue,
   placeholder = 'Quick Search',
+  searchTextDelimiters = ['\\s', ','],
   Icon = <SearchIcon />,
   LoadingIcon = <SearchIcon />,
   PinnedValueComponent = SQONBubble,
   translateSQONValue = x => x,
   InputComponent = TextInput,
-  renderFound,
   ...props
 }) => (
   <QuickSearchFieldsQuery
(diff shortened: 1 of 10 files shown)


cy:
image
OK, more confident in this now; seems to be working. Also tested it with normal non-copy-to fields like before and direct matches still work so probably wouldn't break kidsfirst.

modules/components/src/Arranger/QuickSearch/QuickSearchQuery.js
```diff
@@ -1,10 +1,30 @@
...
+import { capitalize, flatMap, isArray, isObject } from 'lodash';
 import { compose, withProps } from 'recompose';
 import jp from 'jsonpath/jsonpath.min';
 
 import { withQuery } from '../../Query';
 import splitString from '../../utils/splitString';
 
+export const findMatchingValues = ({ item, searchText }) => {
+  let value;
+  Object.keys(item || {}).some(function(k) {
```
hlminh2000: `Object.entries` or `_.toPairs` might help this read a little more easily :)

modules/components/src/Arranger/QuickSearch/QuickSearch.js
```diff
@@ -66,8 +67,11 @@ const QuickSearch = ({
         {...props}
         {...{ primaryKeyField, quickSearchFields }}
         searchText={value}
+        searchTextDelimiters={searchTextDelimiters}
         render={({ results: searchResults, loading }) => (
           <div className={`quick-search ${className}`}>
+            {console.log('-------------')}
+            {console.log(searchResults)}
```
jgnieuwhof: :fire:

cy:
thanks for the comments! yeah I think it'll make sense to resolve the copy_to sourceFields as a json obj in resolveHits, then the copy_to field wouldn't be null in the graphql api. Going to try that out now!

cy:
replaced by #421

PR #434, opened 2018-07-12
AdvancedFacetView props propagation

bdolly:
simple passing of props up to the AdvancedFacetView Component. These are event props mostly used as hooks for analytics events

props propagated:
- onTermSelected
- onClear
-onFacetNavigation
- onFilterChange

### DIFF
--- a/modules/components/src/AdvancedFacetView/FacetView.js
+++ b/modules/components/src/AdvancedFacetView/FacetView.js
@@ -9,15 +9,15 @@ const flattenDisplayTreeData = displayTreeData => {
   return displayTreeData.reduce(
     (acc, node) => [
       ...acc,
-      ...(node.children ? flattenDisplayTreeData(node.children) : [node])
+      ...(node.children ? flattenDisplayTreeData(node.children) : [node]),
     ],
-    []
+    [],
   );
 };
 
 export default class FacetView extends React.Component {
   state = {
-    focusedPath: null
+    focusedPath: null,
   };
   scrollToPath = ({ path, behavior = 'smooth', block = 'start' }) => {
     const targetElementId = serializeToDomId(path);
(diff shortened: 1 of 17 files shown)


modules/components/src/Arranger/Aggregations.js
```diff
@@ -8,7 +8,7 @@ const BaseWrapper = ({ className, ...props }) => (
 );
 
 const Aggregations = ({
-  onTermSelected = () => {},
+  onTermSelected,
```
hlminh2000: this prop can just be called `onValueChange`, the `onValueChange` passed into `aggComponents` isn't differentiating between types.

modules/components/src/Arranger/CurrentSQON.js
```diff
@@ -40,7 +40,7 @@ export const CurrentSQON = ({
       <Bubble
         className="sqon-clear"
         onClick={() => {
-          onClear();
+          onClear?.();
```
hlminh2000: I actually kinda prefer the default props just for consistency here :P

modules/components/src/AdvancedFacetView/index.js
```diff
@@ -111,13 +111,15 @@ export default class AdvancedFacetView extends React.Component {
...
+      onFacetNavigation,
+      onTermSelected,
+      onClear,
       InputComponent = TextInput,
       ...props
     } = this.props;
     const scrollFacetViewToPath = path => {
       this.facetView.scrollToPath({ path });
-      onFacetNavigation(path);
+      onFacetNavigation?.(path);
```
hlminh2000: I actually kinda prefer the default props for consistency and also in case it needs to be called at multiple places, we don't forget the check.

modules/components/src/AdvancedFacetView/FacetView.js
```diff
@@ -105,6 +106,7 @@ export default class FacetView extends React.Component {
                 },
                 () => {
                   onValueChange({ sqon, value });
+                  onTermSelected?.(value);
```
hlminh2000: why not use the existing `onValueChange`?

bdolly: onTermSelected propagates up from the facet itself from an earlier PR I made #375

PR #438, opened 2018-07-18
⚡️ hits resolution performance boost

hlminh2000:
**What was done**
Due to a large amount of data tranformation happening in the hits resolver (`modules/mapping-utils/src/resolveHits.js`), file download is extremely slow. To address this, the following changes were made:

1) A slight modification to the resolveNested function. The original had a very mutation-free approach that creates a new accumulator on every iteration, which implied an iteration over previously accumulated results. Removing this operation significantly improves the performance (consider this benchmark: https://www.measurethat.net/Benchmarks/ShowResult/27779). Only ephemeral mutation is added for this, so all existing tests passed. However, to be extra confident, I would like to get some tests from @cy written to assure accuracy when the copy_to field is involved.

2) Parallelizing the processing of hits. Fortunately, the mapping operations were independent of each other, so we were are able to **perform this operation in multiple parallel processes and combine the output** at the end. As there are overheads associated with running too many processes, this "ludicrous mode" is **only triggered if there are more than 1000 entries**, otherwise only one new process is spawn. The **number of parallel process depends on an environment variable for # of cores** available in the system, with a **default fallback to 2**. With two processes available, we also **increased the default stream buffer size from 1000 to 2000** to increase the gains in processing power and decrease other overheads.

**Benchmark result**
On my local machine, file download time of 15,838 entries (with some nested data) was reduced from 91687ms (1.5 minutes) to 37812ms (37 seconds). Total hits resolution compute time was reduced from 1 minute to 10 seconds.

The remaining time is attributed to the tsv generation and awaiting data from elasticsearch.

### DIFF
--- a/modules/mapping-utils/src/resolveHits.js
+++ b/modules/mapping-utils/src/resolveHits.js
@@ -102,7 +102,7 @@ export const hitsToEdges = ({
                       nestedFields,
                       parent: fullPath,
                     });
-              return Object.assign({}, acc, resolvedNested);
+              return Object.assign(acc, resolvedNested);
             }, {});
           };
           let source = x._source;
@@ -120,8 +120,8 @@ export const hitsToEdges = ({
                 )
               : [],
             node: Object.assign(
-              { id: x._id },
               source,
+              { id: x._id },
               nested_nodes,
               copied_to_nodes,
             ),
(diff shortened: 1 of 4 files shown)


alex-wilmer:
> "ludicrous mode"

🤣 

Anyways, amazing work 💯

PR #456, opened 2018-10-16
Feature/809 boolean agg disabled

ciaranschutte:


modules/components/src/Aggs/BooleanAgg.js
```diff
@@ -54,6 +54,11 @@ export default ({
     field: dotField,
   });
 
+  const isTrueBucketDisabled =
+    trueBucket === undefined || trueBucket?.doc_count <= 0;
+  const isFalseBucketDisabled =
+    falseBucket === undefined || falseBucket?.doc_count <= 0;
```
hlminh2000: Correct me if I'm wrong :P

I think we just need the `trueBucket?.doc_count <= 0` and `falseBucket?.doc_count <= 0` case, since if `trueBucket === undefined`, then the latter will always be `true`

modules/components/stories/Aggs.js
```diff
@@ -57,6 +57,46 @@ const bolleanAggs = [
...
+        doc_count: 2312,
+        key_as_string: 'false',
+      },
+      {
+        key: '1',
+        doc_count: 0,
+        key_as_string: 'true',
+      },
+    ],
+  },
```
hlminh2000: 👍

PR #467, opened 2018-11-01
✨ Feat/admin api

hlminh2000:


modules/mapping-utils/src/createConnectionResolvers.js
```diff
@@ -3,28 +3,36 @@ import { get } from 'lodash';
...
     },
-    extended: (obj, { fields }) => {
+    extended: async (obj, { fields }, { es, projectId }) => {
+      const { index } = type;
+      const extendedFields = await loadExtendedFields({ es, projectId, index });
       return fields
-        ? type.extendedFields.filter(x => fields.includes(x.field))
-        : type.extendedFields;
+        ? extendedFields.filter(x => fields.includes(x.field))
+        : extendedFields;
```
hlminh2000: this removes the in memory caching of mapping and extended mapping of the search api

PR #477, opened 2019-01-09
#820 Add # Files to Facet UI

nyanofthemoon:


modules/components/src/Aggs/TermAgg.js
```diff
@@ -99,7 +99,7 @@ const enhance = compose(
 const TermAgg = ({
   field = '',
   displayName = 'Unnamed Field',
-  headerTitle = '# Files',
+  headerTitle = null,
```
hlminh2000: this will need to be passed down from `modules/components/src/Arranger/Aggregations.js` for `kf-portal-ui` to be able to access

nyanofthemoon: @hlminh2000 Ok - this is part of another PR in kids-first-ui though- can we merge this one since it is "optional"?

hlminh2000: why not just do it here? that'd be 2 more PR, one in this repo, one in kf-portal-ui, so why don't we just do all the arranger stuff needed in 1 PR?

nyanofthemoon: Oh, I thought you meant in KF- my bad.

PR #478, opened 2019-01-10
Adds  ability to override cols with custom col based on field

ciaranschutte:


ciaranschutte:
https://github.com/kids-first/kf-portal-ui/pull/1061

modules/components/src/DataTable/utils.js
```diff
@@ -20,18 +20,33 @@ export function normalizeColumns({
...
-        : column.hasCustomType,
-      ...(!column.accessor && !column.id ? { id: column.field } : {}),
-      ...(customTypeConfigs[column.type] || {}),
-    }))
+    .map(column => {
+      const customColIndex = customColumns.findIndex(
+        cc => cc.content.field === column.field,
+      );
+
+      let customCol = {};
```
hlminh2000: I think we can achieve this with:
```
const customCol = customColumns.find(cc => cc.content.field === column.field)?.content || {}
```
Then, I don't think there's a need for the `if` bellow, but please correct me if I'm wrong :P

modules/components/src/DataTable/utils.js
```diff
@@ -20,18 +20,33 @@ export function normalizeColumns({
...
+      if (customColIndex > -1) {
+        customCol = customColumns[customColIndex].content;
+        customColumns.splice(customColIndex, 1);
```
hlminh2000: This mutation feels troubling to me, especially in this case as `customColumns` is then used outside of this map. It's better imo to compute a new filtered array outside the map for use later.

ciaranschutte: So before this mapping create two arrays "custom columns to use overriding" and a "custom columns to add"?

hlminh2000: I think after makes more sense, a const that's the filtered `customColumns` to feed in here https://github.com/overture-stack/arranger/pull/478/files/aaf642fcd7a51a475ff507fd9a9484c96126e6de#diff-f0ae8de5b156ad4c703b2c127b034abcR50

modules/components/src/DataTable/utils.js
ciaranschutte: yup tested :)

modules/components/src/DataTable/utils.js
```diff
@@ -20,19 +20,33 @@ export function normalizeColumns({
...
+          : column.hasCustomType,
+        ...(!column.accessor && !column.id ? { id: column.field } : {}),
+        ...(customTypeConfigs[column.type] || {}),
+        ...customCol,
+      };
+    })
     .filter(x => x.show || x.canChangeShow);
-  return sortBy(customColumns, 'index').reduce(
+
+  // filter out override columns
```
hlminh2000: this comment will really help in future! thanks :)

PR #511, opened 2019-04-05
Add color to selected row

jaouad-benassila:


modules/components/src/DataTable/Table/Table.js
```diff
@@ -208,6 +208,17 @@ class DataTable extends React.Component {
...
+            if (rowInfo && rowInfo.row) {
+              return {
+                style: {
+                  background: this.isSelected(rowInfo.original.id) ? '#edf9fe' : '',
+                },
+              };
+            } else {
+              return {};
+            }
+          }}
```
hlminh2000: This is a very usage specific thing, we can support this by extending this component to receive a `reactTableProps` prop and provide this logic externally, as a way to expose the full react-table API to kids first. Will be a good thing overall forward so little project-specific UI updates does not have to go into arranger source :)

jaouad-benassila:
Where do you want to put it exactly ?

modules/components/src/DataTable/Table/Table.js
jberube: Would passing down a prop called `selectedRowColor` in the Table ok for you as a solution?

Not passing down a color would disable selected column highlighting. Or we could default to a color of you choice.

PR #512, opened 2019-04-05
select all in TermFilter should only select filtered results

jberube:
https://github.com/kids-first/kf-portal-ui/issues/1551

### DIFF
--- a/modules/components/src/AdvancedSqonBuilder/filterComponents/TermFilter.js
+++ b/modules/components/src/AdvancedSqonBuilder/filterComponents/TermFilter.js
@@ -23,6 +23,13 @@ const AggsWrapper = ({ children }) => (
   <div className="aggregation-card">{children}</div>
 );
 
+const filterStringsCaseInsensitive = (values, search, path = null) => {
+  const normalizedSearch = search.toLowerCase();
+  return values.filter(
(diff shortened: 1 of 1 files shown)


modules/components/src/AdvancedSqonBuilder/filterComponents/TermFilter.js
```diff
@@ -89,7 +89,9 @@ export const TermFilterUI = props => {
         ...currentFieldSqon,
         content: {
           ...currentFieldSqon.content,
-          value: buckets.map(({ key }) => key),
+          value: buckets
+            .filter(({ key }) => key.includes(s.state.searchString))
```
hlminh2000: there's a new PR recently to apply case insensitive search, maybe split the filter logic from that commit to a common function and reuse here? :) https://github.com/overture-stack/arranger/commit/098e2a95fc3780205dcb610a6ced422e0ba354fd

jberube: Good point. The same logic is also in the SearchAll filter, and could use that same common function. I'll make that, even if it's out of the scope of this PR.

jberube: https://github.com/overture-stack/arranger/pull/510

PR #513, opened 2019-04-08
Add tooltip to the queries

jaouad-benassila:


jaouad-benassila:
you need to install it:
npm install --save react-tippy

PR #519, opened 2019-06-11
Feat/1688 Number of participants in queries of the Virtual Study

jberube:
Adds the number of participants in each query of the virtual study.

Related to https://github.com/kids-first/kf-portal-ui/issues/1688


aalex:
@jberube We have forked Arranger, so this PR should rather be in https://github.com/kids-first/arranger

jberube:
Sorry, my bad, I didn`t target the good repo. Closing.

PR #572, opened 2020-04-13
export functions to create schema without server

jaserud:
Exported the functions used to create a project schema, so the executable schema can be used else where
List of chagnes:
- exports functions
- update package-lock

### DIFF
--- a/modules/server/src/startProject.js
+++ b/modules/server/src/startProject.js
@@ -41,7 +41,7 @@ const getTypesWithMappings = async ({ es, id }) => {
   if (!mappings.length) return; // gate to not start a project that doesn't exist
 
   const extended = await Promise.all(
-    hits.map(async (type) => {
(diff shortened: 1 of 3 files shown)


modules/server/src/startProject.js
```diff
@@ -58,6 +127,24 @@ export const createProjectEndpoint = async ({
...
+};
+
+export const createProjectEndpoint = async ({
+  es,
+  id,
+  graphqlOptions = {},
+  enableAdmin,
+  typesWithMappings,
+}) => {
+  const { schema, mockSchema } = createProjectSchema({
```
jaserud, line 140: @hlminh2000 This is the function that I've exported for argo gateway to use. Its literally everything before the apollo-router is created.

jaserud:
Never mind looks like tests are failing :disappointed: will take a look

PR #585, opened 2020-06-11
Fix Junk Indices issue

joneubank:
Prevent server from generating an empty index when `getTypes` is called. We will first check if a project ID exists in `arranger-projects` before initializing the index.

Includes some dev-dependency updates that were needed to run the server in 'watch' mode.


hlminh2000:
except for the failed tests...

hlminh2000:
looks like an npm package issue:
```
Binary has a problem: Error: Error relocating /app/node_modules/node-sass/vendor/linux_musl-x64-79/binding.node: _ZNSt7__cxx1118basic_stringstreamIcSt11char_traitsIcESaIcEEC1Ev: symbol not found

    at Object.Module._extensions..node (internal/modules/cjs/loader.js:1173:18)

    at Module.load (internal/modules/cjs/loader.js:972:32)

    at Function.Module._load (internal/modules/cjs/loader.js:872:14)

    at Module.require (internal/modules/cjs/loader.js:1012:19)

    at require (internal/modules/cjs/helpers.js:72:18)

    at module.exports (/app/node_modules/node-sass/lib/binding.js:19:10)

    at Object.<anonymous> (/app/node_modules/node-sass/lib/index.js:14:35)

    at Module._compile (internal/modules/cjs/loader.js:1123:30)

    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1143:10)

    at Module.load (internal/modules/cjs/loader.js:972:32)

Building the binary locally

Building: /usr/bin/node /app/node_modules/node-gyp/bin/node-gyp.js rebuild --verbose --libsass_ext= --libsass_cflags= --libsass_ldflags= --libsass_library=
```
https://jenkins.qa.cancercollaboratory.org/blue/organizations/jenkins/Overture.bio%2Farranger/detail/feat%2Fjunk-indices/2/pipeline

joneubank:
This solution isn't complete. The creation of the index is done by `getTypes`, which has other callers. To fix this I need to move the check for existing project into the getTypes method.

joneubank:
The complete solution is to understand the specific scenario where we need to automatically create the index and let that situation create it BEFORE calling getTypes, but in lieu of that we can move this check to getTypes. Doing this check there is annoying, but since it only comes up in cases where the caller is inputting the project wrong there should be no performance cost when calling the correct endpoints.

PR #593, opened 2020-06-29
⚡️ 503 - calculate bucket count in aggs

anncatton:
Adds `bucket_count` field in Aggregations to calculate number of buckets when resolving agg query.

### DIFF
--- a/modules/middleware/src/flattenAggregations.js
+++ b/modules/middleware/src/flattenAggregations.js
@@ -18,9 +18,13 @@ function flattenAggregations({ aggregations, includeMissing = true }) {
         ...value.buckets,
         ...(includeMissing && missing ? [{ ...missing, key: MISSING }] : []),
       ];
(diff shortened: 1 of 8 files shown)


modules/middleware/src/flattenAggregations.js
```diff
@@ -18,9 +18,13 @@ function flattenAggregations({ aggregations, includeMissing = true }) {
         ...value.buckets,
         ...(includeMissing && missing ? [{ ...missing, key: MISSING }] : []),
       ];
+      const bucket_count = buckets.filter(bucket => bucket.doc_count > 0)
+        .length;
```
hlminh2000, line 22: do we really need this filter? other than `MISSING`, are there any other expected case of `doc_count <= 0` that can show up?

If we can avoid it that'd be best to reduce load on the main thread

hlminh2000:
I guess we'll have to publish this and bring into Argo gateway for it to be testable eh

PR #602, opened 2020-08-27
Typo

denis-yuen:


hlminh2000:
thank you @denis-yuen for the update!

PR #616, opened 2020-09-28
♿️ a11y fixes for HCMI

mistryrn:
* ♿️ Make input field autofocus optional
* ♿️ Add aria labels to table checkboxes
* ♿️ Set unique IDs for TermAgg labels
* ♿️ Add descriptive text for RangeAgg sliders


modules/components/src/DataTable/Table/EnhancedReactTable.js
```diff
@@ -1,6 +1,6 @@
 import ReactTable from 'react-table';
 import { compose } from 'recompose';
-import checkboxHOC from 'react-table/lib/hoc/selectTable';
+import checkboxHOC from './checkboxHOC';
```
hlminh2000, line 3: can we try something like this? if it works, we can avoid having to maintain a fork of third-party libs :)

```
import checkboxHOC from 'react-table/lib/hoc/selectTable';

const defaultSelectInputComponent = props => {
  <input
    type={props.selectType || 'checkbox'}
    checked={props.checked}
    aria-label={`${props.checked ? 'Deselect' : 'Select'} ${props.id ? 'this row' : 'all rows'}`}
    onClick={e => {
      const { shiftKey } = e;
      e.stopPropagation();
      props.onClick(props.id, shiftKey, props.row);
    }}
    onChange={() => {}}
  />;
}

const withDefaultSelectInputComponent = Component => {
  return props => (
    <Component SelectInputComponent={props.SelectInputComponent || defaultSelectInputComponent} />
  )
}

const enhance = compose(
  checkboxHOC,
  withDefaultSelectInputComponent,
);
```

hlminh2000 (commented):
Just an alternative idea to try out for the default prop :)

modules/components/src/DataTable/Table/EnhancedReactTable.js
mistryrn: This looks a lot better! Let me give it a try, thanks :)

hlminh2000 (approved):
🌮 🌮 🌮 🌮 🌮

modules/components/src/Aggs/RangeAgg.js
```diff
@@ -170,6 +179,18 @@ class RangeAgg extends Component {
...
               <RangeLabel isLeft>{this.formatRangeLabel(min)}</RangeLabel>
               <RangeLabel>{this.formatRangeLabel(max)}</RangeLabel>
+              <span
+                id={getLabelId(displayName)}
+                css={`
+                  position: absolute;
+                  height: 0;
+                  width: 0;
+                  top: -9999px;
+                  left: -9999px;
```
hlminh2000, line 189: @joneubank brought up a great point about potential weird scrolling behaviour. 

We don't really have any specific problems in mind, but before merging, can we get a quick check of this in Chrome, FF, Safari and Edge to make sure there's no strange side effect of this? :)

Just want to be extra careful in a shared library ;)

mistryrn: Just tested in all of those browsers on the Mac, haven't seen any funkiness, looks just like normal. No unexpected scrolling behaviour. Don't have access to a Windows machine/VM on my OICR laptop so I haven't been able to test Windows browsers.

However, if you want we can eliminate the visually hidden element altogether and just point the aria-labelledby at an existing element. It won't be as descriptive but it'll satisfy the scanner

mistryrn: Just completed re-testing using Windows via BrowserStack -- all good, no weird issues to report :)

hlminh2000: Awesome! thank you for the diligence man!

Issue #618, opened 2020-10-06
Add optional auth for elasticsearch clients


rtisma:
If the elasticsearch service requires authorization, a username and password will be needed by the ES client. Currently, there is no support for a secured ES cluster.

b-f-chan:
Consensus among team after consultation with @hlminh2000:

- Add ES_USERNAME, ES_PASSWORD as optional & backwards-compatible configs to Arranger
- This will take priority over DMS Installer work so we can unblock DMS-UI dev & testing, and finally resolve the Arranger issues

**Some notes for context:**

- ES has auth and we do not want to disable this as general security best practice
- Just for testing, @rtisma tried to disable ES auth to see if he could get Arranger working, but could not
- In ARGO, Arranger is not aware of ES_USERNAME, ES_PASSWORD, yet still works with ES auth enabled
- This is because for ARGO, Arranger is not deployed standalone but has the benefit of being instantiated as part of the platform gateway, with an ES client provided conveniently as a dependency in memory
- Arranger is standalone in DMS and used differently than in ARGO, hence we need to make this change
- The solution proposed above is the quickest and most practical way to get around this

Adding to Sprint 7.

b-f-chan:
Confirmed with @rtisma Arranger has been updated to handle ES auth settings, has been reviewed and released

PR #619, opened 2020-10-08
Rtisma es passwd

rtisma:
Related to #618 

added ES_USER and ES_PASS as OPTIONAL env variable. If both ES_USER and ES_PASS are not defined, then will work as it did before. If ES_USER is defined and ES_PASS is not, will log an error. If both are defined, then will use the username and password for the elasticsearch client

Also added test data (from genomic_data_playground) for the ES index.

### DIFF
--- a/modules/server/src/server.js
+++ b/modules/server/src/server.js
@@ -8,11 +8,11 @@ import startProject, { getDefaultServerSideFilter } from './startProject';
 import { ES_HOST, ES_USER, ES_PASS, ES_LOG, PROJECT_ID, MAX_LIVE_VERSIONS } from './utils/config';
 import { fetchProjects } from './projects/getProjects';
 
-export const buildClientViaEnv = () => {
-  return buildClient(ES_HOST, ES_USER, ES_PASS, ES_LOG)
+export const buildEsClientViaEnv = () => {
+  return buildEsClient(ES_HOST, ES_USER, ES_PASS, ES_LOG)
 }
 
-export const buildClient = (esHost, esUser, esPass, esLog) => {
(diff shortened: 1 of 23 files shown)


modules/server/src/server.js
```diff
@@ -5,9 +5,35 @@ import bodyParser from 'body-parser';
...
 import startProject, { getDefaultServerSideFilter } from './startProject';
-import { ES_HOST, PROJECT_ID, MAX_LIVE_VERSIONS } from './utils/config';
+import { ES_HOST, ES_USER, ES_PASS, ES_LOG, PROJECT_ID, MAX_LIVE_VERSIONS } from './utils/config';
 import { fetchProjects } from './projects/getProjects';
 
+export const buildClientViaEnv = () => {
+  return buildClient(ES_HOST, ES_USER, ES_PASS, ES_LOG)
+}
+
+export const buildClient = (esHost, esUser, esPass, esLog) => {
```
hlminh2000, line 15: should we just call these `buildEsClientViaEnv` and `buildClient`?

modules/server/src/server.js
```diff
@@ -5,9 +5,35 @@ import bodyParser from 'body-parser';
...
+
+  if (esUser){
+	  if (!esPass) {
+		  console.error('ES user was defined, but password was not');
+	  }
+	  esConfig['auth'] = {
+		  username: esUser,
+		  password: esPass
+	  }
+  }
```
hlminh2000, line 33: there's some inconsistent indentation here... we normally use [prettier](https://prettier.io/) to handle code formatting as a commit hook, but I guess we never installed it for this repo

hlminh2000 (approved):

PR #625, opened 2020-10-15
add configurable nginx path to dockerfile

rtisma:
this is neccessary, since in some k8s environment, the disk is readonly. This allows us to define an volitile and writable directory in k8s (emptyDir), and then configure the container render the template to that directory, and nginx to read the config from the directory


hlminh2000 (approved):
lol that was simple!

PR #627, opened 2020-10-20
add track total hits in hit resolver + schema

wajiha-oicr:
https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/7.x/api-reference.html#_search

Add parameter for `track_total_hits`, defaulted to true

### DIFF
--- a/modules/mapping-utils/src/createConnectionTypeDefs.js
+++ b/modules/mapping-utils/src/createConnectionTypeDefs.js
@@ -28,6 +28,7 @@ export default ({ type, fields = '', createStateTypeDefs = true }) => `
       first: Int
       last: Int
       searchAfter: JSON
+      trackTotalHits: Boolean
     ): ${type.name}Connection
 
     aggregations(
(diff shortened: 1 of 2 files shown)


modules/mapping-utils/src/createConnectionTypeDefs.js
```diff
@@ -28,6 +28,7 @@ export default ({ type, fields = '', createStateTypeDefs = true }) => `
       first: Int
       last: Int
       searchAfter: JSON
+      trackTotalHits: Boolean
```
hlminh2000, line 31: You can provide the default value here in the schema actually, this is actually preferred so the API consumer can see it :)

hlminh2000 (changes requested):
A small request :)

modules/mapping-utils/src/createConnectionTypeDefs.js
wajiha-oicr: added in https://github.com/overture-stack/arranger/pull/627/commits/5774e409d0fe578469e37798268b270b4bc9621b !

hlminh2000 (approved):

Issue #640, opened 2021-01-18
Update Column Selection Behaviour/UI


rosibaj:
Zeplins: https://zpl.io/a8XzvvJ
- [x] Add "Select all" Link.  This selects all possible columns and displays them in the table
- [x] Add "Reset to default" Link.  `Default` is the basic `column-state-config` defined by arranger on first load. 
- [x] Move the checkboxes to the left of the label (but ensure both checkbox and label are clickable)

PR #643, opened 2021-02-11
✨ Add 'Select All' and 'Reset to Default' Buttons to Column Select Dropdown (#640)

mistryrn:
Includes a rebuilt version of the dropdown component without using external libraries, and with a11y in mind, called `MultiSelectDropDown`.

This new dropdown is not enabled by default, to avoid any unintended breaking behaviour wherever this library is used.

To enable it, pass the prop `enableDropDownControls={true}` to the Arranger Table.

## Commits:
✨ Add 'Select All' and 'Reset to Default' Buttons to Column Select Dropdown
    * ♿️ Rewrite `DropDown` component for a11y in new `MultiSelectDropDown` component
    * Add ability to make multiple changes to `ColumnsState` at once via `toggleMultiple` method
    * Add `defaultColumns` to `ColumnsState` state object
💄 Styling for updated Column Select Dropdown (#640)
    * Update styling of new `MultiSelectDropDown`
    * Restore original `DropDown` component as the default in `TableToolbar`
    * Enable new `MultiSelectDropDown` by passing `enableDropDownControls={true}` to Table


modules/components/src/DataTable/DataTable.js
```diff
@@ -37,7 +37,9 @@ class DataTableWithToolbar extends React.Component {
       toolbarStyle,
       onFilterChange,
       onColumnsChange = () => {},
+      onMultipleColumnsChange = () => {},
       columnDropdownText,
+      enableDropDownControls = false,
```
joneubank, line 42: thanks for including this.

modules/components/src/DataTable/TableToolbar/TableToolbar.js
```diff
@@ -75,23 +78,48 @@ const TableToolbar = ({
...
-          </DropDown>
-        )}
+        {allowTogglingColumns &&
+          (enableDropDownControls ? (
+            <MultiSelectDropDown
+              buttonAriaLabelClosed={`Open column selection menu`}
+              buttonAriaLabelOpen={`Close column selection menu`}
+              itemSelectionLegend={`Select columns to display`}
+              selectAllAriaLabel={`Select all columns`}
+              resetToDefaultAriaLabel={`Reset to default columns`}
```
joneubank, line 88: Is it typical to have custom Aria Labels as inputs to components? I don't have a better suggestion off the top of my head but this seems cumbersome. Not asking for changes just want to put some thought into the best way to manage this kind of thing across the whole library of components.

modules/components/src/DropDown/MultiSelectDropDown.js
```diff
@@ -0,0 +1,319 @@
+import React, { useEffect, useRef, useState } from 'react';
+import './DropDown.css';
+
+let instances = 0;
+
+const ArrowIcon = ({ isOpen }) => {
```
joneubank, line 6: Can we move SVG icons to their own dedicated files? Typically we have a folder full of svg's to reference instead of burying them in files.

modules/components/src/DropDown/MultiSelectDropDown.js
```diff
@@ -0,0 +1,319 @@
...
+      stroke="#979797"
+      strokeWidth="1.1px"
+      transform={isOpen ? 'rotate(180)' : null}
+    >
+      <path d="M1,6 L10,15 L19,6" />
+    </svg>
+  );
+};
+
+const CheckIcon = () => {
```
joneubank, line 22: Repeat comment. SVGs to their own files and directory.

modules/components/src/DropDown/MultiSelectDropDown.js
```diff
@@ -0,0 +1,319 @@
...
+        <path
+          d="M.225 4.877c-.687-.686.357-1.73 1.043-1.044L2.596 5.16l4.136-4.135c.686-.687 1.73.357 1.043 1.044L3.118 6.726c-.289.289-.756.289-1.044 0L.224 4.877z"
+          transform="translate(-1105 -366) translate(0 63) translate(265 17) translate(0 237) translate(830 11) translate(0 27) translate(1 6) translate(9 4)"
+        />
+      </g>
+    </svg>
+  );
+};
+
+const ResetIcon = () => {
```
joneubank, line 35: Repeat comment. SVGs to their own files and directory.

modules/components/src/DropDown/MultiSelectDropDown.js
```diff
@@ -0,0 +1,319 @@
...
+      case 'End':
+        e.preventDefault();
+        focusLast();
+        break;
+      default:
+        break;
+    }
+  };
+
+  const handleEsc = (e) => {
```
joneubank, line 160: Nice touch.

modules/components/src/DropDown/MultiSelectDropDown.js
```diff
@@ -0,0 +1,319 @@
...
+      setIsOpen(false);
+      buttonRef.current.focus();
+    }
+  };
+
+  const handleBlur = (e) => {
+    const currentTarget = e.currentTarget;
+    // blur happens before click, preventing any click events in children from firing due to rerender from state change
+    // so wait a tick for child component events to fire before changing open state and causing rerender
+    window.setTimeout(() => {
```
joneubank, line 171: With event handling, I believe we can take an action without preventing other events from firing. Returning true will allow the event to continue and other handlers to take actions. So we shouldn't need this arbitrary delay.

joneubank: PS. thank you for comments explaining the timeout.

modules/components/src/DropDown/MultiSelectDropDown.js
```diff
@@ -0,0 +1,319 @@
...
+  useEffect(() => {
+    if (isOpen) {
+      window.addEventListener('keydown', handleEsc);
+      focusFirst();
+    }
+
+    return () => {
+      window.removeEventListener('keydown', handleEsc);
+    };
+  }, [isOpen]);
```
joneubank, line 223: Well handled.

modules/components/src/DropDown/MultiSelectDropDown.js
```diff
@@ -0,0 +1,319 @@
...
+      >
+        <div className="dropDownButtonContent">{children}</div>
+        <ArrowIcon isOpen={isOpen} />
+      </button>
+      {isOpen && (
+        // tabIndex is required for the onBlur handler to function, but
+        // adding tabIndex triggers this linter rule requiring a role
+        // to be added to the element. there are no roles which match the
+        // function of this, since it is a non-interactive element, hence
+        // the disabling of the rule here.
```
joneubank, line 246: 👍

joneubank (changes requested):
Lots of really good work. Can we take a look at the onBlur event, I think we can write that without the timeout.

Also, any thoughts about how React components typically handle all the aria label customization? Is it typically through a list of component properties like you did?

joneubank (approved):

PR #647, opened 2021-03-02
Session Storage for Table browsing properties

joneubank:
Arranger Table property to enabled sessionStorage, which stores for a user their selected column selections, pageSize, and sort order. These are set for the table on page load.

Initial plan was to expand the components to add props that would specify columns or pagination properties, but developing all the onEvent properties to make setting these properties outside the table was going to be significantly worse for the component. Instead, the default Arranger table is given an optional prop to let it handle all the session storage itself. a `storageKey` prop is also provided allowing each Arranger Table in a site to have a unique identifier to store their own settings, or use the same key and this share these settings (not this would have undesired consequences if the two tables have different column options).


modules/components/src/DataTable/ColumnsState.js
```diff
@@ -34,6 +34,20 @@ export default class extends Component {
     };
   }
 
+  getStorageKey() {
+    return `arranger-columnstate-toggled-${this.props.storageKey || ''}`;
+  }
+
+  getStoredToggled() {
+    if (this.props.sessionStorage) {
+      console.log('jondebug', 'ColumnState getStoredToggled', 'attempting to read session storage');
```
mistryrn, line 43: Do we need this debug log?

modules/components/src/DataTable/ColumnsState.js
```diff
@@ -138,12 +155,21 @@ export default class extends Component {
     this.setState({ temp }, () => this.save(temp));
   };
 
+  setColumnSelections(toggled) {
```
mistryrn, line 158: This is sweet! 👏

modules/components/stories/index.js
```diff
@@ -1,5 +1,5 @@
 import './Portal';
-import './table';
+import './Table';
```
mistryrn, line 2: Nice catch

mistryrn (approved):
Looks good! Just one console.log to clean up, but other than that good to merge :)

PR #651, opened 2021-03-09
[650] get environment variables on runtime

justincorrigible:
It seems desirable to support passing a basename path, in case the app is deployed as a nested service inside of a larger platform. This code grabs the environment variables on runtime (on page load), both to enable calling the server proxy, and to ensure React Router handles history accordingly.


Dockerfile
```diff
@@ -54,7 +54,7 @@ ENV APP_USER=node
 WORKDIR $APP_HOME
 
 RUN cd modules/admin-ui \
-	&& REACT_APP_ARRANGER_ADMIN_ROOT=/admin/graphql npm run build
+	&& npm run build
```
caravinci, line 57: Env variable being set at runtime, whenever available, else we can better set it in a `.env` than hardwiring them while building.

Dockerfile
```diff
@@ -68,6 +68,7 @@ ENV APP_USER=node
 ENV APP_HOME=/app
 ENV PORT=3000
 ENV NGINX_CONF_PATH=/etc/nginx/nginx.conf
+ENV REACT_APP_BASE_URL=${REACT_APP_BASE_URL:-''}
```
caravinci, line 71: This sets a default, to be used when the app stands alone (i.e. without basename)

Dockerfile
```diff
@@ -82,12 +83,14 @@ RUN addgroup -S -g $APP_GID $APP_USER \
...
 
 COPY --from=builder2 /app $APP_HOME
+RUN chown -R $APP_UID:$APP_GID $APP_HOME/arranger-admin
 
 WORKDIR $APP_HOME
 
 USER $APP_USER
 
-CMD envsubst '$PORT,$REACT_APP_ARRANGER_ADMIN_ROOT' < /etc/nginx/nginx.conf.template > $NGINX_CONF_PATH && exec nginx -c $NGINX_CONF_PATH -g 'daemon off;'
+CMD envsubst < docker/ui/env-config.template.js > /app/arranger-admin/env-config.js \
```
caravinci, line 92: Grabs the process environment and replaces values in a file that will be loaded by the browser.

docker/ui/env-config.template.js
```diff
@@ -0,0 +1,4 @@
+window.__env__ = {
+  REACT_APP_BASE_URL: '${REACT_APP_BASE_URL}',
+  REACT_APP_ARRANGER_ADMIN_ROOT: '/${REACT_APP_BASE_URL}/admin/graphql'.replace('//', '/'),
```
caravinci, line 3: if no value is passed, this ensures the proxy remains within the host URL, rather than replacing it. (e.g. `http://localhost/admin/graphql` ✔️  vs `http://admin/graphql` ❌ )

modules/admin-ui/public/env-config.js
```diff
@@ -0,0 +1 @@
+window.__env__ = {};
```
caravinci, line 1: default, to avoid misleading error messages

caravinci (commented):
Some comments to provide context and assist review.
Thanks for your assistance! 🙏

joneubank (approved):

PR #660, opened 2021-03-18
DataTable store row selections

joneubank:
Added more sessionStorage controlled functionality. Row selections in the DataTable will be stored in sessionStorage when changed and read in when the table is constructed. The existing storage methods now have slightly more generic methods to reduce the code repetititition.

Setting sessionStorage to true is sufficient to keep the selections while navigating through the app, but to make the selections persists over page loads, the data table also requires `keepSelectedOnPageChange={true}`. This is an old existing property that afaik wasn't previously doing anything. I kept this as it was just in case it had an existing use I am unaware of, and also since it was working as expected based on the property name.

### DIFF
--- a/modules/components/src/DataTable/DataTable.js
+++ b/modules/components/src/DataTable/DataTable.js
@@ -5,18 +5,30 @@ import urlJoin from 'url-join';
 import { ARRANGER_API, PROJECT_ID } from '../utils/config';
 import { Table, TableToolbar } from './';
 
+const STORED_PROPS = {
+  PAGE_SIZE: 'PAGE_SIZE',
+  SORT_ORDER: 'SORT_ORDER',
+  SELECTED_ROWS: 'SELECTED_ROWS',
+};
+
 class DataTableWithToolbar extends React.Component {
   constructor(props) {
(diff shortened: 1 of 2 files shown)


modules/components/src/DataTable/DataTable.js
```diff
@@ -88,8 +103,8 @@ class DataTableWithToolbar extends React.Component {
       downloadUrl = urlJoin(ARRANGER_API, projectId, 'download'),
       onSortedChange = () => {},
       alwaysSorted = [],
-      initalSelectedTableRows = [],
-      keepSelectedOnPageChange = false,
+      initalSelectedTableRows,
+      keepSelectedOnPageChange = false, // If false, this will reset the selection to empty on reload even if sessionStorage is enabled. To keep selections after reload st this to true.
```
caravinci, line 107:
```suggestion
      keepSelectedOnPageChange = false, // If false, this will reset the selection to empty on reload even if sessionStorage is enabled. To keep selections after reload set this to true.
```

caravinci (approved):
LGTM 👍

PR #662, opened 2021-03-19
[658] allow multiple table exporters and column filters

### LINKED ISSUE #658: Allow passing multiple custom table exporters, and display them as a dropdown
> ### Placeholder text / ticket

> Other projects may choose to offer more than the default  Download TSV" option. This ticket documents the implementation for that feature.

caravinci:
Enables clients to generate a dropdown button to Download files, etc. based on the data displayed on the table.

Closes #658.


modules/components/src/DataTable/TableToolbar/helpers.js
```diff
@@ -0,0 +1,67 @@
...
+          ),
+    );
+
+  const multipleExporters = exporterArray && exporter.length > 1;
+
+  // this checks whether a single custom function has been provided
+  // by itself, or as the single item in an array
+  const customExporter =
+    exporter && //
+    (exporter instanceof Function // check it's a valid function
```
caravinci, line 50: if exporter is a single function, use it.

modules/components/src/DataTable/TableToolbar/helpers.js
```diff
@@ -0,0 +1,67 @@
...
+      ? exporter // then use it
+      : exporterArray // or check whether it's an array of them
+      ? multipleExporters // that contains more than one
+        ? exporterArray // return it, but may lead to bugs if misused
+        : exporterArray[0]?.exFunction // or a single element in an array
```
caravinci, line 55: If it's an array, either use it as such for more than one item, or get a single item's function.
i.e. `exporterArray = [{function: noop}]` will make `customExporter=noop`

modules/components/src/DropDown/DropDown.js
```diff
@@ -61,29 +69,39 @@ class DropDown extends React.Component {
...
-                    <input
-                      readOnly
-                      type="checkbox"
-                      checked={selectedItem.indexOf(item) > -1}
-                      aria-label={`Select column ${item.id || itemToString(item)}`}
-                    />
-                  </div>
-                ))}
+                {items.map((item, index) =>
+                  singleSelect ? (
```
caravinci, line 81: This dropdown component will now allow single and multiple selection

caravinci (commented):
Some comments to try assisting the review process, as usual

modules/components/src/DataTable/TableToolbar/helpers.js
```diff
@@ -0,0 +1,67 @@
...
+    params: {
+      fileName,
+      files: files.map(({ allColumns, columns, exColumns = null, ...file }, i) => ({
+        ...file,
+        columns: exColumns // if the component gave you custom columns to show
+          ? Object.values(
+              exColumns.length > 0 // if they ask for any specific columns
+                ? Object.keys(allColumns).filter((fieldName) => exColumns.includes(fieldName)) // use them
+                : allColumns, // else, they're asking for all the columns
+            )
```
caravinci, line 17: For future reference, this is where the default columns are given, in case we decide that option doesn't work.

Line 16, particularly, could either have `columns`, to provide only the 'active' columns.

modules/components/src/DataTable/TableToolbar/helpers.js
```diff
@@ -0,0 +1,67 @@
...
+    ...options,
+    params: {
+      fileName,
+      files: files.map(({ allColumns, columns, exColumns = null, ...file }, i) => ({
+        ...file,
+        columns: exColumns // if the component gave you custom columns to show
+          ? Object.values(
+              exColumns.length > 0 // if they ask for any specific columns
+                ? exColumns.map((fieldName) => allColumns[fieldName]) // use them
+                : allColumns, // else, they're asking for all the columns
```
caravinci, line 16: For future reference, this block is where the default columns are determined, if the client app gives us an empty array (which is not 'nothing'), in case we decide responding with all the fields in the index doesn't work.

Line 16, particularly, could instead have columns, to provide only the 'active' columns (i.e. as set in admin-ui) .

modules/components/src/DataTable/TableToolbar/TableToolbar.js
```diff
@@ -120,35 +131,66 @@ const TableToolbar = ({
...
-              {exportTSVText}
-            </button>
-          </div>
+
+        {multipleExporters ? ( // check if we're given more than one custom exporter
+          <DropDown
+            aria-label={`Download options`}
+            itemToString={(i) => i.exLabel}
+            items={exporterArray}
+            onChange={({ exColumns, exLabel, exFunction }) =>
```
joneubank, line 140: For these variables it is unclear what `ex` means. I assumed `extra` but thats not aligned with the code here, so probably means something else I am missing. As a general rule, don't use abbreviations in variable names. Long variable names are no issue.

modules/components/src/DataTable/TableToolbar/helpers.js
```diff
@@ -0,0 +1,67 @@
...
+          ),
+    );
+
+  const multipleExporters = exporterArray && exporter.length > 1;
+
+  // this checks whether a single custom function has been provided
+  // by itself, or as the single item in an array
+  const customExporter =
+    exporter && //
+    (exporter instanceof Function // check it's a valid function
```
joneubank, line 50: This ternary chain is particularly tough to read, need to clean this up to make it easier to read and work with.

Because you are trying to assign the value, using an anonymous method could be useful. Here's one potential implementation using a switch for all the cases instead of an if/else ladder or ternary chain.

```
const resolveCustomExporter = () => {
    switch(true) {
        case exporter instanceof Function:
            return exporter;
        case exporterArray && exporter.length > 1:
            return exporterArray;
        case exporterArray && exporter.length === 1;
            return exporterArray[0]?.exFunction;
        default:
            console.error('...');
    }
}
```

modules/components/src/DropDown/DropDown.js
```diff
@@ -61,29 +69,39 @@ class DropDown extends React.Component {
               <div className="dropDownButtonContent">{children}</div>
               <ArrowIcon isOpen={isOpen} />
             </button>
-            {!isOpen ? null : (
+            {isOpen && (
```
joneubank, line 72: 😆 nice catch.

joneubank (changes requested):
Reasonable solution. As we discussed, we need to record in the file the structure of the `exporter` param, otherwise the next dev will never know how to use it. Let's add that as comments.

My comments here are not functional, just about code readability. Thanks Justin!

modules/components/src/DataTable/TableToolbar/TableToolbar.js
```diff
@@ -120,35 +131,66 @@ const TableToolbar = ({
...
-                    ],
-                  }),
-                );
-              }}
-            >
-              {exportTSVText}
-            </button>
-          </div>
+
+        {multipleExporters ? ( // check if we're given more than one custom exporter
```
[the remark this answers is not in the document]
caravinci: Though I agree 100% with that, I envision abstracting this whole "exporter"  thing into its own separate component, leveraging a certain upcoming Arranger context provider ;)
The props drilling is quite unwieldy as it stands.

modules/components/src/DataTable/TableToolbar/TableToolbar.js
caravinci: You're 100% right, and I'm normally of the same policy. 
Guess working late at night is not doing wonders for my code quality these days, uh?
😆 💀 

Changes committed!

modules/components/src/DataTable/TableToolbar/helpers.js
caravinci: Gotta be honest: I'm a big fan of nested ternaries, because I usually indent them in what I find is a very readable way.
Prettier here works against that style, so your suggestion to use a switch is of course perfectly adequate.
Will amend this now. Thanks for the suggestion!

caravinci:
Ready for reevaluation, @joneubank

caravinci:
Updated the branch with the latest from `develop` to ensure integration

joneubank (approved):
Good work!

PR #665, opened 2021-03-23
add row selection filter for table exporters

caravinci:
@joneubank realised not all projects want to use the row selection as a filter, so this adds the option to disable that feature.

### DIFF
--- a/modules/components/src/DataTable/TableToolbar/TableToolbar.js
+++ b/modules/components/src/DataTable/TableToolbar/TableToolbar.js
@@ -53,6 +53,7 @@ const TableToolbar = ({
   defaultColumns,
   downloadUrl,
   enableDropDownControls = false,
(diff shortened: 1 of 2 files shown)


modules/components/src/DataTable/TableToolbar/TableToolbar.js
```diff
@@ -85,20 +86,21 @@ const TableToolbar = ({
...
-              op: 'in',
-              content: { field: 'file_autocomplete', value: selectedTableRows },
-            },
-          ],
-        },
-        sqon,
-      )
-    : sqon;
+  const downloadSqon =
+    enableSelectedTableRowsExporterFilter && selectedTableRows.length
```
joneubank, line 90: 👍

joneubank (approved):

PR #667, opened 2021-03-24
add capacity to customise export filenames

caravinci:
This one should be straightforward:
Custom exporters may want to specify their desired file names.


joneubank (approved):
Cool stuff.

PR #686, opened 2021-05-17
Style and Prop enhancements for DataTable Toolbar

joneubank:
- Missing space in the summary count
- DataTable given prop to control multiple exporter button label
- Added consistent spacing between column and exporters dropdown.
- Filter by selected rows filter made disabled by default to protect functionality where currently used
  - previously we decided to enable this by default, but the changes caused a bug with an existing site so restoring the original behaviour as default.
- Prop to customize the filter property for selected row filter


modules/components/src/DataTable/TableToolbar/TableToolbar.js
```diff
@@ -188,48 +189,50 @@ const TableToolbar = ({
...
-                  url: downloadUrl,
-                }),
-                download,
-              )
-            }
-            singleSelect={true}
-          >
-            {exporterLabel}
-          </DropDown>
+          <div className="buttonWrapper">
```
joneubank, line 192: Lots of lines changed due to indenting. Only change is adding this wrapper.

PR #691, opened 2021-05-20
fix sqon display conditional logic for date values

caravinci:
Logic behooves being logical 🤡

### DIFF
--- a/modules/components/src/Arranger/CurrentSQON.js
+++ b/modules/components/src/Arranger/CurrentSQON.js
@@ -35,10 +35,9 @@ export const CurrentSQON = ({
             translateSQONValue,
             internalTranslateSQONValue,
           )(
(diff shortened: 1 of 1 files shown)


modules/components/src/Arranger/CurrentSQON.js
```diff
@@ -35,10 +35,9 @@ export const CurrentSQON = ({
             translateSQONValue,
             internalTranslateSQONValue,
           )(
-            (findExtendedMappingField(field)?.displayValues || {})[value] ||
-              findExtendedMappingField(field)?.type === 'date'
```
caravinci, line 39: this || was all bundled up as the single condition for the ternary, causing boolean values to be misidentified as dates

mistryrn (approved):

PR #696, opened 2021-05-25
Use displayName instead of Header for data stream if available

joneubank:
Due to custom columns in search table, it is possible to not have a string value in the Header property for a column. In its place, the displayName value is used. This is a common pattern used with the columns object, adding it here to have header string values available for dataStream exports.


mistryrn (approved):
👍

PR #699, opened 2021-07-07
feat: No global aggregation wrapper if not required

adipaul1981:
If some of these changes are already supported by arranger in its current form, please let me know. We'll make the necessary changes to the portal. Arranger is quite vast, its possible we did not understand a thing or 2...

### No Global Aggregation

[https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-global-aggregation.html](url)

In some queries, we do not want the buckets to be wrapped in the global aggregation, rather to **be** influenced by the search query.

```
{
   "observed_phenotype.name:global":{
      "global":{},
      "aggs":{
           "observed_phenotype.name:nested":{
               "nested":{
                  "path":"observed_phenotype"
               },
               "aggs":{
                     [--- rest of aggs ---]
               } 
          }
      }
  }
}
```
But we would rather this:
    
```
"observed_phenotype.name:nested":{
    "nested":{
          "path":"observed_phenotype"
      },
      "aggs":{
           [--- rest of aggs ---]
       } 
}
```

Maybe i'm not understanding something in arranger, but in `buildAggregation/index.js` the `createGlobalAggregation` function is always called. We have created a separate filter parameter to be able to specify that the global aggregation wrapper is not required (`no_global_aggregation`). 
     
        
### Exists query

Input sqon:
```
{
   "op":"in",
   "content":{
      "field":"tag.keyword",
      "value":[
         "__missing__"
      ]
   }
}
```
Currently in arranger if a value of `__missing__` is given, is will generate a query like this:

```
{
   "bool":{
      "must_not":[
         {
            "exists":{
               "boost":0,
               "field":"some_field"
            }
         }
      ]
   }
}
```
Returning all documents for which `some_field` does **not** exist. 

If we want all documents for which `some_field` **do** exists, it does not appear possible. In `getMissingFilter `function in `buildQuery/index/js` the value `isNot` is hardcoded to true.  If we enable to set `isNot` to false (as with the `__exists__` parameter), it the case the output of the function would be:
```
 {
    "exists":{
       "boost":0,
       "field":"some_field"
    }
 }
```
giving the result we would want.

### Regex Filter
[https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-regexp-query.html](url)
vs.
[https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-wildcard-query.html](url)

These changes were made in out arranger fork, however after verification we don't seem to use this functionality (regex filter) at all. Thus, all these changes could be ignored. I'm including this in here because the way its implemented in arranger does not appear to me to be correct. 

Currently Arranger does not really allow to truly pass a regex as input. Only more like a wildcard type of query. 
`esFilter: { regexp: { [field]: value.replace('*', '.*') } },` 
that get converted into a regex (.*)

but a wildcard query already exist trough passing `op: 'filter' ` in the input sqon, so why the regex filter would be required in the current state?

The following proposed changes enable to pass a regex as a value, but like i said, its not used in the portal.

### DIFF
--- a/modules/middleware/src/buildQuery/index.js
+++ b/modules/middleware/src/buildQuery/index.js
@@ -25,6 +25,7 @@ import {
   ALL_OP,
   ES_SHOULD,
   ES_WILDCARD,
+  EXISTS,
 } from '../constants';
 import normalizeFilters from './normalizeFilters';
 import {
@@ -62,7 +63,7 @@ function getRegexFilter({ nestedFields, filter }) {
   const esFilter = wrapFilter({
     filter,
     nestedFields,
-    esFilter: { regexp: { [field]: value.replace('*', '.*') } },
+    esFilter: { regexp: { [field]: value.replace(REGEX, '') } },
     isNot: NOT_IN_OP === op,
   });
 
@@ -120,15 +121,16 @@ function getFuzzyFilter({ nestedFields, filter }) {
   );
 }
 
-function getMissingFilter({ nestedFields, filter }) {
+function getMissingFilter({ nestedFields, filter, isNotRequired = true }) {
   const {
     content: { field },
   } = filter;
+
   return wrapFilter({
     esFilter: { exists: { field: field, boost: 0 } },
     nestedFields,
     filter,
-    isNot: true,
+    isNot: isNotRequired,
   });
 }
 
@@ -261,6 +263,8 @@ export const opSwitch = ({ nestedFields, filter }) => {
       return getSetFilter({ nestedFields, filter });
     } else if (`${value[0]}`.includes(MISSING)) {
       return getMissingFilter({ nestedFields, filter });
+    } else if (`${value[0]}`.includes(EXISTS)) {
+      return getMissingFilter({ nestedFields, filter, isNotRequired: false });
     } else {
       return getTermFilter({ nestedFields, filter });
     }
--- a/modules/mapping-utils/src/createConnectionTypeDefs.js
+++ b/modules/mapping-utils/src/createConnectionTypeDefs.js
@@ -37,6 +37,7 @@ export default ({ type, fields = '', createStateTypeDefs = true }) => `
       include_missing: Boolean
       # Should term aggregations be affected by queries that contain filters on their field. For example if a query is filtering primary_site by Blood should the term aggregation on primary_site return all values or just Blood. Set to False for UIs that allow users to select multiple values of an aggregation.
       aggregations_filter_themselves: Boolean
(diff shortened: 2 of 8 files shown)


joneubank:
For the Global Aggregation changes - I believe to global filter is applied to every aggregation in `buildAggregations/index.js`, specifically in the `wrapWithFilters` method where you added the no_global_filter check, the cleanedQuery is added as a filter on every aggregation. This is done to give us the option to remove filters on this field from the query for the aggregation through the `aggregations_filter_themselves` property. So to have facets filtered by the global sqon just setting `aggregations_filter_themselves` to `true` should accomplish the same thing?
```
function wrapWithFilters({ field, query, aggregationsFilterThemselves, aggregation }) {
  if (!aggregationsFilterThemselves) {
    const cleanedQuery = removeFieldFromQuery({ field, query });
    // TODO: better way to figure out that the field wasn't found
    if (!isEqual(cleanedQuery || {}, query || {})) {
      return createGlobalAggregation({
        field,
        aggregation: createFilteredAggregation({
          field,
          filter: cleanedQuery, // <====== FILTER ADDED HERE SO THIS GLOBAL AGG IS SAME AS IF IT WAS NOT GLOBAL
          aggregation,
        }),
      });
    }
  }
  return aggregation;
}
```
I may be missing the difference here, in which case let me know and I will try it out in more detail.

modules/middleware/src/buildQuery/index.js
```diff
@@ -62,7 +63,7 @@ function getRegexFilter({ nestedFields, filter }) {
   const esFilter = wrapFilter({
     filter,
     nestedFields,
-    esFilter: { regexp: { [field]: value.replace('*', '.*') } },
+    esFilter: { regexp: { [field]: value.replace(REGEX, '') } },
```
joneubank, line 66: This is certainly better than the old implementation, but will definitely be breaking for any applications using it as it was. If you're not in need of this change, perhaps we should remove the change for now and open an issue to fix this properly with the next major version.

joneubank:
The changes to filter for property existence is a great addition. If we put the Exists Filter changes into a PR of its own we can merge that right away, or lets resolve the other changes on this PR if you'd prefer updating this one.

modules/middleware/src/buildQuery/index.js
adipaul1981: that works for me. No we dont seem to be using it.

adipaul1981:
This PR can be closed.

PR #700, opened 2021-07-12
Feat/set mutations delete update

adipaul1981:
### Sets 
The user has to possibility to create his own set. This PR enables this functionality.

**Additional `tag` parameter**

the "Name" the user can give to his set that he is saving. 
image

**New Set Mutation**

_1. Delete_

image

Deletes a user set

ex. query

```
mutation($setIds: [String!]) {
  deleteSets(setIds: $setIds)
}
```

`{"setIds": ["c4b078c3-9a0e-433d-9b9b-5b1839f25fd1"]}`

_2.  Update_

- Add or remove id from a set

image

image

ex. 
query:

```
mutation(
  $data: SetUpdateInputData!
  $setUpdateAction: SetActionTypes!
  $target: SetUpdateTarget!
) {
  updateSet(
    data: $data
    setUpdateAction: $setUpdateAction
    target: $target
  ) {
    setSize
    updatedResults
  }
}
```

variables:

```
{
  "data":{
    "sqon":{"op":"and","content":[{"op":"in","content":{"field":"family.family_compositions.composition","value":["trio+"]}}]},"path":"kf_id","type":"participant"
  },
  "setUpdateAction": "REMOVE_IDS",
  "target":{"setId":"c1311eb4-4355-4d09-8741-08f50d10605c"}
}
```

- Rename Tag

image

ex. 
query:

```
mutation(
  $data: SetUpdateInputData!
  $setUpdateAction: SetActionTypes!
  $target: SetUpdateTarget!
) {
  updateSet(
    data: $data
    setUpdateAction: $setUpdateAction
    target: $target
  ) {
    setSize
    updatedResults
  }
}
```

variables:
```
{
  "data":{"newTag":"Reported Unknown4444"},
  "setUpdateAction":"RENAME_TAG",
  "target":{"setId":"719d9b80-9d57-4fdc-8823-a069d01f0f3b"}
}
```

### DIFF
--- a/modules/mapping-utils/src/index.js
+++ b/modules/mapping-utils/src/index.js
@@ -1,6 +1,6 @@
 export { default as addMappingsToTypes } from './addMappingsToTypes';
 export { default as createConnectionResolvers } from './createConnectionResolvers';
-export { default as resolveSets, saveSet } from './resolveSets';
+export { default as resolveSets, saveSet, deleteSets, updateSet } from './resolveSets';
 export { default as mappingToFields } from './mappingToFields';
 export { default as mappingToAggsType } from './mappingToAggsType';
 export { default as mappingToAggsState } from './mappingToAggsState';
--- a/modules/mapping-utils/src/resolveSets.js
+++ b/modules/mapping-utils/src/resolveSets.js
@@ -3,6 +3,148 @@ import uuid from 'uuid/v4';
 import { CONSTANTS, buildQuery } from '@arranger/middleware';
 import esSearch from './utils/esSearch';
 import compileFilter from './utils/compileFilter';
+import {
+  addSqonToSetSqon,
+  makeUnique,
+  removeSqonToSetSqon,
+  retrieveIdsFromQuery,
+  truncateIds,
+} from './utils/sets';
+import mapHits from './utils/mapHits';
+
+const isQueryEmpty = (sqon) => !sqon || sqon.content.length === 0;
+
+const ActionTypes = {
+  RENAME_TAG: 'RENAME_TAG',
+  ADD_IDS: 'ADD_IDS',
+  REMOVE_IDS: 'REMOVE_IDS',
+};
+
+const renameTag = async ({ es, setId, newTag, subAction, userId }) => {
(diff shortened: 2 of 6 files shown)


joneubank:
These extensions to the sets API are reasonable but, after some discussion, we decided that this isn't the direction Arranger should take. I'm going to close this PR, and in the last paragraph here I've included some suggestions for an alternative solution to maintaining sets for users. If we want to discuss these options further, please let @rosibaj know and we can set up a quick chat. 

Our reasoning for this decision:

Sets in Arranger are not intended to be a long lived source of data: they’ve been included in the service to quickly provide a reference to a search result, or subset of a search result. They are intended to be transient and not stored for future reference.

We don’t want to expand the scope of Arranger sets to include managing existing sets, or adding custom user data (like names/tags), as this opens up the scope well beyond the service’s focus of searching and filtering ES data.

A couple reasons why Arranger isn’t the best place to maintain long-lived sets that you want to reference in the future:
ElasticSearch is not a durable data store. It is expected that all indices could lose their data and need to be regenerated in the future (examples of this include cluster upgrades, cluster split brain events between nodes). Thus, long term storage of user data sets should not be maintained in ES.
To access sets in the future already requires another service to store SetIDs for a user, so even the simplest implementation of this will start to mix responsibilities between Arranger and another service.
All sets are publicly available. Although the Set IDs are obscure due to being UUID, they in theory could be modified/deleted by any user, not just the user that created and is interested in them.
Since this data is publicly available, be conscious that any string label a user provides could be presented on your web site to another user without any sort of validation or filtering. This has concerns for vulgarity or private information leaks depending on what a user enters.

Instead of expanding the Arranger sets API, perhaps moving the set data into a system designed to store and maintain custom user data is what you want to do. Riff is one option, as it uses Ego authorization already and allows storage of JSON blobs with any custom structure you want. The content of a set (document IDs, SQON, tags, notes) can be stored there and the existing APIs used for recalling sets associated with a user. If you are taking advantage of using set documents to do quick ES queries, you can combine these strategies: use Arranger to make the set, then store the set ID and metadata about it in Riff for the user to reference in the future.

Issue #642, opened 2021-02-11
BUG - (Server module): undeclared variables prevent the use of an external context from graphql options.


evans-g-crsj:
## Description

When using the Arranger Server, various options can be injected into it such as `graphqlOptions`. For instance:

```
Arranger({
  projectId, 
  esHost,
 graphqlOptions: {
    context: ({ jwt }) => ({ jwt })
  },
}).then(... more code ...);
```

When the type of the `context` 's value is `function` the code breaks. 
## Expected Behaviour

When passing down a context function, the Arranger server should, in turn, passing it down to  Apollo's context smoothly.
## Actual Behaviour

When passing down a context function, the Arranger server silently throws a reference error.
## Possible Fix

I think that the reason for the bug is attributable to this #commit: c2311787938e115e15317193b711c87439585a0c . 

before the commit **request, response, graphQLParams** were declared:
```
graphqlExpress(async (request, response, graphQLParams) => {
          const externalContext =
            typeof graphqlOptions.context === 'function'
              ? await graphqlOptions.context(request, response, graphQLParams)
              : graphqlOptions.context;
          return {
            schema,
            context: {
              es,
              projectId: id,
              ...(externalContext || {}),
            },
          };
        })
```
after the commit **request, response, graphQLParams** are not declared (not even elsewhere):
```
const externalContext =
   typeof graphqlOptions.context === 'function'
      ? await graphqlOptions.context(request, response, graphQLParams)
      : graphqlOptions.context;
  const apolloServer = new ApolloServer({
    schema,
    context: {
      es,
      projectId: id,
      ...(externalContext || {}),
    },
  });
```
   
## Steps to Reproduce

Start an Arranger server and pass down options this way (or similar ones): 

```
Arranger({
  projectId,
  esHost,
 graphqlOptions: {
    context: ({ jwt }) => ({ jwt }), /* <== context's value must be of type 'function' */
  },
}).then((router) => {
  app.use(router);

  app.listen(port, async () => {
    console.log(`⚡️ Listening on port ${port} ⚡️`);
  });
});
```
It should log "request is not defined". _By the way, shouldn't the app throw an error instead of swallowing it and log?_

PR #702, opened 2021-07-26
fix:#642 fix missing request response and connection in context function

celinepelletier:
Fixes https://github.com/overture-stack/arranger/issues/642


caravinci (approved):
LGTM 👍

Issue #717, opened 2021-09-01
GQL Aggregation Query should allow selection of max buckets to return


joneubank:
## Detailed Description

Currently Arranger will return up to 300000 buckets for each aggregation. This is [hardcoded for field aggregations](https://github.com/overture-stack/arranger/blob/develop/modules/middleware/src/buildAggregations/createFieldAggregation.js#L7). When retrieving facets, having 100s or 1000s of buckets returned makes the request very slow.

Elastic search has an argument to specify the max buckets returned for a field aggregation: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html#search-aggregations-bucket-terms-aggregation-size

There should be a mechanism in the GQL query interface to allow users to specify the max number of buckets returned.

We should also change the default number to something reasonable for the facet sidebar (I suggest 10, the ES default). We can also make this value a configurable option provided when Arranger is started.

## Possible Implementation
Simplest implementation would be to add an argument to the `buckets` property of `aggregations`, this argument (example `max: number`) would specify the max number of buckets to return. If this argument is omitted then the default value is used. That would make a query look like:

```
query {
  document_type {
    aggregations {
      some_id_with_many_buckets {
        buckets(max: 25) {
          doc_count
          key_as_string
        }
      }
    }
  }
}
```

PR #718, opened 2021-09-02
feat: #717 enable to set a maximum number of aggregations

adipaul1981:


modules/middleware/src/buildAggregations/createFieldAggregation.js
```diff
@@ -25,6 +25,9 @@ const createNumericAggregation = ({ type, field, graphqlField }) => {
 };
 
 const createTermAggregation = ({ field, isNested, graphqlField }) => {
+  const maxAggregations = get(graphqlField, ['buckets', '__arguments', 0], {
+    max: { value: MAX_AGGREGATION_SIZE },
+  }).max.value;
```
caravinci, line 30: We can also `get` the value directly using the path array, making it easier to read.

```suggestion
  const maxAggregations = get(
    graphqlField,
    ['buckets', '__arguments', 0, 'max', 'value'],
    MAX_AGGREGATION_SIZE
  );
```

caravinci (changes requested):
Thanks for the contribution!
Just a tiny request/suggestion, to make life easier for future devs.

modules/middleware/src/buildAggregations/createFieldAggregation.js
adipaul1981: done

caravinci (approved):
LGTM :shipit:

PR #744, opened 2022-01-14
Issue with nested aggregations

adipaul1981:
## Issue:

We don’t get same results for aggregations vs. hits when dealing with nested fields. The aggreagtions in this case are not accurate. 

For a given Sqon:

```
{
   "sqon":{
      "op":"and",
      "content":[
         {
            "op":"in",
            "content":{
               "field":"participants.affected_status",
               "value":[
                  "true"
               ]
            }
         },
         {
            "op":"in",
            "content":{
               "field":"participants.gender",
               "value":[
                  "Female"
               ]
            }
         }
      ],
      "pivot":"participants"
   }
}
```

1. ### Aggregation
    we get 2394 participants **affected status == true**  
    image

    
    we get 2394 (same) participants **gender == female**
    image

2. ### Hits
    We get a different result here for participants that are both **affected status == true** and **gender == female**. 2030 participants
    image
    
    
The correct count is the result given by the Hits (2030 particpants)

## Proposed Solution

Wrap the existing aggreagtion **for nested term** generated by arranger with a extra term filter as so:

image
image

## Response

1. ### Aggregation
    we get 2030 participants **affected status == true**  
   image

    
    we get 2030 (same) participants **gender == female**
    image

2. ### Hits
    (no change) 2030 Participants
    image

------------------------------------------------------------------

## Changes to buildQuery.js

The pivot in the sqon with only be applied if elements are in proper order in SQON.

ex these 2 sqons will give 2 very different results (juste order change):

donors -> consequences -> consequences

```
{
  "sqon":{
    "content":[
      {"content":{"field":"donors.patient_id","value":["PA00049"]},"op":"in"},
      {"content":{"field":"consequences.biotype","value":["protein_coding"]},"op":"in"},
      {"content":{"field":"consequences.consequences","value":["downstream_gene_variant"]},"op":"in"}
  ],"op":"and",
    "pivot":"donors"
  }
 }
```
consequences -> consequences -> donors

```
{
  "sqon":{
    "content":[
      {"content":{"field":"consequences.biotype","value":["protein_coding"]},"op":"in"},
      {"content":{"field":"consequences.consequences","value":["downstream_gene_variant"]},"op":"in"},
      {"content":{"field":"donors.patient_id","value":["PA00049"]},"op":"in"}
  ],"op":"and",
    "pivot":"donors"
  }
 }
```
No 1 gives "total": 1892,
No 2 gives "total": 1303,

The pivot field in the no. 2 case will have no effect as only the 1st element in the list of fields is looked at and compared.


caravinci:
Thanks for catching that bug and for the PR @adipaul1981 🙏 
The code looks great. However, as we have been refactoring some parts of Arranger (e.g. `middleware` now is just part of the `server` module), some conflicts have come up. 

Likely what we'll have do is merge this into the v2.x stream (which we intended to keep just for maintenance), and port these changes also in the new Arranger. Id' like to chat with @joneubank, just to confirm that's the right course of action, and get back to you asap.

joneubank:
Thanks for digging into the ES queries here @adipaul1981 . The part of this change that is fixing how nested queries are being added to aggs queries is a good bug fix for us.

I don't think it is a good idea to introduce the concept of "Ordering Matters" to SQONs. I think we can accomplish the same logic using nested operations instead. Building off of your donors/consequences example, I think you can separate out the pivot logic into its own nested operation like this instead:

```json
{
  "op": "and",
  "content": [
    {
      "op": "and",
      "pivot": "donors",
      "content": [
        {"content":{"field":"donors.patient_id","value":["PA00049"]},"op":"in"}
      ]
    },

    {"content":{"field":"consequences.biotype","value":["protein_coding"]},"op":"in"},
    {"content":{"field":"consequences.consequences","value":["downstream_gene_variant"]},"op":"in"}
  ]
}
```

Can you test using SQONs with this type of structure?

modules/middleware/src/buildAggregations/index.js
```diff
@@ -97,10 +98,18 @@ export default function ({
   const aggs = Object.entries(graphqlFields).reduce((aggregations, [fieldKey, graphqlField]) => {
     const field = fieldKey.replace(/__/g, '.');
     const nestedPaths = getNestedPathsInField({ field, nestedFields });
+    const contentsFiltered = (normalizedSqon?.content || []).filter((c) =>
+      aggregationsFilterThemselves
+        ? c.content?.field?.startsWith(nestedPaths)
+        : c.content?.field?.startsWith(nestedPaths) && c.content?.field !== field,
+    );
+    const termFilters = contentsFiltered.map((filter) => opSwitch({ nestedFields: [], filter }));
```
joneubank, line 106: Can we remove the consideration for ordering of filters (ie. startsWith) . We should assume all filters inside a pivot operation are meant to be nested in ES.

joneubank (changes requested):
Lets keep SQONs indifferent to the order of filters. Thanks!

adipaul1981:
> Thanks for digging into the ES queries here @adipaul1981 . The part of this change that is fixing how nested queries are being added to aggs queries is a good bug fix for us.
> 
> I don't think it is a good idea to introduce the concept of "Ordering Matters" to SQONs. I think we can accomplish the same logic using nested operations instead. Building off of your donors/consequences example, I think you can separate out the pivot logic into its own nested operation like this instead:
> 
> ```json
> {
>   "op": "and",
>   "content": [
>     {
>       "op": "and",
>       "pivot": "donors",
>       "content": [
>         {"content":{"field":"donors.patient_id","value":["PA00049"]},"op":"in"}
>       ]
>     },
> 
>     {"content":{"field":"consequences.biotype","value":["protein_coding"]},"op":"in"},
>     {"content":{"field":"consequences.consequences","value":["downstream_gene_variant"]},"op":"in"}
>   ]
> }
> ```
> 
> Can you test using SQONs with this type of structure?

@joneubank  For the pivot issue (and the order of the elements in sqon). I've tried the above sqon.   


I get a wrong value again (1892), 
image

modules/middleware/src/buildAggregations/index.js
adipaul1981: Do yuo mean doind something like this instead?:
```
const contentsFiltered = (normalizedSqon?.content || []).filter((c) =>
      aggregationsFilterThemselves && c.content?.field !== field,
 );
```

joneubank: Having read through this more in depth and tested it out, the original code makes sense - I was confused. This works as written.

modules/middleware/src/buildQuery/index.js
```diff
@@ -190,7 +190,7 @@ function getGroupFilter({ nestedFields, filter: { content, op, pivot } }) {
   const applyBooleanWrapper = wrappers[op];
   const esFilters = content.map((filter) => opSwitch({ nestedFields, filter }));
   const isNested = !!esFilters[0]?.nested;
-  if (isNested && pivot === esFilters[0]?.nested.path) {
+  if (isNested && esFilters.map((f) => f.nested?.path).includes(pivot)) {
```
joneubank, line 193: This definitely needed changing to remove the impact of ordering of the entries in the filter. What I'm trying to determine is if there is any reason to check that the filters include that path.. if the sqon has the `pivot` property then we should use it without any other checks. So, why not just:
`if (isNested && pivot) {`

joneubank (approved):
I ran some tests locally and was able to reproduce the aggregation count issue. Your changes do fix the issue for the common use case. I think this is ok to include as it is, but I did leave one additional comment to consider.

Thank you @adipaul1981 .

PR #755, opened 2022-02-02
add default dataFetcher in Components Context

caravinci:
- Ports the old data fetching function from the `<Arranger />` root component into the new context provider, and allows for local customisation of fetcher per context invocation.
- Improves the TypeScript implementation, adding type definitions to the published package, and enabling the use of absolute paths (to avoid `../` hell).
- Includes collateral improvements for some existing helper functions.
- unrelated: updates the Elasticsearch docker image version in the `docker-compose` file.


caravinci (commented):
Base comments for the Data Fetcher changes

Dockerfile
```diff
@@ -1,22 +1,29 @@
...
+ARG APP_GID=1000
+ARG APP_UID=1000
+ARG APP_USER=node
+ARG ES_HOST=http://localhost:9200
+
 #######################################################
 # Builder
 #######################################################
-FROM node:13.13.0-alpine as builder
+FROM node:16-alpine as builder
```
caravinci, line 10: of note: Arranger should be worked with using Node v16+

Jenkinsfile
```diff
@@ -43,16 +43,16 @@ spec:
...
-            steps {
-                container('docker') {
-                    sh "printenv; id; cat /etc/passwd"
-                }
-                container('node') {
-                    sh "printenv; id; cat /etc/passwd"
-                }
-            }
-        }
+        // stage('Diagnostics') {
```
caravinci, line 46: these add noise in the jenkins console, that though could be useful while troubleshooting, are not necessary for most builds

modules/components/package.json
```diff
@@ -3,19 +3,23 @@
...
     "build-storybook": "build-storybook -s ./public",
+    "cleanDist": "rm -rf dist",
     "dashboard": "STORYBOOK_DASHBOARD=true start-storybook -p 6060 -c .storybook -s ./public",
-    "flow": "flow",
+    "declarations": "ttsc --emitDeclarationOnly",
     "lint": "eslint src/**/*.js",
     "portal": "STORYBOOK_PORTAL=true start-storybook -p 7070 -c .storybook -s ./public",
-    "prepare": "rm -rf dist && babel src --out-dir dist --copy-files --extensions \".tsx,.ts,.js\"",
+    "postpack": "npm run declarations",
+    "prepare": "npm run cleanDist && npm run build",
```
caravinci, line 19: separated the type definitions to a packaging script, so the errors currently coming up from Flow code won't crash the bootstraping of modules.

modules/components/src/utils/uri/__tests__/uri.test.js
```diff
@@ -1,6 +1,4 @@
-/* @flow */
-
-import { parseIntParam, parseSQONParam } from '..';
+import { parseIntParam, parseSQONParam } from '@/utils/uri';
```
caravinci, line 1: tests being executed from the monorepo's root are unable to handle the relative path correctly.

modules/server/package.json
```diff
@@ -36,7 +36,7 @@
     "chalk": "^2.3.1",
     "convert-units": "^2.3.4",
     "cors": "^2.8.4",
-    "date-fns": "^1.29.0",
+    "date-fns": "^2.28.0",
```
caravinci, line 39: long over-due. was wreaking havoc with some seriously outdated methods

modules/server/src/middleware/constants.js
```diff
@@ -27,7 +27,7 @@ export const OP_ALIASES = {
   '=': IN_OP,
   '!=': NOT_IN_OP,
 };
-export const DATE_FORMAT = 'YYYY-MM-DD';
+export const DATE_FORMAT = 'yyyy-MM-dd';
```
caravinci, line 30: change in format to comply with current ISO requirements.

caravinci (commented):
Adding a few fixes/updates in order to move forward with the `test` jenkins stage. 
Changes seem unwieldy, but are mostly just dependency updates in a few manifest locks, and updating a helper that was badly broken.

PR #801, opened 2022-10-27
Feat/drop id field from mapping

andricDu:
Allows arranger to connect to an index that makes use of an `id` field. 

Drops the `id` field from mapping object so it does not cause a conflict in gql. A better more long term solution is mapping `id` to something else that is searchable and viewable should be decided by project ownership. 

image


joneubank (changes requested):
Is good solution for now, thank you. Unblocks people without modifying existing functionality.

Can we remove the .vscode specific file (`.vscode/launch.json`). Generally we haven't included IDE specific configs in the repos. This looks like the default auto generated launch file anyways so hopefully its no great loss.

andricDu:
@joneubank done!

joneubank (approved):
A+ ⭐

Issue #809, opened 2023-01-10
BUG - __missing__ has the same behavior with IN or SOME_NOT_IN operators 


creativeyann17:
## Description
Arranger 2.x build the same ES query for a `__missing__` variable whatever the operator is `in` or `some-not-in`

## Expected Behaviour
`"value":["__missing__"]},"op":"some-not-in"}`
```
"bool": {
    "must": [
        {
            "exists": {
                "field": "clinvar.clin_sig",
                "boost": 0
            }
        }
    ]
}
```

## Actual Behaviour
```
"bool": {
    "must_not": [
        {
            "exists": {
                "field": "clinvar.clin_sig",
                "boost": 0
            }
        }
    ]
}
```

## Possible Fix

[`middleware/dist/buildQuery/index.js`](https://github.com/overture-stack/arranger/blob/v2.19.2/modules/middleware/src/buildQuery/index.js#L131)

The `isNot` param should be `true` only if we want the field to be missing like this:

`isNot: op === _constants.IN_OP,`

## Steps to Reproduce
```
query MyQuery(
  $sqon: JSON
  $first: Int
  $offset: Int
  $sort: [Sort]
) {
  Index {
    hits(filters: $sqon, first: $first, offset: $offset, sort: $sort) {
      total
      edges {
        node {
          id
         foo
       }
    }
  }
}
```
```
{
    "first": 20,
    "offset": 0,
    "sqon": {
        "content": [
            {
                "content": {
                    "field": "foo",
                    "index": "index",
                    "value": [
                        "__missing__"
                    ]
                },
                "op": "some-not-in"
            }
        ]
    },
    "sort": [ ]
}
```
## Your Environment
* Version used: 2.16.2
* Environment name and version: chrome 108.0.5359.124 (Official Build) (64-bit), node: v18.1.0
* Operating System and version: Ubuntu 22.04 
* Link to your project: Ferlab

PR #811, opened 2023-01-17
fix: 809 - missing with not operator

creativeyann17:
Related issue: https://github.com/overture-stack/arranger/issues/809


caravinci (approved):
Nice catch, and thanks for your contribution, @creativeyann17!
Took us a minute to understand the nuances of the negation upon negation in the logic (the probable culprit for this bug being there to begin with), and it does seem your PR covers all relevant cases for this scenario.

The main conclusion here is: we should revisit the filter generation logic into something easier to maintain 😆.

PR #818, opened 2023-02-14
filter multi select dropdown

leoraba:
Adding filter to Column Select Dropdown.
Enabled by default.


caravinci (approved):

caravinci:
Squash that thing 🚢

PR #824, opened 2023-03-23
temp fix quicksearch props

leoraba:


modules/components/src/QuickSearch/QuickSearch.js
```diff
@@ -39,39 +39,45 @@ const toggleValue = ({ sqon, setSQON, primaryKeyField, primaryKey }) =>
...
-	PinnedValueComponent = SQONBubble,
-	placeholder = 'Quick Search',
-	searchLowercase = false,
-	searchTextDelimiters = ['\\s', ','],
-	setSQON,
-	sqon,
-	style = emptyObj,
-	...props
-}) => {
+const QuickSearch = ({ ...props }) => {
```
headsoup, line 42: does this affect any other components using the QuickSearch? (not sure how the legacy stuff behaves, as we don't have storybook up and running)

modules/components/src/QuickSearch/QuickSearch.js
```diff
@@ -104,6 +110,22 @@ const QuickSearch = ({
 
 									return (
 										<>
+											<InputComponent
```
leoraba, line 113: changing position of InputComponent

modules/components/src/QuickSearch/QuickSearch.js
```diff
@@ -39,39 +39,45 @@ const toggleValue = ({ sqon, setSQON, primaryKeyField, primaryKey }) =>
...
 
 	return (
-		<QuickSearchWrapper displayName={headerTitle} className={className} style={style}>
+		<QuickSearchWrapper
+			displayName={themeQuickSearchWrapperHeaderTitle}
+			className={themeQuickSearchWrapperClassName}
+			style={themeQUickSearchWrapperStyle}
+		>
 			<QuickSearchFieldsQuery
 				{...props}
```
headsoup, line 82: following up with the previous comment, if anyone passes `theme` into the parent component, it'll be passed down here. It may be a good idea for now to just remove that `props` spread altogether, and we can rebuild this thing as needed later.

headsoup, line 82:
```suggestion
```

PR #834, opened 2023-06-13
fix duplicates in arranger downloads

leoraba:
portal issue [#392](https://github.com/cancogen-virus-seq/portal/issues/392)

### DIFF
--- a/modules/server/src/utils/getAllData.js
+++ b/modules/server/src/utils/getAllData.js
@@ -81,10 +81,10 @@ export default async ({
 							...(Object.entries(query).length ? { query } : {}),
 						},
 					})
-					.then(({ body }) => mapHits(body));
+					.then(({ body }) => body.hits.hits);
 				console.timeEnd(timerLabel);
 
-				stream.write({ hits, total });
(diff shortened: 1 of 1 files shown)


modules/server/src/utils/getAllData.js
```diff
@@ -81,10 +81,10 @@ export default async ({
 							...(Object.entries(query).length ? { query } : {}),
 						},
 					})
-					.then(({ body }) => mapHits(body));
+					.then(({ body }) => body.hits.hits);
 				console.timeEnd(timerLabel);
 
-				stream.write({ hits, total });
+				stream.write({ hits: hits.map((hit) => hit?._source), total });
```
justincorrigible, line 87: the changes described here serve the exact same purpose as that `mapHits` function.
what's the trick here? I don't see it.

leoraba: Moving the `mapHits` logic out of the variable `hits` will leave the attributes required for sorting to be used in the next cycle by `previousHits`

justincorrigible (approved):

PR #871, opened 2024-05-31
Search docs

ciaranschutte:


design/search.md
```diff
@@ -0,0 +1,300 @@
...
+| status | node status eg. "connected" |
+
+## Request Nodes
+Outline:
+- GQL HTTP calls to endpoints
+- Specify only returning "arrangerField" in gql query
+- Handles network level errors eg. node not found
+
+Query step:
+- Read `aggregations` field which Arranger generates by default (`aggsState` isn't always generated - only for UI?)
```
ciaranschutte, line 38: not 100% sure on the `aggregations` vs `aggsState` yet. 
`aggsState` doesn't appear to be generated by default when I ran the setup locally with Overture docs
there are other places in nested nodes where aggregations surface

design/search.md
```diff
@@ -0,0 +1,300 @@
+
+
+# Federated Search
+![VPCzJyCm48Pt_ufJftP0x1bGAuIgIhG22XDYEE9hQicnW-qW-FVu8qT2HTlfSkzpxtrONVg0BlIj5bW7ww3yNZnnY3v_YIvYgbOTcW2pbNDe6d8LR56PMOHoS0wwjUQWceoLy1ou9tJrCOCbl0oEninVjjzPIJ1trDf0YrILCqA8lE_LJLu2edkw2N1Tr7C-wiM-WYVwwCa7gFCt79GcKRI_5Ce1yV2h3stuAcpEsOrH (1)](https://github.com/overture-stack/arranger/assets/1486054/87bb0689-38c6-49eb-ab65-96696a04d86a)
+
+## Prerequisites
+- Local data must be configured
+- ENV flag passed into process
```
joneubank, line 8: what env flag? what does the flag set/enable?

design/search.md
```diff
@@ -0,0 +1,300 @@
+
+
+# Federated Search
+![VPCzJyCm48Pt_ufJftP0x1bGAuIgIhG22XDYEE9hQicnW-qW-FVu8qT2HTlfSkzpxtrONVg0BlIj5bW7ww3yNZnnY3v_YIvYgbOTcW2pbNDe6d8LR56PMOHoS0wwjUQWceoLy1ou9tJrCOCbl0oEninVjjzPIJ1trDf0YrILCqA8lE_LJLu2edkw2N1Tr7C-wiM-WYVwwCa7gFCt79GcKRI_5Ce1yV2h3stuAcpEsOrH (1)](https://github.com/overture-stack/arranger/assets/1486054/87bb0689-38c6-49eb-ab65-96696a04d86a)
+
+## Prerequisites
+- Local data must be configured
```
joneubank, line 7: This was a bit confusing cause I read it assuming it meant local developer setup should have data available, but now I assume this is "local" in the context of "local search" vs "network search". This is a bit unclear.

Perhaps it could be phrased like `Local Search configuration of this arranger node is performed separately from Network Search`, since we do not require a local search config for network search

design/search.md
```diff
@@ -0,0 +1,300 @@
...
+## Config
+- Read "search config" from file
+- Export object containing node data for use in gql endpoint
+
+Example Config provided:
+
+| Field         | Type   | Note                                                                                                                            |
+| ------------- | ------ | ------------------------------------------------------------------------------------------------------------------------------- |
+| endpoint      | string | Full url to gql endpoint                                                                                                        |
+| arrangerField | string | Endpoints can have other data on the root object. Need to get specific Arranger config object. Appears to be "file" by default. |
```
joneubank, line 20: I don't think it is `file` by default, but this is common for setups that use maestro to index `files` from song.

design/search.md
```diff
@@ -0,0 +1,300 @@
...
+
+Example Search Node data:
+
+| Field  | Description                 |
+| ------ | --------------------------- |
+| url    | gql endpoint                |
+| name   | name of node eg. Toronto    |
```
joneubank, line 27: Seeing this here, we may want to include a `name` property to our node config. Arranger nodes aren't currently reporting a name of their own, so this would help us label each node in our responses.

design/search.md
```diff
@@ -0,0 +1,300 @@
...
+| schema | version of Arranger running |
+| status | node status eg. "connected" |
```
joneubank, line 29: 👍 
We should list the possible statuses and what they mean.

joneubank (commented):
This is a good start at capturing our process. I wouldn't mind an example of the config file, and I think we should consider adding a `name` property to each search node in that config.

It would be good to add a bit of description at the start describing how arranger will support local and network search and that this document describes how network search is setup.

ciaranschutte:
@ciaranschutte mermaid diag

PR #876, opened 2024-06-05
Network aggregation search - Initial Config

ciaranschutte:
(merging into feature branch)
initial config for network aggregation search
- reads config from file using the same process as other config files
- env var conditional to add additional generated schema to gql endpoint


modules/server/src/config/types.ts
```diff
@@ -143,6 +150,12 @@ export interface TableConfigsInterface {
 	[ConfigProperties.ROW_ID_FIELD_NAME]?: string;
 }
 
+export interface NetworkAggregationInterface {
+	[NetworkAggregationProperties.ENDPOINT]: string;
+	[NetworkAggregationProperties.ARRANGER_FIELD]: string;
+	[NetworkAggregationProperties.NAME]: string;
+}
+
```
ciaranschutte, line 158: Might need to expand on the config object to be nested because it's merged into a single object. a key like `name` isn't very unique and there isn't a unique field guard

modules/server/src/graphqlRoutes.js
```diff
@@ -249,19 +251,35 @@ export const createSchemasFromConfigs = async ({
...
+				schema: mergedSchema,
+				mockSchema: mergedMockSchema,
+			};
+		} else {
+			return {
+				...commonFields,
+				mockSchema,
+				schema,
+			};
+		}
```
ciaranschutte, line 282: personally preferring the readability of verbose returns instead of adding `let` variables to conditional assign values, then more inline in a return statement like: `return {schema: mergedSchema || schema}`.

modules/server/src/network/index.ts
```diff
@@ -0,0 +1,2 @@
+export const createSchemaFromNetworkConfig = () => {};
+export const mergeSchemas = () => {};
```
ciaranschutte, line 2: use TS as early as possible.

modules/server/src/config/types.ts
ciaranschutte: structure is dependant on our json config files. no code enforcement. maybe a future improvement using zod to match a config schema 
eg. if a field is nested, it will be nested in config

modules/server/src/network/index.ts
```diff
@@ -0,0 +1,10 @@
+import { NetworkAggregationInterface } from '@/config/types';
+
+export const createSchemaFromNetworkConfig = ({
+	networkConfig,
+}: {
+	networkConfig: NetworkAggregationInterface[];
+}) => {
+	console.log('network config', networkConfig);
+};
+export const mergeSchemas = () => {};
```
ciaranschutte, line 10: placeholders.

modules/server/src/config/utils/getConfigFromFiles.ts
```diff
@@ -56,23 +56,29 @@ const getConfigFromFiles = (
 
 			const configObj = (files as [string, any][]).reduce(
 				(configsAcc: Partial<ConfigObject>, [fileName, fileData]) => {
-					const fileDataJSON = JSON.parse(fileData);
+					try {
+						const fileDataJSON = JSON.parse(fileData);
```
ciaranschutte, line 60: failure can be empty file

modules/server/README.md
```diff
@@ -18,3 +18,9 @@ Build (required before publish):
 ```
 npm run prepare
 ```
+
+## Federated Search
+
+### Config
+
+Pass `true` to the environment variable `ENABLE_NETWORK_AGGREGATION` to enabled federated search.
```
justincorrigible, line 26: quick typo. however, I'm thinking how this will need to be extended into profiles, rather than on/off (boolean)

```suggestion
Pass `true` to the environment variable `ENABLE_NETWORK_AGGREGATION` to enable federated search.
```

modules/server/configTemplates/base.json
```diff
@@ -1,4 +1,4 @@
 {
-	"documentType": "",
-	"index": ""
+	"documentType": "file",
+	"index": "file_centric_1.0"
```
justincorrigible, line 3: Please leave the `templates` empty. These will be used later on for automated config generation.

```suggestion
	"documentType": "",
	"index": ""
```

modules/server/configTemplates/network.json
```diff
@@ -0,0 +1,9 @@
+{
+	"network": [
+		{
+			"name": "Toronto",
+			"endpoint": "www",
+			"arrangerField": "file"
```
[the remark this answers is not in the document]
ciaranschutte: I think `documentType` works better

modules/server/README.md
ciaranschutte: Not 100% clear on what profiles entails. 
For this PR I think its ok to just have something there to say "this functionality is based on env"
let me know if I can create a ticket re profiles

ciaranschutte:
investigating ci issues

justincorrigible (commented):
Nice first stab. Definitely starts putting things in place at the right spots.

Reading through it three "main" thoughts came to me, we should discuss further:
- how to handle the "runtime profile" definitions
- configuration model
- configuration validation, with Zod (surely @joneubank will want to be involved here)

modules/server/src/graphqlRoutes.js
```diff
@@ -249,19 +250,35 @@ export const createSchemasFromConfigs = async ({
...
 			types: typesWithMappings,
 		});
 
-		return {
-			fieldsFromMapping,
-			mockSchema,
-			schema,
-			typesWithMappings,
-		};
+		if (false) {
```
ciaranschutte, line 262: temp until profile config is added

ciaranschutte:
Runtime profiles to be added in follow up PR
Config validation to be discussed, scoped etc before implementing

justincorrigible (approved):
Squash away :shipit:

PR #877, opened 2024-06-05
Feat/#873 retrieve remote schemas

ciaranschutte:
- retrieval of remote schemas
- responses build to full schema using gql tools (means we can use the well tested inbuilt methods without manually parsing a gql request as intermediary objects)
- output is array of full schemas as GraphQlObjects 

Please note this is part of the bigger schema federation work merging into feat/featured_search
The broken build issues are for some missing TS types, implicit `any`s
These will be addressed before merging full feature into `develop` branch


modules/server/src/network/index.ts
```diff
@@ -1,10 +1,73 @@
...
+	networkConfig: NetworkAggregationConfigInput[];
+}): Promise<NetworkAggregationConfig[]> => {
+	// query remote notes
+	// type cast because rejected errors are handled
+	const networkQueries = (await Promise.allSettled(
+		networkConfig.map(async (config) => fetchRemoteSchema(config)),
+	)) as PromiseFulfilledResult<FetchRemoteSchemaResult>[];
+
+	// build schema
+	const schemaResults = networkQueries.map((networkResult) => {
```
ciaranschutte, line 45: keep separate from network so we can write tests

modules/server/src/network/index.ts
```diff
@@ -1,10 +1,73 @@
...
+			.then(toJSON)
+			.then((json) => json.data);
+
+		return { config, schemaJSON: remoteSchema };
+	} catch (error) {
+		console.log(`Failed to retrieve schema from url: ${config.graphqlUrl}`);
+		console.error(error);
+		return { config, schemaJSON: null };
+	}
+};
```
ciaranschutte, line 31: return `schema: null` on error and log errors
need all queries to run even if errors

.gitignore
```diff
@@ -8,3 +8,5 @@ lerna-debug.log
 **/.DS_Store
 **/*.env*
 !**/.env.schema
+
+.ignore/
```
ciaranschutte, line 12: does what it says on the tin

joneubank: Is this something you think the project needs due to an expected folder that our code generates? Or are you using the committed .gitignore to ignore files in your own local setup? If the latter, may I suggest a local file in your .git directory: https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#excluding-local-files-without-creating-a-gitignore-file

ciaranschutte:
change `then` chaining to `awaits`

modules/server/src/network/index.ts
joneubank: This kind of information goes nicely into a TSDoc comment before the funciton. Let users know what happens on successful and unsuccessful fetches.

modules/server/src/network/index.ts
```diff
@@ -1,10 +1,96 @@
-import { NetworkAggregationInterface } from '@/config/types';
+import { buildClientSchema, getIntrospectionQuery, IntrospectionQuery } from 'graphql';
+import { fetchGql } from './gql';
+import { NetworkAggregationConfig, NetworkAggregationConfigInput } from './types';
 
-export const createSchemaFromNetworkConfig = ({
+const isGqlIntrospectionQuery = (input: any): input is IntrospectionQuery => {
+	return (input as IntrospectionQuery).__schema !== undefined;
```
ciaranschutte, line 6: this is the top level field that is defined on gql spec getIntrospectionQuery return type

modules/server/src/graphqlRoutes.js
```diff
@@ -259,7 +259,7 @@ export const createSchemasFromConfigs = async ({
 			types: typesWithMappings,
 		});
 
-		if (false) {
+		if (true) {
 			const { networkSchema } = await createSchemaFromNetworkConfig({
 				networkConfig: configsFromFiles[ConfigProperties.NETWORK_AGGREGATION],
```
ciaranschutte, line 264: placeholder right now - merging into feature branch - will depend on "mode" configuration

.gitignore
```diff
@@ -7,4 +7,4 @@ lerna-debug.log
 **/node_modules
 **/.DS_Store
 **/*.env*
-!**/.env.schema
+!**/.env.schema
```
justincorrigible, line 10:
```suggestion
!**/.env.schema

```

modules/server/src/graphqlRoutes.js
```diff
@@ -259,7 +259,7 @@ export const createSchemasFromConfigs = async ({
 			types: typesWithMappings,
 		});
 
-		if (false) {
+		if (true) {
```
justincorrigible, line 262: at this point, might well remove the "if", no?

modules/server/src/network/gql.ts
```diff
@@ -0,0 +1,18 @@
+import { print } from 'graphql';
+
+export const createGqlQuery = ({ query: queryAST, variables }) => {
+	// queries need to be strings when using regular http
+	const query = print(queryAST);
+	return JSON.stringify({ query, variables });
```
justincorrigible, line 6: no validation needed? or is that ToDo?

modules/server/src/network/gql.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const fetchGql = ({ url, gqlRequest }) => {
+	const query = JSON.stringify(gqlRequest);
+	const options = {
+		method: 'POST',
+		headers: { 'Content-Type': 'application/json' },
+		body: query,
+	};
+
+	return fetch(url, options);
```
justincorrigible, line 17: I'd prefer we used Axios, to keep things consistent with the rest of the server (unless I'm mistaken, which is always possible/likely).

Open to discuss, though!

modules/server/src/network/index.ts
justincorrigible: these "is" and "as" always get me mixed up... may I ask for a bit of explanation on the intent here?
e.g. there must be a reason you're using "any" instead of the desired type.

modules/server/src/network/index.ts
```diff
@@ -1,10 +1,93 @@
...
+const fetchRemoteSchemas = async ({
+	networkConfig,
+}: {
+	networkConfig: NetworkAggregationConfigInput[];
+}): Promise<NetworkAggregationConfig[]> => {
+	// query remote notes
+	// type cast because rejected errors are handled
+	const networkQueries = (await Promise.allSettled(
+		networkConfig.map(async (config) => fetchRemoteSchema(config)),
+	)) as PromiseFulfilledResult<FetchRemoteSchemaResult>[];
```
justincorrigible, line 61: wow! there's a lot going on here! any chance this can be chunked out into separate steps?

modules/server/src/network/types.ts
```diff
@@ -0,0 +1,13 @@
...
+// environment config
+export type NetworkAggregationConfigInput = {
+	graphqlUrl: string;
+	documentType: string;
+	displayName: string;
+};
+
+// additional properties after initialising network
+export type NetworkAggregationConfig = NetworkAggregationConfigInput & {
+	schema: GraphQLSchema | null;
```
justincorrigible, line 12: have a feeling Jon won't like `null` (I'm a fan of it, though)

modules/server/src/network/gql.ts
ciaranschutte: the actual queries will be written when we write the resolvers
this function is just to go from gqlObject to string request for gql
made more sense to keep it in for future usage

ciaranschutte: originally I was creating handmade queries to get the schemas, but I found an inbuilt gql method to get an introspection query

modules/server/src/network/gql.ts
ciaranschutte: makes sense! easy one

modules/server/src/network/index.ts
ciaranschutte: https://github.com/overture-stack/arranger/pull/877/commits/050bf5130ddc819d4a2d3a33b36d492f380276b1

ciaranschutte:
https://github.com/overture-stack/arranger/pull/877#discussion_r1635107541

can't seem to comment on this one inline...

using a type predicate so our response data is not `any` type
specifically so that our `buildClientSchema` param is checked before using it
we can add more to the predicate to ensure it's an "arranger instance" but right now there doesn't seem to be anything that specifically defines something as an arranger instance versus just a Gql endpoint. end user can always alter schema.

as I understand it , the `as` is just making sure you can do the property check without that type erroring (no need for TS to type check in a type predicate as you're taking the reigns of the typechecking)
and then checking for the property `__schema` is how we're telling TS that we've made sure the object is an `introspectionGqlQuery`

ref: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

modules/server/src/network/types.ts
ciaranschutte: I feel the same way. In this case I think it makes sense.

We've tried and failed to get a schema for a remote connection.
Errors have been caught and logged, but aren't fatal, doesn't stop other remote connections.
We still want to know that we have this connection in our env, we tried to connect + process, but no schema (versus processing/attempts still to complete ie. undefined)

justincorrigible: sound precisely as my own line of thought: I like using `null` as a deliberate value to differentiate from just "missing".

modules/server/src/network/index.ts
ciaranschutte: https://github.com/overture-stack/arranger/pull/877#issuecomment-2161111471

modules/server/src/graphqlRoutes.js
ciaranschutte: I'm going to leave this in. With the upcoming "mode" configuration , this will still be an `if`
We won't add any network aggregation functionality if config is just normal mode

modules/server/src/network/index.ts
```diff
@@ -35,16 +35,16 @@ const fetchRemoteSchema = async (
...
 
-		if (response.status !== 200) {
-			throw Error('network error');
-		}
-
-		const jsonData = await response.data.json();
-
-		if (jsonData && jsonData.data && isGqlIntrospectionQuery(jsonData.data)) {
-			return { config, introspectionResult: jsonData };
+		if (response.status === 200 && response.statusText === 'OK') {
```
justincorrigible, line 38: Fetch hasn't ever not relied on negatives... Affirmative logic FTW. #accessibility

modules/server/src/network/index.ts
```diff
@@ -35,16 +35,16 @@ const fetchRemoteSchema = async (
...
+			// axios response "data" field, graphql response "data" field
+			const responseData = response.data?.data;
+			if (isGqlIntrospectionQuery(responseData)) {
```
justincorrigible, line 41:
```suggestion

			if (isGqlIntrospectionQuery(responseData)) {
```

modules/server/src/network/index.ts
```diff
@@ -35,16 +35,16 @@ const fetchRemoteSchema = async (
...
-			throw Error('network error');
-		}
-
-		const jsonData = await response.data.json();
-
-		if (jsonData && jsonData.data && isGqlIntrospectionQuery(jsonData.data)) {
-			return { config, introspectionResult: jsonData };
+		if (response.status === 200 && response.statusText === 'OK') {
+			// axios response "data" field, graphql response "data" field
+			const responseData = response.data?.data;
```
justincorrigible, line 40: Glad we're abstracting this within a function, and I wonder if there's a way to remove the second `.data` from the responses (since it's "likely" an Arranger thing anyway)

modules/server/src/network/index.ts
```diff
@@ -35,16 +35,16 @@ const fetchRemoteSchema = async (
...
+			if (isGqlIntrospectionQuery(responseData)) {
+				return { config, introspectionResult: responseData };
+			} else {
+				throw Error('response data unexpected');
+			}
 		} else {
-			throw Error('response data unexpected');
+			throw Error('network error');
```
justincorrigible, line 47: when I throw at any point in the logic, I tend to also log.debug out what function is the one that's throwing
...hopefully in another PR later on we can add winston/equivalent as discussed over Slack

modules/server/src/network/index.ts
ciaranschutte: it's a GQL thing

justincorrigible: GraphQL: here, have some data in your data
image

modules/server/src/graphqlRoutes.js
joneubank: Probably want to drop a `// TODO` here then just to track the expected change. In this very obvious case I don't think we'll lose track that a change is needed, but its useful to let people reading/helping with reviews know what the plan is for this.

modules/server/src/network/types.ts
joneubank: In the process of validating `undefined` vs `null` you are both discussing states. Specifically I have seen the following states described:
- schema available (has type `GraphQLSchema`)
- schema not available (has type `null`)
- an attempt to fetch the schema has not been completed (has type `undefined`)

So using this logic, we are encoding stateful information about our fetch history into the `NetworkAggregationConfig.schema` type. When future devs interact with this type, none of this state information is being communicated, they just know what types the property could be. If we want to convey important state information I would suggest expanding your type to include either a `state` property, and maybe even use differentiated unions (see below).

More immediately relevant here though is that we don't have 3 states, we only have 2 - value present and value missing. TS has a built in convention for optional properties: `schema?: GraphQLSchema`. For this reason alone, I would encourage using undefined by default unless there is a need to introduce `null` into the code - the most common reason is that "external API uses null types" - this is relevant for GQL interfaces, but not here.

### Example
Example for encoding many states with differentiated union, not saying we need this:
```ts
type RemoteNodeSchemaAvailable = {
  state: 'available',
  schema: GraphQLSchema
}

type RemoteNodeSchemaUnavailable = {
  state: 'unavailable',
  reason: string
}

type RemoteNodeSchemaFetching = {
  state: 'fetching',
  start: Date
  timeout: Date
}

type RemoteNodeSchema = RemoteNodeSchemaAvailable | RemoteNodeSchemaUnavailable | RemoteNodeSchemaFetching;

type NetworkAggregationConfig = NetworkAggregationConfigInput & RemoteNodeSchema;
```

joneubank: One last thought here: there isn't much practical difference in most case. Code that consumes this type is mostly going to do like:

```ts
if(config.schema) {
  // do stuff with schema
} else {
  // handle case without it
}
```

or otherwise filter out the missing case. If this is your expectation, what have we added to our code base by mixing the use of `undefined` and `null`.

modules/server/src/network/index.ts
```diff
@@ -1,10 +1,111 @@
...
+}: {
+	networkConfig: NetworkAggregationConfigInput[];
 }) => {
-	console.log('network config', networkConfig);
+	const remoteSchemasResult = await fetchRemoteSchemas({
+		networkConfig,
+	});
+
+	const mergedFields = mergeRemoteSchemas(remoteSchemasResult);
+};
```
joneubank, line 107: No return statement? When I know what type a function needs to return I like to specify this in the signature. This does two things:

1. TS will enforce that your returns are the correct type
2. Future devs can see at a glance what the function is supposed to return.

I think a function named "createXYZConfig" is a good candidate for this, it should state:

```ts
export const createSchemaFromNetworkConfig = async(inputs: {networkConfigs: NetworkAggregationConfigInput[] }): NetworkSearchSchema => {

```

modules/server/src/network/index.ts
```diff
@@ -1,10 +1,111 @@
...
+
+const mergeRemoteSchemas = (input) => {
+	/**placeholder */
```
joneubank, line 110: This placeholder is telling other devs reading this nothing about this function, I would want more information here before being able to review if this is the correct step to use. As a placeholder, lets at least explicitly declare `(input: any): any => {` so TS will not complain about implicit any.

It will benefit this entire process to be strict with our types. Define the expected inputs. Declare the expected output. Start with the type signatures and everything else will just be the logic to convert inputs to outputs. If the specific types are still unclear, than use the `any` suggestion above. In either case, give me a function comment or description to explain the placeholder.

joneubank (changes requested):

modules/server/src/network/index.ts
```diff
@@ -1,10 +1,111 @@
...
+	// query remote notes
+	const networkQueryPromises = networkConfig.map(async (config) => fetchRemoteSchema(config));
+	// type cast because rejected errors are handled
+	const networkQueries = (await Promise.allSettled(
+		networkQueryPromises,
+	)) as PromiseFulfilledResult<FetchRemoteSchemaResult>[];
+
+	// build schema
+	const schemaResults = networkQueries.map((networkResult) => {
+		const { config, introspectionResult } = networkResult.value;
```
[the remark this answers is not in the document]
ciaranschutte: that's how the code is currently handling

ciaranschutte:
@joneubank addressed feedback. TS building. please take another look when you have some time.

modules/server/src/network/gql.ts
```diff
@@ -1,13 +1,26 @@
...
+ */
+export const createGqlQuery = ({
+	query: queryAST,
+	variables,
+}: {
+	query: ASTNode;
+	/**
+	 * TODO: Add typing for variables. Type will be dependant on gql endpoint being queried.
+	 */
+	variables: unknown;
```
joneubank, line 16: Looking at how this is used, we don't really care about the specific contents of the variables. This function is doing no work to ensure that the variables expected in the `query` are included in `variables` (or vice versa). The only thing we know is we want to be able to run `JSON.stringify(variables)` which takes `any`. In practice, we expect this parameter to be an object with some variable values which themselves could be anything, and we probably don't want it to be `string` or `number` etc.  So in this case I might reccomend:

```ts
variables: Record<string, any>
```

modules/server/src/network/gql.ts
```diff
@@ -1,13 +1,26 @@
 import axios from 'axios';
-import { print } from 'graphql';
+import { ASTNode, print } from 'graphql';
 
-export const createGqlQuery = ({ query: queryAST, variables }) => {
+/**
+ * Creates a graphql query string with variables for use in a POST request
+ * @returns string
+ */
+export const createGqlQuery = ({
```
joneubank, line 8: Noticed this function isn't actually used in your code here, is that expected?

modules/server/src/network/gql.ts
```diff
@@ -1,13 +1,26 @@
...
+	 */
+	variables: unknown;
+}) => {
 	// queries need to be strings when using regular http
 	const query = print(queryAST);
 	return JSON.stringify({ query, variables });
 };
 
-export const fetchGql = ({ url, gqlRequest }) => {
+export const fetchGql = ({ url, gqlRequest }: { url: string; gqlRequest: { query: string } }) => {
```
joneubank, line 23: TSDoc will be wanted before feature branch is complete.

modules/server/src/network/index.ts
```diff
@@ -18,7 +19,7 @@ const isGqlIntrospectionQuery = (input: unknown): input is IntrospectionQuery =>
 
 type FetchRemoteSchemaResult = {
 	config: NetworkAggregationConfigInput;
-	introspectionResult: IntrospectionQuery | null;
+	introspectionResult: IntrospectionQuery | undefined;
```
joneubank, line 22: why not just:
```ts
introspectionResult?: IntrospectionQuery
```

modules/server/src/network/index.ts
```diff
@@ -53,59 +54,63 @@ const fetchRemoteSchema = async (
 				return { config, introspectionResult: responseData };
 			} else {
 				console.error('unexpected response data in fetchRemoteSchema');
-				throw Error('unexpected response data');
+				throw new NetworkAggregationError(
+					`Unexpected data in response object. Please verify the endpoint at ${graphqlUrl} is returning a valid GQL Schema.`,
```
joneubank, line 58: 👍

modules/server/src/network/index.ts
```diff
@@ -1,10 +1,116 @@
...
+			console.error('network error in fetchRemoteSchema');
+			throw new NetworkAggregationError(
+				`The request to endpoint ${graphqlUrl} has failed. Please verify the endpoint is accessible and responding to requests.`,
+			);
+		}
+	} catch (error) {
+		console.log(`failed to retrieve schema from url: ${config.graphqlUrl}`);
+		console.error(error);
+		return { config, introspectionResult: undefined };
+	}
```
joneubank, line 71: Axios throws errors on status 4xx and 5xx. That means that the else block you have above is behaviour that you are likely expecting to be handled here as well. You should try checking if the error here is an instance of AxiosError, and if so you can handle these error statuses in the catch block. I imagine you will want to throw your NetworkAggregationError.

modules/server/src/network/index.ts
```diff
@@ -1,10 +1,116 @@
...
+	const networkQueryPromises = networkConfigs.map((config) => fetchRemoteSchema(config));
+
+	const networkQueries = await Promise.allSettled(networkQueryPromises);
+
+	// build schema
+	const schemaResults = networkQueries
+		.filter(
+			(result): result is PromiseFulfilledResult<FetchRemoteSchemaResult> =>
+				result.status === 'fulfilled',
+		)
```
joneubank, line 89: With TS 5.5 we no longer require the `result is` return type :D

joneubank (commented):
Looking strong throughout.

The axios request failure handling I think is not what you would expect to see so consider that comment and if there are changes needed.

modules/server/src/network/gql.ts
ciaranschutte: expected. https://github.com/overture-stack/arranger/pull/877#discussion_r1635117424

modules/server/src/network/index.ts
```diff
@@ -1,10 +1,116 @@
...
+		} else {
+			console.error('network error in fetchRemoteSchema');
+			throw new NetworkAggregationError(
+				`The request to endpoint ${graphqlUrl} has failed. Please verify the endpoint is accessible and responding to requests.`,
+			);
+		}
+	} catch (error) {
+		console.log(`failed to retrieve schema from url: ${config.graphqlUrl}`);
+		console.error(error);
+		return { config, introspectionResult: undefined };
```
joneubank, line 70: PS. don't need to do `introsepectionResult: undefined` if you have the type defined as `introspectionResult?: IntrospectionQuery`

modules/server/src/network/gql.ts
joneubank: I think your TSDoc is incorrect, instead of returning an axios instance that you can use to send future requests, this sends your defined request with axios and returns the AxiosResponse (or throws error, stupid axios).

modules/server/src/network/types.ts
```diff
@@ -0,0 +1,13 @@
+import { GraphQLSchema } from 'graphql';
+
+// environment config
+export type NetworkAggregationConfigInput = {
+	graphqlUrl: string;
+	documentType: string;
+	displayName: string;
+};
```
joneubank, line 8: The comment here doesn't tell me anything new. You can use TSDoc on types as well to leave a description of the type in intellisense for future devs that use it.

modules/server/src/network/types.ts
```diff
@@ -0,0 +1,13 @@
...
+
+// additional properties after initialising network
+export type NetworkAggregationConfig = NetworkAggregationConfigInput & {
+	schema?: GraphQLSchema;
+};
```
joneubank, line 13: This is a helpful comment but it wasn't easily understandable. I had to read the type information to understand what you mean. Maybe something like:


```suggestion
/**
 * Complete aggregation config that defines the network search setup. This includes the original
 * config information plus computed fields that are generated in the network search initialization process.
 */
export type NetworkAggregationConfig = NetworkAggregationConfigInput & {
	schema?: GraphQLSchema;
};
```

modules/server/src/network/gql.ts
justincorrigible: Great feedback, useful insight... but let's try keeping unhelpful personal opinions to a minimum.

modules/server/src/network/types.ts
justincorrigible: That's a nice improvement, for sure!
Naive question: am I the only one who gets bugged by the line length inconsistency?
personally, I'd chop it like this
```
* Complete aggregation config that defines the network search setup.
* This includes the original config information plus computed fields
* that are generated in the network search initialization process.
```

modules/server/src/network/index.ts
```diff
@@ -1,10 +1,121 @@
...
+ * This is doing a cursory check that the response from a GQL Server appears to be a IntrospectionQuery object.
+ * If this check passes we will optimistically treat the object as a GQL IntrospectionQuery.
+ *
+ * Should be used sparingly, and ideally all downstream consumers of the object operate within try/catch blocks.
+ *
+ * @param input
+ * @returns
+ */
+const isGqlIntrospectionQuery = (input: unknown): input is IntrospectionQuery =>
+	!!input && typeof input === 'object' && '__schema' in input && typeof input.__schema === 'object';
```
justincorrigible, line 23: wouldn't `typeof input === 'object'` inherently cover the `!!`?
(an `object` cannot be falsy)

```suggestion
	typeof input === 'object' && '__schema' in input && typeof input.__schema === 'object';
```

justincorrigible: does it matter that this function would allow `input` to be an array customised to include `__schema`?

modules/server/src/network/index.ts
```diff
@@ -1,10 +1,121 @@
...
+		)
+		.map((networkResult) => {
+			const { config, introspectionSchema } = networkResult.value;
+
+			try {
+				const schema =
+					introspectionSchema !== undefined ? buildClientSchema(introspectionSchema) : undefined;
+				return { ...config, schema };
+			} catch (error) {
+				console.error('build schema error', error);
```
joneubank, line 103: It would be nice if these messages were more descriptive. The best error messages let the reader know what was being attempted, why it failed, and (if possible) what action should be taken.

It is likely that this is about on par with other error messaging in this repo, but working to improve these as we go would be a huge help.

ciaranschutte: for sure - this is just the "retrieve schema" step
`buildClientSchema` is not implemented
will be keeping this in mind working on the "build schema" ticket

modules/server/src/network/index.ts
ciaranschutte: 1st point: 
`typeof null` is object - <3 js

ciaranschutte: 2nd point:
I don't think that's a concern. I'm using the `getIntrospectionQuery` which is part of the graphql library (conforming to spec) and this check is only happening to check response data from a GQL server.

It queries an automatically generated field in a graphql server, not a user created field.
For an array to occur the graphql spec would need to be changed.

justincorrigible: is there a `null._schema`? otherwise, `!!` is still redundant

PR #878, opened 2024-06-09
Feat/#874 merge schemas

ciaranschutte:
- adds merging remote schema functionality, uses graphql tools
- adds unit tests for merging schemas (also adds watch test command)
- merging into feature branch


modules/server/src/network/__tests__/index.test.ts
```diff
@@ -0,0 +1,122 @@
+// @ts-nocheck
+// graphql types have readonly modifiers all the way down
```
ciaranschutte, line 2: disabling TS on this file. It's a test file so it won't break the build. 
It's a larger piece of work to type because all the graphql lib types have `readonly` modifiers , which means util funcs like `pick` and `omit` don't work. The full types are too broad as we're using a specific subset of the functionality.

modules/server/src/network/__tests__/index.test.ts
```diff
@@ -0,0 +1,122 @@
...
+		(definition) => definition.kind === 'ObjectTypeDefinition' && definition.name.value === parent,
+	).fields;
+	return parentTypeDefFields.some(
+		(definitions) => (definitions.kind = 'FieldDefinition' && definitions.name.value === child),
+	);
+};
+
+/**
+ * At no point do we modify the type definitions within the tests
+ */
```
ciaranschutte, line 38: schema outside of test fixtures as we don't modify the schema inputs

modules/server/src/network/index.ts
```diff
@@ -108,14 +115,31 @@ const fetchRemoteSchemas = async ({
...
+}) => {
+	const networkConfigsWithSchemas = await fetchRemoteSchemas({
 		networkConfigs,
 	});
 
-	return remoteSchemasResult;
+	// filter remote connections with valid schemas
+	const gqlSchemas = networkConfigsWithSchemas
+		.filter((config) => config.schema !== null)
+		.map((config) => config.schema as GraphQLSchema);
```
ciaranschutte, line 135: can change this to a filter func if you'd like, I feel there's a cleaner way to do this overall.

modules/server/src/network/__tests__/index.test.ts
```diff
@@ -0,0 +1,131 @@
...
+
+	type Analysis {
+		type: String
+	}
+`;
+
+const networkSchemas = [buildSchema(typeDefsA), buildSchema(typeDefsB)];
+
+describe('network aggregation', () => {
+	test('it should have defined GQL Object types', () => {
```
joneubank, line 75: Unless this is the pattern used in tests throughout arranger, it would be nice to use proper capitalization.

modules/server/src/network/__tests__/index.test.ts
```diff
@@ -0,0 +1,131 @@
...
+			(definition) =>
+				definition.kind === 'ObjectTypeDefinition' && definition.name.value === 'Donor',
+		).length;
+		expect(numberOfResults).toEqual(1);
+	});
+
+	/**
+	 * example: multiple remote connections have the same field
+	 */
+	test('it should de-deupe FieldDefinition', () => {
```
joneubank, line 107: de-dupe typo

modules/server/src/network/__tests__/index.test.ts
```diff
@@ -0,0 +1,131 @@
...
+ * @returns true if field is defined
+ */
+const isFieldDefined = (typeDefs, parent, child) => {
+	const parentTypeDefFields = typeDefs.find(
+		(definition) => definition.kind === 'ObjectTypeDefinition' && definition.name.value === parent,
+	).fields;
+	return parentTypeDefFields.some(
+		(definitions) => (definitions.kind = 'FieldDefinition' && definitions.name.value === child),
+	);
+};
```
joneubank, line 34: Can probably move some test utilities to separate files to organize this code better.

modules/server/src/network/__tests__/index.test.ts
joneubank, line 1: Ideas for things that should be tested:
- merging many schemas (we only test the case with 2)
- no issues merging schemas with no overlapping types/fields
- all types/fields from each schema can be found in output schema

You already kind of test this last condition but didnt declare it explicitly, you just wrote that the test was checking if "types are defined".

modules/server/src/network/schema.ts
```diff
@@ -0,0 +1,21 @@
...
+ * exclusions and customising how to handle conflicts.
+ * See full documentation here: https://the-guild.dev/graphql/tools/docs/api/modules/merge_src#mergetypedefs
+ *
+ * @param networkConfigs
+ * @returns
+ */
+export const createNetworkAggregationTypeDefs = (gqlTypes: GraphQLSchema[]): DocumentNode => {
+	const options: Omit<Partial<Config>, 'commentDescriptions'> = {
+		forceSchemaDefinition: true,
+	};
```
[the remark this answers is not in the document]
ciaranschutte: removed, I think it's better to leave it out for now.

modules/server/src/network/__tests__/index.test.ts
ciaranschutte: lower case is consistent with the project
"Object" is capitalized because it's a specific GQL type not a js object

ciaranschutte:
@joneubank addressed feedback. 
- added each additional test you suggested. 
- added additional test for type collisions (currently error thrown is expected, we can discuss correct behaviour but hopefully this is enough of a test to merge this PR)
- added documentation to clarify some points
- ensured naming of tests is consistent with lower case
- created utils folder, created fixtures folder
- naming the schemas for tests is difficult so I added comment blocks describing their intent and what the objects contain

joneubank (approved):

PR #880, opened 2024-07-10
Add core functionality for remote connection resolvers

ciaranschutte:
~~- adds basic resolver functionality for remote connections (does not hook them up yet)~~
- add resolver functionality
- add resolver method to resolver map
- modify gql server to expose functionality
- basic healthcheck/ping  for remote connections (this can be expanded on if we want to introduce some "connection manager" functionality, holding the status in state and pinging from there)
- resolver functionality can be memoized or potentially broken down into sub resolvers because we likely don't want to ping all servers so often for a gql query that may not need it (can discuss offline once I get a concrete example)
- adds TS util type for const style maps
- tested connecting to two remote connections
  - both available
  -  1 available, 1 not
  - neither available


joneubank:
RE: memoization - I have hesitations around serving stale data, but if we do see performance issues then adding a caching layer could be helpful. I say we build this out fully and then do some performance testing to see if there are situations where caching would help.

modules/server/src/network/resolvers/connections/index.ts
```diff
@@ -0,0 +1,70 @@
+import { NetworkAggregationConfig, NetworkAggregationConfigInput } from '@/network/types';
+import { ObjectValues } from '@/utils/types';
+import axios from 'axios';
+import urljoin from 'url-join';
+
+/**
+ * Connection status types
+ */
```
joneubank, line 8: This description provides no more information than the name of the variable... Possibly not needed.

modules/server/src/network/resolvers/connections/index.ts
```diff
@@ -0,0 +1,70 @@
...
+		console.error(error);
+		return CONNECTION_STATUS.ERROR;
+	}
+};
+
+type RemoteConnectionData = NetworkAggregationConfigInput & {
+	availableAggregations: string[];
+	status: ConnectionStatus;
+};
+export const createRemoteConnectionResolvers = async (
```
joneubank, line 52: This exported function probably deserves a TSDoc block though ;)

modules/server/src/network/resolvers/connections/index.ts
```diff
@@ -0,0 +1,70 @@
...
+ * @param url - Remote connection url
+ * @returns A connection status
+ */
+const checkRemoteConnectionStatus = async (url: string): Promise<ConnectionStatus> => {
+	/**
+	 * recommended way to health check gql server is to run the `__typename` query that every server has
+	 * very small query with no additional params to, so using GET is not a concern for the GQL server
+	 * adds recommended pre flight header to make sure Apollo doesn't block request by CSRF protection
+	 */
+	const healthCheckQuery = '?query=%7B__typename%7D';
```
joneubank, line 28: Very helpful comment. Can you change the opening line comment symbol to be `/*` insted of `/**`. As written, intellisense software will use this comment block as the description of `healthCheckQuery` lol.

modules/server/src/network/resolvers/connections/index.ts
```diff
@@ -0,0 +1,70 @@
...
+		// only need to check the server returns successfully (just a health check)
+		if (pong.data.data?.__typename) {
+			return CONNECTION_STATUS.OK;
+		} else {
+			throw Error('no data object returned from GQL server __typname query');
+		}
+	} catch (error) {
+		console.error(error);
+		return CONNECTION_STATUS.ERROR;
+	}
```
joneubank, line 45: try-throw-catch is the correct handling here to ensure consistent response from both axios outcomes... but I still want to rant about it being the most convoluted flow control mechanism.

justincorrigible:
Agreed. Some may argue this counts as premature optimisation, which in our case as Jon correctly puts it, puts Arranger at risk of serving outdated data, or even worse, stuff that was meant to not exist anymore (deleted or unpublished).

modules/server/src/network/resolvers/connections/index.ts
```diff
@@ -0,0 +1,70 @@
...
+};
+
+type RemoteConnectionData = NetworkAggregationConfigInput & {
+	availableAggregations: string[];
+	status: ConnectionStatus;
+};
+export const createRemoteConnectionResolvers = async (
+	networkConfigs: NetworkAggregationConfig[],
+): Promise<RemoteConnectionData[]> => {
```
joneubank, line 54: Is this actually returning a resolver (function that graphql server uses to collect requested data)? Its actually just returning the `RemoteConnectionData`. It might be named wrong or returning the wrong type.

joneubank: Because this returns a static object, it is only resolving the `status` value once, when `createRemoteConnectionResolvers` is first called. This means the `status` field is not providing any real value... we already know the remote arranger node was available at startup time because we include the node in our config. If we're going to report `status` in our node type then we should be pinging it at the time the gql request is made.

joneubank (commented):
I think we should build out a bit more functionality to see how this connects with GQL to resolve requests before finalizing the review.

ciaranschutte:
> I think we should build out a bit more functionality to see how this connects with GQL to resolve requests before finalizing the review.

I'm trying to keep the PRs very small and contained for all the good reasons.
This _does not_ generate pass a function back to a resolver and hookup.
My first line on the description could be more description possibly?
"adds basic resolver functionality for remote connections (does not hook them up yet)"
Still merging into a feature branch 👍

ciaranschutte:
> I think we should build out a bit more functionality to see how this connects with GQL to resolve requests before finalizing the review.

Taking this into account I will add the following:
- returning async function that calls this created functionality 
- adding to resolver map to expose from gql server (for server)
- modification to schema to expose functionality correctly

modules/server/src/network/resolvers/connections/index.ts
ciaranschutte: my train of thought: I could see this being moved based on how we build out connection health checks - it's a nice fence of here's the connection types.

Just helps. More comment blocks. I don't think it confuses. So no negative impact. Bigger fish to fry and overall we want more docs not less 👍

ciaranschutte:
> I think we should build out a bit more functionality to see how this connects with GQL to resolve requests before finalizing the review.

Updated with PR feedback and expanded to included these modifications: https://github.com/overture-stack/arranger/pull/880#issuecomment-2227720080

This should fulfil the "connecting to GQL" request. Let me know if you'd like it expanded more.

modules/server/src/network/resolvers/index.ts
```diff
@@ -0,0 +1,8 @@
+import { NetworkAggregationConfig } from '../types';
+import { resolveRemoteConnectionNodes } from './remoteConnections';
+
+export const createResolvers = (networkConfigsWithSchemas: NetworkAggregationConfig[]) => ({
+	Query: {
+		nodes: async () => await resolveRemoteConnectionNodes(networkConfigsWithSchemas),
+	},
+});
```
joneubank, line 8: 🙌

modules/server/src/network/typeDefs/remoteConnections.ts
```diff
@@ -0,0 +1,16 @@
...
+  }
+
+	type RemoteConnectionNode {
+		url: String
+		name: String
+		description: String
+		documentName: String
+		availableAggregations: [String]
+		totalHits: Int
+		status: String
```
joneubank, line 13: isn't status an enum? fine to update this later.

joneubank (approved):
looks generally good.

modules/server/src/network/typeDefs/remoteConnections.ts
ciaranschutte: Just a TS type currently, not a GQL type. Will add enum for future use 👍

PR #881, opened 2024-07-16
Feat/agg resolver

ciaranschutte:
nb: merging into feature branch

- reworks typedefs for remote aggregations
- sets up resolver creation logic for all aggregations

does not cover: 
- logic to convert single to network agg, eg. NumericAggregations into NumericNetworkAggregations


modules/server/package.json
```diff
@@ -96,7 +96,10 @@
 		},
 		"transform": {
 			"^.+\\.[jt]sx?$": [
-				"ts-jest"
+				"ts-jest",
+				{
+					"diagnostics": false
+				}
```
ciaranschutte, line 102: stops type checking entire project while running tests

modules/server/src/network/resolvers/index.ts
```diff
@@ -1,8 +1,24 @@
...
+export const createResolvers = (
+	configs: NetworkAggregationConfig[],
+	networkFieldTypes: NetworkFieldType[],
+) => {
+	const aggregationsResolvers = createAggregationResolvers(configs, networkFieldTypes);
+
+	return {
+		Query: {
+			nodes: async () => await resolveRemoteConnectionNodes(configs),
+			aggregations: () => [], // TODO: Why need array returned? needs some truthy val?
```
ciaranschutte, line 20: unclear about this and currently investigating. We don't return anything for just an `aggregations` field, but it does need to return something truthy so sub fields resolve correctly

modules/server/src/network/resolvers/index.ts
```diff
@@ -1,8 +1,24 @@
...
+export const createResolvers = (
+	configs: NetworkAggregationConfig[],
+	networkFieldTypes: SupportedNetworkFieldType[],
+) => {
+	const aggregationsResolvers = createAggregationResolvers(configs, networkFieldTypes);
+
+	return {
+		Query: {
+			nodes: async () => await resolveRemoteConnectionNodes(configs),
+			aggregations: () => [], // TODO: Why do we need truthy val returned?
```
joneubank, line 20: Since we are handling the Aggregations type resolver at line 22, do we need this resolver at all? Or... why not just put our aggregationsResolvers object here instead?

modules/server/src/network/common.ts
```diff
@@ -0,0 +1,45 @@
+import { ObjectValues } from '@/utils/types';
+
+/*
+ * Network aggregations
+ */
+export const NETWORK_AGGREGATIONS = {
+	NetworkAggregations: 'NetworkAggregations',
+	NetworkNumericAggregations: 'NetworkNumericAggregations',
+} as const;
+export type NetworkAggregation = ObjectValues<typeof NETWORK_AGGREGATIONS>;
```
ciaranschutte, line 10: output types for resolved data

modules/server/src/network/resolvers/aggregations.ts
```diff
@@ -0,0 +1,120 @@
...
+			const gqlString = rootQueryFields.reduce((gqlString, fieldName) => {
+				const matchedAggregationField = findMatchedAggregationField(config, fieldName);
+				if (matchedAggregationField) {
+					const { name, type } = matchedAggregationField;
+					// get gql query string for supported aggregation
+					const aggregationFieldQueryString = supportedAggregationQueries.get(type);
+					return gqlString + `${name}${aggregationFieldQueryString}`;
+				}
+				return gqlString;
+			}, '');
```
ciaranschutte, line 104: build GQL query string for each remote connection based on how many fields in the original query it supports

modules/server/src/network/resolvers/aggregations.ts
```diff
@@ -1,101 +1,120 @@
...
+			 * convert string to AST object to use as query
+			 * not needed if gqlString is formatted correctly but this acts as a validity check
+			 */
+			try {
+				const gqlQuery = gql`{${gqlString}}`;
+				return { url: config.graphqlUrl, gqlQuery };
+			} catch (err) {
+				console.error(err);
+				return false;
+			}
```
joneubank, line 116: I'm not sure what to suggest here, but it seems weird that a whole remote node might just get no query sent because the gql threw an error. Shouldn't we have an object that indicates there is no gql query for that node and why?

modules/server/src/network/resolvers/aggregations.ts
```diff
@@ -1,101 +1,120 @@
...
+			const gqlString = rootQueryFields.reduce((gqlString, fieldName) => {
+				const matchedAggregationField = findMatchedAggregationField(config, fieldName);
+				if (matchedAggregationField) {
+					const { name, type } = matchedAggregationField;
+					// get gql query string for supported aggregation
+					const aggregationFieldQueryString = supportedAggregationQueries.get(type);
+					return gqlString + `${name}${aggregationFieldQueryString}`;
+				}
+				return gqlString;
+			}, '');
```
joneubank, line 104: Will we get an invalid gql query if there are no matching aggregationFields for that node? How is this empty query handled?

modules/server/src/network/resolvers/aggregations.ts
```diff
@@ -1,101 +1,120 @@
...
 };
 
-/**
- * TODO: Placeholder for resolving multiple remote connection responses into single aggregation
- */
-const resolveAggregation = (response: any, type: string) => {};
+export const queryConnections = async (queries: NetworkQuery[]) => {
+	const networkQueryPromises = queries.map(async (query) => {
+		const response = await queryConnection(query);
+		return { config, response };
```
joneubank, line 55: Where is `config` coming from? Why is it being added to each object in this array?

modules/server/src/network/resolvers/index.ts
```diff
@@ -1,24 +1,36 @@
...
+				args: {},
+				context: unknown,
+				info: GraphQLResolveInfo,
+			) => {
+				const networkQueries = createNetworkQueries(configs, info);
+				const networkResults = await queryConnections(networkQueries);
+				// TODO: implement aggregating
+				const resolvedResults = resolveAggregations(networkResults);
+				return resolvedResults;
+			},
```
joneubank, line 33: :+1:

joneubank (approved):

PR #884, opened 2024-08-15
Feat/resolver responses

ciaranschutte:
- adds caller resolver code to process network results into resolved aggregations
- returns resolved aggregations as result object
- moves common utils into common folder
- fixes schema

### DIFF
--- a/modules/server/src/network/aggregations/index.ts
+++ b/modules/server/src/network/aggregations/index.ts
@@ -1,22 +1,39 @@
 import { SupportedAggregation, SUPPORTED_AGGREGATIONS } from '../common';
 import { Aggregations, NetworkAggregation, NumericAggregations } from '../types';
 
-// TODO: implement
-export const resolveAggregations = (networkResults: void) => null;
+/**
+ * Pick each field from network result aggregations and reduce into single aggregation
+ *
+ * @param networkResults
+ * @param rootQueryFields
+ */
+export const resolveAggregations = (networkResults, rootQueryFields) => {
+	const resolvedNetworkAggregations = rootQueryFields.map((fieldName) => {
+		const fieldAggregations = networkResults.map((networkResult) => {
+			const documentName = Object.keys(networkResult)[0];
+			return networkResult[documentName][fieldName];
+		});
+		const aggregationType = fieldAggregations[0].__typename;
+		const resolvedAggregation = resolveToNetworkAggregation(aggregationType, fieldAggregations);
+		return { fieldName: fieldName, aggregation: resolvedAggregation };
+	});
+
+	return resolvedNetworkAggregations;
+};
 
 /**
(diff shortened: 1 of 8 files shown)


modules/server/src/network/aggregations/index.ts
```diff
@@ -1,22 +1,39 @@
...
 /**
  * Resolve aggregation based on aggregation type
  * @param type
  * @param aggregations
  */
-export const resolveToNetworkAggregation = <T>(
+export const resolveToNetworkAggregation = (
 	type: SupportedAggregation,
 	aggregations: Aggregations[],
 ) => {
```
joneubank, line 32: Wouldn't mind declaring a response type here if you know what should be returned, since we have multiple return statements this would validate that they are returning consistent types. Also helps me as I'm reading to know what the output of this function will be.

modules/server/src/network/index.ts
```diff
@@ -117,14 +122,7 @@ const fetchRemoteSchemas = async ({
...
 	const configs = networkQueries
-		.filter(
-			(
-				result,
-			): result is PromiseFulfilledResult<{
-				config: NetworkAggregationConfigInput;
-				gqlResponse: GQLTypeQueryResponse;
-			}> => result.status === 'fulfilled',
-		)
+		.filter(fulfilledPromiseFilter<NetworkQueryResult>)
```
joneubank, line 125: as of aTS 5.5 you shouldn't need a function with `input is Xyz`, the type predicate will be inferred by TS. Nothing wrong with this just FYI

modules/server/src/network/queries/index.ts
```diff
@@ -21,6 +22,13 @@ export type GQLTypeQueryResponse = {
 	};
 };
 
+// Data query response
+export type GQLResponse = {
```
joneubank, line 26: I think this name is too generic, there are lots of GQL apis and they don't all have a response in this form.

modules/server/src/network/aggregations/index.ts
```diff
@@ -1,22 +1,39 @@
...
+ * @param networkResults
+ * @param rootQueryFields
+ */
+export const resolveAggregations = (networkResults, rootQueryFields) => {
+	const resolvedNetworkAggregations = rootQueryFields.map((fieldName) => {
+		const fieldAggregations = networkResults.map((networkResult) => {
+			const documentName = Object.keys(networkResult)[0];
+			return networkResult[documentName][fieldName];
+		});
+		const aggregationType = fieldAggregations[0].__typename;
```
demariadaniel, line 16: Any potential for empty arrays / need for falsy handling ?

ciaranschutte: at this point falseys should be filtered out and every apollo gql server returns `__typename`

PR #885, opened 2024-09-12
Fix network agg resolution

ciaranschutte:
Fixes bucket resolution 

example: 
Server1 buckets: [Cat, Dog]
Server2 buckets: [Cat, Bird]
Output: [Cat, Dog, Bird]

### DIFF
--- a/modules/server/src/network/aggregations/index.ts
+++ b/modules/server/src/network/aggregations/index.ts
@@ -44,7 +44,7 @@ export const resolveToNetworkAggregation = (
  * Takes an array of the same aggregation type and computes the singular type
  * eg. NumericAggregation => NetworkNumericAggregation
  *
- * Note for operations on Buckets -  size can be large
+ * Note for operations on Buckets -  the size of the array can be large (e.g. total bucket count), complicating lookups, etc.
  *
  * @param aggregations
  * @returns
--- a/modules/server/src/network/aggregations/index.ts
+++ b/modules/server/src/network/aggregations/index.ts
@@ -44,6 +44,8 @@ export const resolveToNetworkAggregation = (
  * Takes an array of the same aggregation type and computes the singular type
  * eg. NumericAggregation => NetworkNumericAggregation
  *
+ * Note for operations on Buckets -  size can be large
+ *
  * @param aggregations
  * @returns
  */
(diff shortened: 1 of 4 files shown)


modules/server/src/network/typeDefs/networkAggregations.ts
```diff
@@ -1,13 +1,28 @@
...
 // TODO: Placeholder to expand type
 const networkAggregations = new GraphQLObjectType({
 	name: 'NetworkAggregations',
 	fields: {
 		bucket_count: {
 			type: GraphQLInt,
 		},
+		buckets: {
+			type: new GraphQLList(bucket),
+		},
```
ciaranschutte, line 25: extend GQL typings for GQL query response

modules/server/src/network/typeDefs/aggregations.ts
```diff
@@ -2,11 +2,26 @@ import { GraphQLObjectType, GraphQLSchema } from 'graphql';
...
+ */
+const convertToGQLObjectType = (networkFieldTypes) => {
+	return networkFieldTypes.reduce((allFields, currentField) => {
+		const field = {
+			[currentField.name]: { type: singleToNetworkAggregationMap.get(currentField.type) },
+		};
+		return { ...allFields, ...field };
+	}, {});
+};
+
```
ciaranschutte, line 19: isolate function

modules/server/src/network/aggregations/index.ts
```diff
@@ -44,21 +44,46 @@ export const resolveToNetworkAggregation = (
...
+				if (existingBucket) {
+					// update existing bucket
+					computedBuckets[existingBucketIndex] = {
+						...existingBucket,
+						doc_count: existingBucket.doc_count + doc_count,
+					};
+				}
+			} else {
+				computedBuckets.push(bucket);
+			}
```
ciaranschutte, line 82: Size of buckets can be large
Using `forEach` to either update "found bucket" or add to resolved buckets

Avoids using functional operations on the entire "resolved buckets" and then having a large array of Buckets and needing to flatten and filter on same Bucket `key`

No way around searching buckets for existing bucket unfortunately as the data structure is an array and a key lookup is not an option.

ciaranschutte:
CI failing is known issue

modules/server/src/network/aggregations/index.ts
justincorrigible: the lookup has always been a pain point... perhaps we should make a ticket for future improvement to make the buckets a dictionary available internally in the Arranger server for this and other lookups, even if it's ultimately turned in the UI as an array for display purposes.

modules/server/src/network/aggregations/index.ts
```diff
@@ -44,21 +44,46 @@ export const resolveToNetworkAggregation = (
...
+				const existingBucket = computedBuckets[existingBucketIndex];
+				if (existingBucket) {
+					// update existing bucket
+					computedBuckets[existingBucketIndex] = {
+						...existingBucket,
+						doc_count: existingBucket.doc_count + doc_count,
+					};
+				}
+			} else {
+				computedBuckets.push(bucket);
```
justincorrigible, line 81: lol I don't think I'll ever be happy/comfortable with mutational logic.

modules/server/src/network/aggregations/index.ts
```diff
@@ -44,21 +44,46 @@ export const resolveToNetworkAggregation = (
  * Takes an array of the same aggregation type and computes the singular type
  * eg. NumericAggregation => NetworkNumericAggregation
  *
+ * Note for operations on Buckets -  size can be large
```
justincorrigible, line 47: "size" alone is ambiguous, in that a large bucket doesn't add complexity, whereas the count of buckets is what I think you're trying to highlight here
```suggestion
 * Note for operations on Buckets -  the size of the array can be large (e.g. total bucket count), complicating lookups, etc.
```

modules/server/src/network/aggregations/index.ts
ciaranschutte: https://github.com/overture-stack/arranger/issues/886

PR #887, opened 2024-09-16
Response data & batching requests

ciaranschutte:
covers feedback from last demo
- batches resolving of aggregations as network node queries return instead of after all queries return
- changes gql response shape (including typedefs and resolver config changes)
- uses Jon's well tested success/failure code for network handling

There are some loose typings (any) to be addressed in future PR

aggregationPipeline mutates objects in order to batch results, comments clearly denote where this happens
open to suggestions for pure functional equivalent

### DIFF
--- a/modules/server/src/network/resolvers/aggregations.ts
+++ b/modules/server/src/network/resolvers/aggregations.ts
@@ -61,6 +61,8 @@ export const aggregationPipeline = async (
 		return { ...accumulator, [field]: emptyAggregation };
 	}, {});
 
+	const nodeInfo: any = [];
+
 	const aggregationResultPromises = queries.map<
 		Promise<{
 			aggregations: any;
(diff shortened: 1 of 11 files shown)


modules/server/src/network/resolvers/index.ts
```diff
@@ -12,29 +11,33 @@ type NetworkSearchRoot = {
...
+				const { requestedAggregations } = getRequestedFields(info);
+				const networkQueries = createNetworkQueries(configs, requestedAggregations);
+
+				// Query remote connections and aggregate results
+				const { aggregationResults, nodeInfo } = await aggregationPipeline(
+					networkQueries,
+					requestedAggregations,
+				);
+				const response = createResponse({ aggregationResults, nodeInfo });
+				console.log('resp', response);
```
demariadaniel, line 40: This looks like test logging

ciaranschutte: thank you - updated - https://github.com/overture-stack/arranger/pull/887/commits/9e3f631e7cdab75e46c020e439b1a29284a7aeaf

demariadaniel (approved):
Looking good!

modules/server/src/network/resolvers/aggregations.ts
```diff
@@ -80,30 +118,39 @@ const findMatchedAggregationField = (config: NetworkAggregationConfig, fieldName
...
+ *
+ * @param requestedAggregations
+ */
+const createGqlFieldsString = (config: NetworkAggregationConfig, requestedAggregations: any[]) => {
+	return requestedAggregations.reduce((gqlString, fieldName) => {
+		const matchedAggregationField = findMatchedAggregationField(config, fieldName);
+		if (matchedAggregationField) {
+			const { name, type } = matchedAggregationField;
+			// get gql query string for supported aggregation
+			// TODO: only query requested fields + bucket_count if nodes is requested
```
demariadaniel, line 131: Discussed offline, TODO's are addressed in next PR

PR #890, opened 2024-09-19
Connectionn cleanup

ciaranschutte:
Changes how gql query string works: https://github.com/overture-stack/arranger/pull/890#discussion_r1768011267

The rest is reorg - renaming of types, variables and organization of files


modules/server/src/network/util.ts
```diff
@@ -31,12 +21,21 @@ export const fulfilledPromiseFilter = <Result>(result: unknown): result is Resul
...
+	// ensure __typename will be queried to network nodes
+	const aggs = Object.keys(aggregations).reduce((aggs, key) => {
+		const element = aggregations[key];
+		if (!element.hasOwnProperty('__typename')) {
+			element['__typename'] = {};
+		}
+		return { ...aggs, [key]: element };
+	}, {});
+	return aggs;
 };
```
ciaranschutte, line 41: parses out `aggregation` field with requested properties to be sent to network nodes
no need to re-send full query to network nodes, in fact, they'll just cause errors
ensures `__typename` is added for ease of use in resolving responses. easier than parsing query ASTs

modules/server/src/network/queries/index.ts
```diff
@@ -46,22 +45,3 @@ export const gqlAggregationTypeQuery = `#graphql
...
-		}
-	}
-`;
-
-/**
- * Returns gql query for gql type
- */
-export const supportedAggregationQueries = new Map<SupportedAggregation, string>();
-supportedAggregationQueries.set(SUPPORTED_AGGREGATIONS.Aggregations, aggregationsQuery);
-supportedAggregationQueries.set(SUPPORTED_AGGREGATIONS.NumericAggregations, '');
```
ciaranschutte, line 67: forwarding relevant part of query to network nodes, instead of splicing in this static query for each field
resulting queries are smaller and only return requested fields (half the point of GQL)

modules/server/src/network/resolvers/aggregations.ts
```diff
@@ -111,19 +115,46 @@ type NetworkQuery = {
...
+		.replaceAll('"', '')
+		.replaceAll(':', '')
+		.replaceAll('{}', '')
+		.replaceAll(',', ' ');
+
+	// add top level field for query and format with correct brackets
+	const gqlString = `{${documentName} ${gqlFieldsString}}`;
+
+	return gqlString;
 };
```
ciaranschutte, line 155: parse out request object to gql string

PR #891, opened 2024-09-20
Reorg connection

ciaranschutte:
Merging into feature/federated_search

Changes how gql query string works: https://github.com/overture-stack/arranger/pull/891#discussion_r1768778059

The rest is reorg - renaming of types, variables and organization of files


modules/server/src/network/util.ts
```diff
@@ -31,12 +21,21 @@ export const fulfilledPromiseFilter = <Result>(result: unknown): result is Resul
...
+	// ensure __typename will be queried to network nodes
+	const aggs = Object.keys(aggregations).reduce((aggs, key) => {
+		const element = aggregations[key];
+		if (!element.hasOwnProperty('__typename')) {
+			element['__typename'] = {};
+		}
+		return { ...aggs, [key]: element };
+	}, {});
+	return aggs;
 };
```
ciaranschutte, line 41: parses out aggregation field with requested properties to be sent to network nodes
no need to re-send full query to network nodes, in fact, they'll just cause errors
ensures __typename is added for ease of use in resolving responses. easier than parsing query ASTs

modules/server/src/network/setup/fields.ts
```diff
@@ -0,0 +1,74 @@
...
+
+const isSupportedType = (
+	fieldObject: NetworkFieldType<string>,
+	supportedList: string[],
+): fieldObject is SupportedNetworkFieldType => {
+	return supportedList.includes(fieldObject.type);
+};
+
+/**
+ * Parse network fields into supporter and unsupported
```
demariadaniel, line 20: 'supported'

demariadaniel (approved):

demariadaniel:
Discussed a few details over Slack but overall I see improvements and logic looks good

PR #893, opened 2024-09-24
Accumulator pipeline

ciaranschutte:
cleans up aggregation pipeline code:
- iterates over configs instead of requests, reducing # of iterations to # of configs
- moves messy accumulator into a class 
- handles "Server not available, connection refused" state reporting node with correct error status


modules/server/src/network/resolvers/aggregations.ts
```diff
@@ -148,36 +91,63 @@ const createGqlFieldsString = (requestedFields: {}, documentName: string) => {
...
-					${gqlString}
-				`;
-				return { url: config.graphqlUrl, gqlQuery };
-			} catch (err) {
-				console.error('invalid gql', err);
-				return false;
-			}
-		})
-		.filter(Boolean);
+	const totalAgg = new AggregationAccumulator(requestedAggregationFields);
```
ciaranschutte, line 124: uses a class to hold accumulator state

modules/server/src/network/resolvers/aggregations.ts
```diff
@@ -33,77 +35,18 @@ const fetchData = async (query: NetworkQuery) => {
...
-			console.error(`Network failure: ${url}`);
-			return failure(CONNECTION_STATUS.ERROR, `Network failure: ${url}`);
-		} else {
-			return failure(CONNECTION_STATUS.ERROR, `Unknown error`);
+		if (axios.isAxiosError(error)) {
+			if (error.code === 'ECONNREFUSED') {
+				console.error(`Network failure: ${url}`);
+				return failure(CONNECTION_STATUS.ERROR, `Network failure: ${url}`);
+			}
 		}
```
ciaranschutte, line 43: catches error is server is unavailable
nested for now - would like to check if there are other codes need to be covered

modules/server/src/network/aggregations/index.ts
```diff
@@ -172,3 +168,31 @@ const resolveNumericAggregation = (aggregations: NumericAggregations) => {
...
+			accumulator: this.totalAgg,
+			networkResult: data,
+			requestedAggregationFields: fields,
+		});
+	}
+
+	result() {
+		return this.totalAgg;
+	}
+}
```
ciaranschutte, line 198: isolates all the accumulator logic from this module cleanly into an exposed class

modules/server/src/network/aggregations/index.ts
```diff
@@ -172,3 +168,31 @@ const resolveNumericAggregation = (aggregations: NumericAggregations) => {
...
+			return { ...accumulator, [field]: emptyAggregation };
+		}, {});
+	}
+
+	resolve(data, fields) {
+		resolveAggregations({
+			accumulator: this.totalAgg,
+			networkResult: data,
+			requestedAggregationFields: fields,
+		});
```
joneubank, line 192: Why wrap another function here... could just move the contents of resolveAggregation to be defined inside this class...

Also, I can't approve a function without argument type declarations. Right now you could do `.resolve(true, false)` and TS will be happy with you.

modules/server/src/network/httpResponses.ts
```diff
@@ -26,9 +26,9 @@ export type Result<T, FailureStatus extends string, FailureData = void> =
  * @returns {boolean} Whether the Result was a Success or not
  */
 export function isSuccess<T, FailureStatus extends string, FailureData>(
-	result: Result<T, FailureStatus, FailureData>,
+	result?: Result<T, FailureStatus, FailureData>,
 ): result is Success<T> {
-	return result.status === 'SUCCESS';
+	return result !== undefined && result.status === 'SUCCESS';
```
joneubank, line 31: We should probably have 'SUCCESS' as a constant somewhere.

modules/server/src/network/aggregations/index.ts
```diff
@@ -172,3 +168,31 @@ const resolveNumericAggregation = (aggregations: NumericAggregations) => {
 	// TODO: implement
 	throw Error('Not implemented');
 };
+
+const emptyAggregation: NetworkAggregation = { bucket_count: 0, buckets: [] };
+
+export class AggregationAccumulator {
```
joneubank, line 174: May want to rename this file after this class. `AggregationAccumulator.ts`

modules/server/src/network/httpResponses.ts
joneubank: After reviewing where this was called... We really shouldnt have the situation where we have `Result<> | undefined`, that goes against the entire purpose of returning that type. We probably should revert this change.

joneubank (changes requested):
Clean up the types before you commit to this approach.

modules/server/src/network/aggregations/index.ts
ciaranschutte: `resolveAggregation` was created already and it works nice, if moved into this class it has to be tied to the class and not an isolated function because it references class properties.

ciaranschutte:
@joneubank fix types for AggregationAcc

modules/server/src/network/aggregations/index.ts
ciaranschutte: tied to module scope already so I'm worried about visibility outside the class

modules/server/src/network/aggregations/AggregationAccumulator.ts
```diff
@@ -172,3 +168,34 @@ const resolveNumericAggregation = (aggregations: NumericAggregations) => {
...
+	totalAgg: RemoteAggregation;
+
+	constructor(requestedFieldsMap: RequestedFieldsMap) {
+		const requestedFields = Object.keys(requestedFieldsMap);
+		this.requestedFields = requestedFields;
+		/*
+		 * seed accumulator with the requested field keys
+		 * this will make it easier to add to using key lookup instead of Array.find
+		 */
+		this.totalAgg = requestedFields.reduce<RemoteAggregation>((accumulator: any, field: any) => {
```
demariadaniel, line 185: Wouldn't `field: string` work here?

demariadaniel (approved):
One question about use of 'any' but I see other PR feedback's been updated. LGTM 🚀

modules/server/src/network/aggregations/AggregationAccumulator.ts
ciaranschutte: Added a generic, just forgot to remove the inline `any`.
The generic types these fields correctly once `any` is removed.

PR #895, opened 2024-09-29
Feat/fed total hits

ciaranschutte:
Major:
- reworks resolution and graphql server schema to return the the number of hits from connection nodes
(please see screenshot with dummy data)

Minor:
- adds basic flag to enable or disable bassed on ENV
- extends timeout for unresponsive query
- type reworking / additions / better naming so more cases are covered, targeted code complete, ready for review, no implicit `any`

Screenshot 2024-10-01 at 3 59 01 PM


modules/server/src/graphqlRoutes.js
```diff
@@ -281,9 +281,19 @@ export const createSchemasFromConfigs = async ({
...
+					...config,
+					/**
+					 * part of the gql schema is generated dynamically
+					 * in the case of the "file" field, the field name and type name are the same
+					 * it's more flexible to define it here as an additional property than to confuse functions further down the pipeline
+					 */
+					documentName: config.documentType,
+				})),
+			});
+		}
```
ciaranschutte, line 296: as comment says, `documentName` and `documentType` end up being the same mostly based on how Arranger dynamically builds the query. It makes more sense to add this early in the configuration than to conditional flip back and forth further down the stack. Other functions keep their distinction on which field they're looking at, and it'll be trivial to have the `documentName` configurable if Arranger dynamic schema generation changes

modules/server/src/network/types/hits.ts
```diff
@@ -0,0 +1 @@
+export type Hits = { total: number };
```
ciaranschutte, line 1: there is not a TS definition for these so adding as I go

demariadaniel: Will this file have more types added? Wondering the need for its own file if it's only a one line definition

modules/server/src/network/setup/query.ts
```diff
@@ -54,22 +62,26 @@ const fetchRemoteSchema = async (
...
-		return { name: config.displayName, fields };
-	});
+	const r = networkQueries
+		.filter(fulfilledPromiseFilter<NetworkQueryResult>)
+		.map((networkResult) => {
+			const { config, gqlResponse } = networkResult.value;
+			const fields = gqlResponse.__type.fields;
+			return { name: config.displayName, fields };
+		});
+	console.log('r', r);
```
demariadaniel, line 85: Test logging?

modules/server/src/network/types/hits.ts
ciaranschutte: Very likely and definitely ideally. It related to a core Arranger schema type which has many more properties than this.
As long as we get to it , it should be the seperate entity types as types which compose into this parent `Hits` type

modules/server/src/network/setup/query.ts
```diff
@@ -22,15 +22,23 @@ type NetworkQueryResult = PromiseFulfilledResult<{
...
+	 * eg. `aggregations: fileAggregations`, `hits: fileConnection`
+	 */
 	try {
+		// targeting just the "aggregations" field using type name
+		const typename = `${documentType}Aggregations`;
 		const response = await fetchGql({
 			url: graphqlUrl,
 			gqlQuery: gqlAggregationTypeQuery,
-			variables: { documentName: documentType },
+			variables: { documentTypeName: typename },
```
demariadaniel, line 41: Camel case `typeName` ?

modules/server/src/graphqlRoutes.js
```diff
@@ -281,9 +281,19 @@ export const createSchemasFromConfigs = async ({
...
+		if (ENABLE_NETWORK_AGGREGATION) {
+			const { networkSchema } = await createSchemaFromNetworkConfig({
+				networkConfigs: configsFromFiles[ConfigProperties.NETWORK_AGGREGATION].map((config) => ({
+					...config,
+					/**
+					 * part of the gql schema is generated dynamically
+					 * in the case of the "file" field, the field name and type name are the same
+					 * it's more flexible to define it here as an additional property than to confuse functions further down the pipeline
+					 */
+					documentName: config.documentType,
```
demariadaniel, line 293: I'd be inclined to define networkConfigs as a variable first

```const networkConfigs = configsFromFiles[ConfigProperties.NETWORK_AGGREGATION].map((config) => ({
					...config,
					/**
					 * part of the gql schema is generated dynamically
					 * in the case of the "file" field, the field name and type name are the same
					 * it's more flexible to define it here as an additional property than to confuse functions further down the pipeline
					 */
					documentName: config.documentType}));
const { networkSchema } = await createSchemaFromNetworkConfig({ networkConfigs });```

**Comment by demariadaniel** [comment]
Few stylistic suggestions but otherwise this LGTM 🛸

File: modules/server/src/network/setup/query.ts (hunk 4 above)
Comment (ciaranschutte) [reply modules/server/src/network/setup/query.ts:41]: consistent with usage in relation to gql schema. although this is a function not directly in the resolvers, it makes sense to keep it consistent case wise as it is referring to the same entity eg. typename is referring to __typename not __typeName

File: modules/server/src/graphqlRoutes.js (hunk 5 above)
Comment (ciaranschutte) [reply modules/server/src/graphqlRoutes.js:293]: interesting. for what reason?
I'm mostly trying to avoid doing:
```
const networkConfigs = configsFromFile......
const adjustedNetworkConfigs = networkConfigs....
``

demariadaniel (approved):

modules/server/src/network/setup/query.ts
demariadaniel: Got it 👍

modules/server/src/graphqlRoutes.js
demariadaniel: Mainly readability, and generally to reduce logic in arguments.

If `createSchemaFromNetworkConfig` arguments ever change it becomes a lot cleaner passing in variable object keys.

If there's ever an error it's easy to trace back to the original definition.

PR #897, opened 2024-10-02
feat/agg acc tests

ciaranschutte:
- aggregation accumulator test added
- cleanup some naming & types


modules/server/src/network/aggregations/AggregationAccumulator.ts
```diff
@@ -43,14 +44,14 @@ const resolveAggregations = ({
  * @param type
  * @param aggregations
  */
-export const resolveToNetworkAggregation = (
-	type: SupportedAggregation,
-	aggregations: Aggregations[],
-): AllAggregations | undefined => {
+const resolveToNetworkAggregation = (
+	type: string,
```
demariadaniel, line 48: This is no longer typed?

modules/server/src/network/aggregations/AggregationAccumulator.ts
```diff
@@ -1,10 +1,10 @@
-import { SupportedAggregation, SUPPORTED_AGGREGATIONS } from '../common';
+import { SUPPORTED_AGGREGATIONS } from '../common';
 import { Aggregations, Bucket, NumericAggregations } from '../types/aggregations';
 import { AllAggregations } from '../types/types';
 import { RequestedFieldsMap } from '../util';
 
 type ResolveAggregationInput = {
-	networkResult: AllAggregations;
+	aggregationsMap: AllAggregations;
```
demariadaniel, line 7: I like this name update

modules/server/src/network/aggregations/tests/aggregation.test.ts
```diff
@@ -1,27 +1,42 @@
-import { resolveAggregation } from '..';
-import { aggregation } from './fixture';
+import { AggregationAccumulator } from '../AggregationAccumulator';
+import { aggregation as fixture } from './fixture';
```
demariadaniel, line 2: We're being gentle today but there are a handful of 'as' which could be improved

ciaranschutte: That's ok. Let's get them right. Can you please point out where we should improve them as I don't see any in this file?

ciaranschutte: Oh - this is an import syntax - not TS assertion

demariadaniel: Sorry my bad, this is just aliasing

modules/server/src/network/aggregations/AggregationAccumulator.ts
```diff
@@ -43,14 +44,14 @@ const resolveAggregations = ({
...
+const resolveToNetworkAggregation = (
+	type: string,
+	aggregations: Aggregations[] | NumericAggregations[],
+): Aggregations | NumericAggregations => {
 	if (type === SUPPORTED_AGGREGATIONS.Aggregations) {
-		return resolveAggregation(aggregations);
+		return resolveAggregation(aggregations as Aggregations[]);
 	} else if (type === SUPPORTED_AGGREGATIONS.NumericAggregations) {
-		return resolveNumericAggregation(aggregations);
+		return resolveNumericAggregation(aggregations as NumericAggregations);
```
demariadaniel, line 54: Wondering how we could make better type checks here rather than using 'as'

ciaranschutte: I think there's likely a more optimal solution for the TS.

I feel comfortable about this usage of `as` because if you look at the conditional, at runtime we are checking that it is of a type, it's just not informing the typechecker correctly.

modules/server/src/network/aggregations/AggregationAccumulator.ts
demariadaniel: Sorry not specific enough, no longer named type `SupportedAggregation` or an enum, just wildcard string?

PR #899, opened 2024-10-04
Individual node config query fields

ciaranschutte:
- get available aggregations per node on api fetch response
- only send nodes queries with their individually available aggregations
- returns available aggregations with resolver info

### DIFF
--- a/modules/server/src/network/resolvers/aggregations.ts
+++ b/modules/server/src/network/resolvers/aggregations.ts
@@ -6,7 +6,7 @@ import { fetchGql } from '../gql';
 import { failure, isSuccess, Result, Success, success } from '../httpResponses';
 import { Hits } from '../types/hits';
 import { NetworkConfig } from '../types/setup';
-import { AllAggregations } from '../types/types';
+import { AllAggregations, NodeConfig } from '../types/types';
 import { ASTtoString, RequestedFieldsMap } from '../util';
 import { CONNECTION_STATUS, NetworkNode } from './networkNode';
 
@@ -58,7 +58,6 @@ type NetworkQuery = {
 
 /**
  * Converts info field object into string
- *
  * @param requestedFields
  *
  * @example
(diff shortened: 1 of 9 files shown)


modules/server/src/network/types/types.ts
```diff
@@ -33,3 +34,5 @@ export type NetworkAggregation = {
 	bucket_count: number;
 	buckets: Bucket[];
 };
+
+export type NodeConfig = NetworkConfig & { aggregations: { name: string; type: string }[] };
```
ciaranschutte, line 38: `NetworkConfig` is input to the setup, `NodeConfig` is the output when relevant data is retrieved.
Distinct types because added `aggregations` has an optional type doesn't make sense - we want to make sure it's present

modules/server/src/network/typeDefs/aggregations.ts
```diff
@@ -39,13 +39,22 @@ export const createNetworkAggregationTypeDefs = (
...
+	});
+
 	const remoteConnectionType = new GraphQLObjectType({
 		name: 'RemoteConnection',
 		fields: {
 			name: { type: GraphQLString },
 			hits: { type: GraphQLInt },
 			status: { type: GraphQLString },
 			errors: { type: GraphQLString },
+			aggregations: { type: new GraphQLList(aggList) },
```
ciaranschutte, line 57: return list of available aggregations, insightful about node

modules/server/src/network/resolvers/aggregations.ts
```diff
@@ -102,11 +101,24 @@ export const createNodeQueryString = (
...
+	// ensure requested field is available to query on node
+	const fieldsToRequest = Object.keys(requestedFields).reduce((acc, requestedFieldKey) => {
+		const field = requestedFields[requestedFieldKey];
+		if (availableFields.some((field) => field.name === requestedFieldKey)) {
+			return { ...acc, [requestedFieldKey]: field };
+		} else {
+			return acc;
+		}
+	}, {});
+
```
ciaranschutte, line 120: only request fields that available on a node

modules/server/src/network/setup/query.ts
```diff
@@ -76,15 +76,16 @@ export const fetchAllNodeAggregations = async ({
...
-	const nodeAggregations = networkQueries
+	const nodeConfigs = networkQueries
 		.filter(fulfilledPromiseFilter<NetworkQueryResult>)
 		.map((networkResult) => {
 			const { config, gqlResponse } = networkResult.value;
 			const fields = gqlResponse.__type.fields;
-			return { name: config.displayName, fields };
+			const aggregations = fields.map(normalizeGqlField);
+			return { ...config, aggregations };
 		});
```
ciaranschutte, line 86: Move normalizing fields up to api fetch call, and pass it down as one config

ciaranschutte: This avoids messy "zipping" of config objects down to resolvers

PR #906, opened 2024-10-24
Fed search cleanup typings

ciaranschutte:
Fix type errors


modules/server/src/network/resolvers/aggregations.ts
```diff
@@ -61,9 +63,9 @@ const fetchData = async <SuccessType>(
 			}
 		}
 		return failure(CONNECTION_STATUS.ERROR, `Unknown error`);
-	} finally {
-		console.log(`Fetch data completing for ${query.url}`);
 	}
+	// TS would like a return value outside of try/catch handling
+	return failure(CONNECTION_STATUS.ERROR, `Unknown error`);
```
ciaranschutte, line 68: even though results are handled in the `try/catch` block, TS doesn't like the implicit undefined from the function.
It is a tsconfig setting, but I think it's best to return something rather than tweak the project setting.

TS also doesn't like nothing returned in a `finally` block, and returning something here would override the returned values in the `catch` block

modules/server/src/network/aggregations/AggregationAccumulator.ts
```diff
@@ -13,18 +13,28 @@ type ResolveAggregationInput = {
...
-const addToAccumulator = ({ existingAggregation, aggregation, type }) => {
+const addToAccumulator = <T extends Aggregations | NumericAggregations>({
+	existingAggregation,
+	aggregation,
+	type,
+}: {
+	existingAggregation: T | undefined;
+	aggregation: T;
+	type: SupportedAggregation;
+}) => {
```
ciaranschutte, line 33: need `existingAggregation` and `aggregation` to be the same type if aggregating

modules/server/src/network/aggregations/AggregationAccumulator.ts
```diff
@@ -37,13 +47,19 @@ const resolveAggregations = ({ data, accumulator, requestedFields }: ResolveAggr
...
+		const type = ALL_NETWORK_AGGREGATION_TYPES_MAP.get(requestedField) as SupportedAggregation;
 		const existingAggregation = accumulator[requestedField];
 
 		if (isFieldAvailable) {
 			accumulator[requestedField] = addToAccumulator({
 				existingAggregation,
-				aggregation: aggregations[requestedField],
+				// similarly due to GQL schema validation this will not be undefined
+				aggregation: aggregations[requestedField]!,
 				type,
```
ciaranschutte, line 63: neither `isFieldAvailable` variable or directly putting a conditional of `if(requestedField in aggregations)` works to assert not undefined

modules/server/src/network/tests/helpers.test.ts
```diff
@@ -1,26 +1,21 @@
-import { getFieldTypes } from '..';
+import { SupportedAggregation } from '../common';
+import { getFieldTypes } from '../setup/fields';
 
 describe('helpers', () => {
 	test('getField returns both supported and unsupported types', () => {
-		const supportedAggregations = ['Aggregations'];
+		const supportedAggregations = ['Aggregations'] as SupportedAggregation[];
```
ciaranschutte, line 6: test file

modules/server/src/network/resolvers/aggregations.ts
```diff
@@ -161,6 +163,7 @@ export const createNetworkQuery = (
 		return gqlQuery;
 	} catch (err) {
 		console.error('invalid gql', err);
+		return undefined;
```
ciaranschutte, line 166: TS doesn't like implicit undefined

modules/server/src/network/aggregations/AggregationAccumulator.ts
```diff
@@ -13,18 +13,28 @@ type ResolveAggregationInput = {
 
 type AggregationsTuple = [Aggregations, Aggregations];
 type NumericAggregationsTuple = [NumericAggregations, NumericAggregations];
+type GenericTuple<T> = [T, T];
```
joneubank, line 16: Declaring a type alias for this, when it is only used once, is a bit overkill. Can just use `[T, T]` in place.

If this pattern is used in the future, a better name may be `TwoTuple` since tuples have arbitrary length.

modules/server/src/network/aggregations/AggregationAccumulator.ts
```diff
@@ -37,13 +47,19 @@ const resolveAggregations = ({ data, accumulator, requestedFields }: ResolveAggr
 		const { aggregations, hits } = data;
 
 		const isFieldAvailable = !!aggregations[requestedField];
-		const type = ALL_NETWORK_AGGREGATION_TYPES_MAP.get(requestedField);
+
+		/*
+		 * requested field will always be in ALL_NETWORK_AGGREGATION_TYPES_MAP
+		 * GQL schema validation will throw an error earlier if a requested field isn't in the schema
+		 */
+		const type = ALL_NETWORK_AGGREGATION_TYPES_MAP.get(requestedField) as SupportedAggregation;
```
joneubank, line 55: This type assertion is likely always true in your current code logic. However, if you only consider the context of this function, there is no gurantee that the `requestedField` is a known key of the `ALL_NETWORK_AGGREGATION_TYPES_MAP`. Therefore, the returned type from map.get is correct... this could be `undefined`. We should handle that undefined case, even if all it does is skip this one `requestedField` in the forEach. Likely you want to include a log statement about why the function could not complete successfully and add some details to the comment block at the top of the function explaining what is required fro this function to be used (a comment like `ALL_NETWORK_AGGREGATION_TYPES_MAP` must be initialized before this is called).

modules/server/src/network/aggregations/AggregationAccumulator.ts
```diff
@@ -37,13 +47,19 @@ const resolveAggregations = ({ data, accumulator, requestedFields }: ResolveAggr
...
 		const existingAggregation = accumulator[requestedField];
 
 		if (isFieldAvailable) {
 			accumulator[requestedField] = addToAccumulator({
 				existingAggregation,
-				aggregation: aggregations[requestedField],
+				// similarly due to GQL schema validation this will not be undefined
+				aggregation: aggregations[requestedField]!,
```
joneubank, line 62: To remove this non-null assertion we can store `aggregations[requestedField]` in a variable and test if that value is defined, vs using the `isFieldAvailable` bool:

```ts
const aggregation = aggregations[requestedField];
		if (aggregation !== undefined) {
			accumulator[requestedField] = addToAccumulator({
				existingAggregation,
				// similarly due to GQL schema validation this will not be undefined
				aggregation,
				type,
			});
		} else {
```

modules/server/src/network/resolvers/aggregations.ts
```diff
@@ -134,7 +136,7 @@ export const createNodeQueryString = (
 export const createNetworkQuery = (
 	config: NodeConfig,
 	requestedFields: RequestedFieldsMap,
-): DocumentNode => {
+): DocumentNode | undefined => {
```
joneubank, line 139: Since this may return undefined, and this funciton is exported for use by other modules, it would be good to include a description in the TSDoc about when this will return `undefined` so other devs understand what this response is communicating.

modules/server/src/network/tests/helpers.test.ts
joneubank: Would prefer: `const supportedAggregations: SupportedAggregation[] = ['Aggregations'];`

joneubank (changes requested):
Couple type assertions that can be quickly refactored to remove.

joneubank (approved):

PR #908, opened 2024-10-30
add agg mode env config

ciaranschutte:
- pass `enabledAggregationMode` flag through system using same style as `enableAdmin`
- if aggregation mode is enabled, strips out `hits` from graphql schema (typedefs and resolvers)


modules/server/.env.schema
```diff
@@ -14,4 +14,5 @@ MAX_LIVE_VERSIONS=3
 MAX_RESULTS_WINDOW=10000
 PING_MS=2200
 PING_PATH=/ping
-PORT=5050
\ No newline at end of file
+PORT=5050
+ENABLE_AGGREGATION_MODE=false
```
justincorrigible, line 18: lol "for consistency" 
(i.e. not sure why it was missing to begin with 🤔)
```suggestion
ENABLE_AGGREGATION_MODE=false

```

modules/server/src/app.js
```diff
@@ -10,7 +10,10 @@ app.use(cors());
 export default async function (rootPath = '') {
 	global.__basedir = rootPath;
 
-	return Arranger({ enableAdmin: ENV_CONFIG.ENABLE_ADMIN }).then((router) => {
+	return Arranger({
+		enableAdmin: ENV_CONFIG.ENABLE_ADMIN,
+		enableAggregationMode: ENV_CONFIG.ENABLE_AGGREGATION_MODE,
```
justincorrigible, line 15: tempted to suggest/request a code comment on what "aggregation" means in this context, as it could be mistaken as a toggle to enable the existing facets' aggregations.

modules/server/src/config/constants.ts
```diff
@@ -22,3 +22,4 @@ export const PING_MS = stringToNumber(process.env.PING_MS) || 2200;
 export const PING_PATH = process.env.PING_PATH || '/ping';
 export const PORT = stringToNumber(process.env.PORT) || 5050;
 export const ROW_ID_FIELD_NAME = process.env.ROW_ID_FIELD_NAME || 'id';
+export const ENABLE_AGGREGATION_MODE = stringToBool(process.env.ENABLE_AGGREGATION_MODE);
```
justincorrigible, line 25: may I bug to keep the alphabetic order here?

modules/server/src/mapping/createConnectionResolvers.ts
```diff
@@ -40,7 +42,7 @@ const createConnectionResolvers: CreateConnectionResolversFn = ({
 				}),
 			};
 		},
-		hits: resolveHits({ type, Parallel, getServerSideFilter }),
+		...(!enableAggregationMode && { hits: resolveHits({ type, Parallel, getServerSideFilter }) }),
```
justincorrigible, line 45: does this conditional affect the `hits` results for a regular arranger instance (i.e. without federation)?

modules/server/src/app.js
ciaranschutte: would renaming the env var to ENABLE_AGGREGATION_ONLY_MODE improve it at all?

modules/server/src/mapping/createConnectionResolvers.ts
ciaranschutte: Yes.

justincorrigible: ah, second read... the negative logic threw me off.
I gather the intent was to disable the basic `hits` if the arranger node is in "Aggregation mode". didn't realise the design for the federated nodes was an "either or".
seems like I'm misunderstanding this flag?
```suggestion
		...(enableAggregationMode || { hits: resolveHits({ type, Parallel, getServerSideFilter }) }),
```

modules/server/src/mapping/createConnectionTypeDefs.js
```diff
@@ -1,6 +1,26 @@
...
+    offset: Int
+    sort: [Sort]
+    filters: JSON
+    before: String
+    after: String
+    first: Int
+    last: Int
+    searchAfter: JSON
+    trackTotalHits: Boolean = true
+  ): ${name}Connection`;
```
justincorrigible, line 20: is this indentation "off" on purpose? or just something Prettier didn't catch?

modules/server/src/schema/Root.js
```diff
@@ -1,16 +1,16 @@
...
 import { GraphQLDate } from 'graphql-scalars';
+import { GraphQLJSON } from 'graphql-type-json';
 import { startCase } from 'lodash';
 import Parallel from 'paralleljs';
 
-import { createConnectionResolvers, saveSet, mappingToFields } from '@/mapping';
+import { createConnectionResolvers, mappingToFields, saveSet } from '@/mapping';
 import { checkESAlias, getESAliases } from '@/mapping/utils/fetchMapping';
 
+import { DEBUG_MODE } from '@/config/constants';
```
justincorrigible, line 9: this can be moved to line 6 to stay with the 'internal' imports

modules/server/src/mapping/mappingToFields.js
```diff
@@ -1,11 +1,12 @@
...
-import mappingToNestedFields from './mappingToNestedFields';
-import mappingToScalarFields from './mappingToScalarFields';
 import createConnectionTypeDefs from './createConnectionTypeDefs';
+import mappingToNestedFields from './mappingToNestedFields';
 import mappingToObjectTypes from './mappingToObjectTypes';
+import mappingToScalarFields from './mappingToScalarFields';
 
-const mappingToFields = ({ type, parent }) => {
+const mappingToFields = ({ enableAggregationMode, type, parent }) => {
+	const fieldsToExclude = enableAggregationMode ? ['hits'] : [];
```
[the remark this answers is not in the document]
justincorrigible: followup note: in a quick conversation over zoom, Ciarán clarified for me that this feature is meant to cover the case where it may be desirable to limit the functionality of a node to only provide aggregations. (i.e. a limited data node)

joneubank:
I think I would flip the logic of this flag... enabling this mode actually just disables the individual record queries, so we should instead have a feature like `ENABLE_DOCUMENT_HITS` and have this defaulted to `true`.

modules/server/src/app.js
joneubank: As far as adding comments, these should be JSDoc style comments on the Arranger function. This is a good place to describe the purpose of all properties of the configuration object.

joneubank (approved):
👏

PR #909, opened 2024-11-25
Feature/threshold calc

ciaranschutte:
- add second conditional `hits` resolution which uses the `aggregations` calculation
- apply data masking of `<threshold value>` to `doc_count`
- add property `belowThreshold` property to `aggregations.buckets`


modules/server/src/mapping/createConnectionResolvers.ts
```diff
@@ -24,35 +26,73 @@ const createConnectionResolvers: CreateConnectionResolversFn = ({
...
+	const defaultHitsResolver = resolveHits({ type, Parallel, getServerSideFilter });
+	const hitsResolver = enableDocumentHits
+		? defaultHitsResolver
+		: async (obj, args, context, info) => {
+				/*
+				 * Checks for aggregations field in full query and retrieves args
+				 * Popular parsing `info` libs do not include these operations properties
+				 */
+				const typeNameConnectionProperty = info.operation.selectionSet.selections[0];
+				const isAggregationsQueried = typeNameConnectionProperty.selectionSet.selections.some(
```
ciaranschutte, line 63: if `aggregations` field is queried also, pull out the variables and call the aggregation resolving code, then resolve hits from the result.

the aggregation resolver code can be memoized so it's only called once. better than sharing potentially huge objects over context

modules/server/src/mapping/createConnectionResolvers.ts
```diff
@@ -24,35 +25,81 @@ const createConnectionResolvers: CreateConnectionResolversFn = ({
...
+				: type.extendedFields,
+			...(createStateResolvers && {
+				facets: type.config?.[ConfigProperties.FACETS],
+				matchbox: type.config?.[ConfigProperties.MATCHBOX],
+				table: type.config?.[ConfigProperties.TABLE],
+			}),
+		};
+	};
+
+	const aggregationsQuery = resolveAggregations({ type, getServerSideFilter });
```
ciaranschutte, line 45: need to use this query for both "hits" and "aggregations" resolvers

modules/server/src/mapping/createConnectionResolvers.ts
```diff
@@ -24,35 +25,81 @@ const createConnectionResolvers: CreateConnectionResolversFn = ({
...
+	const aggregationsSelectionSet = fileNameConnectionProperty.selectionSet.selections.find(
+		(selection) => selection.name.value === 'aggregations',
+	);
+
+	if (aggregationsSelectionSet) {
+		const modifiedInfo = { ...info, fieldNodes: [aggregationsSelectionSet] };
+		const aggregations = await aggregationsQuery(obj, info.variableValues, context, modifiedInfo);
+		const { hitsTotal: total } = applyAggregationMasking({
+			aggregations,
+			thresholdMin: 200,
```
ciaranschutte, line 97: @joneubank pass in as env var or a config var?

modules/server/src/config/constants.ts
```diff
@@ -4,6 +4,7 @@ export const ALLOW_CUSTOM_MAX_DOWNLOAD_ROWS = stringToBool(
 	process.env.ALLOW_CUSTOM_MAX_DOWNLOAD_ROWS,
 );
 export const CONFIG_FILES_PATH = process.env.CONFIG_PATH || './configs';
+export const DATA_MASK_THRESHOLD = process.env.CONFIG_PATH || Number.MAX_SAFE_INTEGER;
```
justincorrigible, line 7: Methinks `process.env.CONFIG_PATH` won't do much good here

modules/server/src/mapping/createConnectionResolvers.ts
```diff
@@ -21,38 +23,85 @@ const createConnectionResolvers: CreateConnectionResolversFn = ({
...
+};
+
+/**
+ * Resolve hits from aggregations
+ * If "aggregations" field is not in query, return 0
+ *
+ * @param aggregationsQuery - resolver ES query code for aggregations
+ * @returns Returns a total count that is less than or equal to the actual total hits in the query.
+ */
+const resolveHitsFromAggs =
```
justincorrigible, line 83: This function should be declared before its usage to facilitate reading, therefor needs to be moved before line 56, or better yet, could be abstracted along with other helper functions into a separate (resolvers) file to separate concerns e.g. "resolving X" vs "connecting the resolvers".

modules/server/src/mapping/createConnectionResolvers.ts
```diff
@@ -21,38 +23,85 @@ const createConnectionResolvers: CreateConnectionResolversFn = ({
...
+	(aggregationsQuery, dataMaskThreshold) => async (obj, args, context, info) => {
+		/*
+		 * Get "aggregations" field from full query if found
+		 * Popular gql parsing libs parse the "info" property which may not include full query based on schema
+		 */
+		const fileNameConnectionProperty = info.operation.selectionSet.selections[0];
```
justincorrigible, line 89: are all the segments of this chain guaranteed to exist? (including the 0th in the selections)

modules/server/src/mapping/createConnectionResolvers.ts
```diff
@@ -21,38 +23,85 @@ const createConnectionResolvers: CreateConnectionResolversFn = ({
...
+		const aggregationsSelectionSet = fileNameConnectionProperty.selectionSet.selections.find(
+			(selection) => selection.name.value === 'aggregations',
+		);
+
+		if (aggregationsSelectionSet) {
```
justincorrigible, line 94: it's not immediately clear here why this condition failing (`else`) would mean total: 0.
a comment would help in order to document the intent and rationale behind this logic

modules/server/src/mapping/mappingToFields.js
```diff
@@ -29,7 +29,7 @@ const mappingToFields = ({ enableDocumentHits, type, parent }) => {
 				type.customFields,
 			],
 			createStateTypeDefs: 'createState' in type ? type.createState : true,
-			fieldsToExclude,
+			showRecords,
```
justincorrigible, line 32: given the meaningful ternaries within `createStateTypeDefs` relying on negatives, it may be good to keep the same original name here (`enableDocumentHits`) to facilitate keeping the larger mental context. i.e. changing the name to this synonym doesn't add clarity while it does add complexity, now having to keep in mind "this means that".

...or better yet, make this `hideHits` if you seek to add clarity through simplicity, with a bonus of "not asking negatives" within that other scope.

modules/server/src/mapping/masking.ts
```diff
@@ -0,0 +1,92 @@
...
+			bucket_count: number;
+			buckets: Array<{
+				doc_count: number;
+				key: string;
+			}>;
+		}
+	>;
+	thresholdMin: number;
+}) => {
+	const THRESHOLD_REPLACEMENT_VALUE = thresholdMin - 1;
```
justincorrigible, line 52: this merits an explanation comment

modules/server/src/schema/Aggregations.ts
```diff
@@ -13,6 +18,7 @@ export let typeDefs = `
     key_as_string: String
     top_hits(_source:[String], size:Int): JSON
     filter_by_term(filter: JSON): JSON 
+    ${!enableDocumentHits ? 'belowThreshold: Boolean' : ''}
```
justincorrigible, line 21: guess this is another spot where `hideHits` would be semantically beneficial.
good comment for context! :+1:

modules/server/src/mapping/createConnectionResolvers.ts
ciaranschutte: https://github.com/overture-stack/arranger/pull/909/commits/46080c121c76e45c599cd9a7ecd97cd21e48ea8e

modules/server/src/mapping/createConnectionResolvers.ts
ciaranschutte: https://github.com/overture-stack/arranger/pull/909/commits/75cb5ab2b25381f200a7798630b92f58713a5d2a
`info` object will always be defined, defaults to empty array

modules/server/src/mapping/createConnectionResolvers.ts
ciaranschutte: https://github.com/overture-stack/arranger/pull/909/commits/13c7b5a282af669ed5a42ad2f8d54f16874cb061

modules/server/src/mapping/masking.ts
ciaranschutte: https://github.com/overture-stack/arranger/pull/909/commits/4581e2bd69b8b8f6a4bfe7424d06bd2692ec3af0

modules/server/src/mapping/types.ts
```diff
@@ -0,0 +1,55 @@
...
+	bucket_count: number;
+	buckets: Bucket[];
+};
+
+export type Root = Record<string, unknown>;
+
+enum Missing {
+	first,
+	last,
+}
```
ciaranschutte, line 19: common file for TS for now, there will be cleanup needed

modules/server/src/mapping/resolvers.ts
```diff
@@ -0,0 +1,114 @@
...
+	};
+
+	// hits
+	const defaultHitsResolver = resolveHits({ type, Parallel, getServerSideFilter });
+	const hits = enableDocumentHits
+		? defaultHitsResolver
+		: // @ts-ignore
+		  // typing resolveAggregations requires typing a lot of code down the chain
+		  // TODO: improve typing
+		  resolveHitsFromAggs(aggregationsQuery, dataMaskThreshold);
```
ciaranschutte, line 111: this is a point where we hit a wall with TS versus legacy JS. There's a lot of typing to be done because it starts to cascade into lots of JS functions.

modules/server/src/mapping/masking.ts
```diff
@@ -0,0 +1,95 @@
...
+		{
+			bucket_count: number;
+			buckets: Array<{
+				doc_count: number;
+				key: string;
+			}>;
+		}
+	>;
+	thresholdMin: number;
+}) => {
```
joneubank, line 50: Very helpful to see the return type declaration. This one is quite large though and may benefit from using some type aliases. No immediate suggestions from me, but the content of aggregations is probably used in multiple places and a good candidate.

Somethign like (naming only very quickly considered):
```ts
type AggregationData = {
    bucket_count: number;
    buckets: Array<{
    doc_count: number;
    key: string;
  }>;
}

const applyAggregationMasking(...): { aggregations: Record<string, AggregationData>, thresholdMin: number } => {...}
```

modules/server/src/mapping/masking.ts
joneubank: The code and comment both say that this value is set to one less than the thresholdMin, but why? What is the reason for this?

I guess while I am reading I don't know what a threshold replacement value is used for, and from the threshold rules I remember buckets being set to threshold value or 1. I don't remember the one less than rule.

modules/server/src/mapping/resolvers.ts
```diff
@@ -0,0 +1,114 @@
...
+			context: Context,
+			info: GraphQLResolveInfo,
+		) => Record<string, Aggregation>,
+		dataMaskThreshold: number,
+	) =>
+	async (obj: Root, args: Hits, context: Context, info: GraphQLResolveInfo) => {
+		/*
+		 * Get "aggregations" field from full query if found
+		 * Popular gql parsing libs parse the "info" property which may not include full query based on schema
+		 */
```
joneubank, line 36: Helpful comment, thank you.

modules/server/src/mapping/types.ts
joneubank: Are these new enums or moved from somewhere else? I would prefer us not to use TS enums, and if we do we should have constant values (ie. `first = 'first'` or `first = 0`) to prevent their literal values changing by re-ordering the enum list in the (likely to cause no error, but can cause real tricky bugs especially if enums are used between client and server). Generally speaking, prefer not to use TS enums at all. Would you consider something like:

```ts
const Modes = {
  avg: 'avg',
  max: 'max',
  min: 'min',
  sum: 'sum',
} as const satisfies Record<string, string>; // as const makes the object store literal types, not string type. satisfies provides type system checking to ensure all values are strings.
type Mode = typeof Modes[keyof typeof Modes]; // gets values as union type.
```

joneubank: Generally speaking, I don't create a dedicated "types" files, in favour of keeping types near where they need to be declared. Treat them more akin to exported functions. Theres several different concepts stored in this file, including enums used for config/input to queries, the data structure returned by our es queries, whatever Root and Context is.

I think my advice here is similar to not having "utils" files, instead name the file based on its content and export similar ideas. At the moment, this file is not too large to manage, but I always fear these become dumping grounds of any "type" information.

joneubank (approved):
Well organized code overall, thanks for this work. Comments are minor suggestions, no changes required.

PR #913, opened 2024-12-15
Feat/cleanup

ciaranschutte:
- renames `createConnectionResolvers` as a top level TS file so everything below can be typed 
- removes `utils` folder dump, created distinct files
- removes `types` folder dump, colocate types as close to source as possible, duplicates removed

- reusable typing adding eg. GQL Resolver generic
- removes `__typename` distinction in network connected aggregations type, uses singular Aggregation type throughout
- network 
  - standardises language "network" / "remote" -> "node"
  - cleans up dead code
  - simplifies types
  - additional TSDoc


modules/server/src/mapping/resolveAggregations.ts
```diff
@@ -26,7 +59,7 @@ export default ({ type, getServerSideFilter }) => {
 			nestedFieldNames,
 			filters: compileFilter({
 				clientSideFilter: resolvedFilter,
-				serverSideFilter: getServerSideFilter(context),
+				serverSideFilter: getServerSideFilter(),
```
ciaranschutte, line 62: this doesn't appear to be needed? As far as I read, es client is only thing in context and the filter ops shouldn't be used directly in that. thoughts? - Jon, Justin

modules/server/src/gqlServer.ts
```diff
@@ -0,0 +1,24 @@
...
+ * @param context The context passed to apollo-server for queries.
+ * @param info The GraphQL info object.
+ * @return Returns resolved value;
+ */
+export type Resolver<Root = {}, QueryArgs = Object, ReturnValue = undefined> = (
+	root: Root,
+	args: QueryArgs,
+	context: Context,
+	info: GraphQLResolveInfo,
+) => ResolverOutput<ReturnValue> | ResolverOutput<ReturnValue>;
```
ciaranschutte, line 24: Add reusable GQL resolver type

PR #920, opened 2025-03-14
Feat/fed search cleanup

ciaranschutte:
- remove unused code
- clean up env config with correct parse functions, default and alphanumeric order
- cleanup schema, mapping and startup config types and functions

### DIFF
--- a/modules/server/src/gqlServer.ts
+++ b/modules/server/src/gqlServer.ts
@@ -1,0 +1,24 @@
+import { Client } from '@elastic/elasticsearch';
+import { GraphQLResolveInfo } from 'graphql';
+
+export type Context = {
+	esClient: Client;
+};
+
+export type ResolverOutput<T> = T | Promise<T>;
+
+/**
+ * GQL resolver
+ *
+ * @param root - Parent object of a query.
+ * @param args - Query arguments.
+ * @param context - Context passed to apollo-server for queries.
+ * @param info - GraphQL info object.
+ * @return Returns resolved value;
+ */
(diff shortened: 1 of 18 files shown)


modules/server/src/gqlServer.ts
```diff
@@ -0,0 +1,24 @@
+import { Client } from '@elastic/elasticsearch';
+import { GraphQLResolveInfo } from 'graphql';
+
+export type Context = {
+	esClient: Client;
```
justincorrigible, line 5: TODO: we will need to abstract this into  an ESClient provider, so users can switch seamlessly between ES v7 and v8

modules/server/src/graphqlRoutes.js
```diff
@@ -243,7 +243,7 @@ const createEndpoint = async ({ esClient, graphqlOptions = {}, mockSchema, schem
 	return router;
 };
 
-const createSchemasFromConfigs = async ({
+export const createSchemasFromConfigs = async ({
```
justincorrigible, line 246: 🙏 I knew we'd need this one at some point, just couldn't remember which was the one

modules/server/src/mapping/resolveAggregations.ts
```diff
@@ -48,6 +84,7 @@ export default ({ type, getServerSideFilter }) => {
 		const response = await esSearch(esClient)({
 			index: type.index,
 			size: 0,
+			// @ts-expect-error - valid search query parameter in ES 7.17, not in types
```
justincorrigible, line 87: hmmm!

justincorrigible (approved):

PR #921, opened 2025-03-15
Feat/fed search networked cleanup

ciaranschutte:
- colocate and consolidate types
- remove redundant code
- fix a couple of bad merges (missing await, missing gql type)


ciaranschutte:
investigating failing tests

modules/server/src/graphqlRoutes.js
```diff
@@ -276,15 +276,21 @@ export const createSchemasFromConfigs = async ({
 		 * Federated Network Search
 		 */
 		if (enableNetworkAggregation) {
+			const networkConfig = configsFromFiles[ConfigProperties.NETWORK_AGGREGATION];
```
justincorrigible, line 279: think this may help reading
```suggestion
			const networkConfigsObj = configsFromFiles[ConfigProperties.NETWORK_AGGREGATION];
```

modules/server/src/mapping/masking.ts
```diff
@@ -1,10 +1,11 @@
 import { ENV_CONFIG } from '@/config';
-import { Aggregation } from './types';
+import { Aggregations, AllAggregationsMap } from './resolveAggregations';
```
justincorrigible, line 2: are these types, classes? think if they're types, we could use adding the explicit import (e.g. `import type { ... }`) for absolute clarity on what these are

modules/server/src/mapping/resolveAggregations.ts
```diff
@@ -1,40 +1,65 @@
 import getFields from 'graphql-fields';
 
-import { buildAggregations, buildQuery, flattenAggregations } from '../middleware';
-
-import { Resolver } from '@/gqlServer';
+import { Resolver, Root } from '@/gqlServer';
```
justincorrigible, line 3: same comment about explicit type imports

modules/server/src/network/utils/promise.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Type guard to filter fulfilled Promises
+ */
+export const fulfilledPromiseFilter = <Result>(result: unknown): result is Result => {
+	return (
+		typeof result === 'object' &&
+		result !== null &&
+		'status' in result &&
+		result.status === 'fulfilled'
```
justincorrigible, line 9: thinking out loud to learn more, not a challenge or request for change:
curiously, "fulfilled" promises make sense, but I'd always thought of them as "resolved". it's a curious choice of synonym that I feel I may have missed became a standard somewhere.
was this "status == 'fulfilled'" a choice for the implementation here? or is it the output from a library? or simply a standard pattern I was not aware of?

modules/server/src/network/typeDefs/index.ts
```diff
@@ -1,4 +1,4 @@
-import { SupportedNetworkFieldType } from '../types/types';
+import { SupportedNetworkFieldType } from '../setup/fields';
```
justincorrigible, line 1: please `import type {`

modules/server/src/network/resolvers/query.ts
```diff
@@ -0,0 +1,106 @@
+import { gql } from 'apollo-server-core';
+import { DocumentNode } from 'graphql';
+import { isEmpty } from 'lodash';
+
+import { NodeConfig } from '../setup/query';
+import { RequestedFieldsMap } from '../utils/gql';
```
justincorrigible, line 6: these looks like they could be imports from `@/network/...`?

modules/server/src/network/resolvers/fetch.ts
```diff
@@ -0,0 +1,68 @@
...
+ */
+export const fetchData = async <SuccessType>(
+	query: NetworkQuery,
+): Promise<Result<SuccessType, typeof CONNECTION_STATUS.ERROR>> => {
+	const { url, gqlQuery, queryVariables } = query;
+
+	console.log(`Fetch data starting for ${url}..`);
+
+	try {
+		const response = await fetchGql({
```
justincorrigible, line 30: thought, not challenge or request: curious to learn how this fetchGql is different from the regular ajax call in the OG Arranger utils, and eager to see what kind of improvements and new features this one would bring if the two can be unified into a single reusable one for the whole server

modules/server/src/network/resolvers/aggregations.ts
```diff
@@ -1,239 +1,85 @@
...
-			nodeInfo.push({
-				name: nodeName,
-				hits: 0,
-				status: CONNECTION_STATUS.ERROR,
-				errors: response?.message || 'Error',
-				aggregations: config.aggregations,
-			});
-		}
-	});
+	await Promise.allSettled(
```
justincorrigible, line 41: TIL!!!

I do wonder how the community is not worried about having so many ambiguously synonymous words in the implementation
`resolved` vs `settled` vs `fulfilled`. it's all getting a bit wonky in my head now

modules/server/src/network/resolvers/aggregations.ts
```diff
@@ -1,239 +1,85 @@
...
+		configs.map(async (config) => {
+			// create node query
+			const gqlQuery = createNetworkQuery(config, requestedAggregationFields);
+
+			// query node
+			const response = gqlQuery
+				? await fetchData<SuccessResponse>({
```
justincorrigible, line 48: is this an instance of the fetchGql function I saw elsewhere? how are those two different? 🤔 
I'm starting to get confused here ("me" problem, cognitive load)

modules/server/src/network/resolvers/aggregations.ts
```diff
@@ -1,239 +1,85 @@
...
+						url: config.graphqlUrl,
+						gqlQuery,
+						queryVariables,
+				  })
+				: failure(CONNECTION_STATUS.ERROR, 'Invalid GQL query');
+
+			const nodeName = config.displayName;
+
+			if (isSuccess(response)) {
+				const documentName = config.documentName;
```
justincorrigible, line 58: this could better be destructured to improve DRY

modules/server/src/network/resolvers/aggregations.ts
justincorrigible: hm i see now this one was the function that summoned the other one where I left that earlier comment...
though it's obvious I need more coffee, I think less generic / more verbose variable names could help keep track of things as one adds more code into mental context

modules/server/src/network/index.ts
```diff
@@ -1,46 +1,61 @@
...
+import { normalizeFieldTypes } from './setup/fields';
 import { fetchAllNodeAggregations } from './setup/query';
 import { createTypeDefs } from './typeDefs';
 import { NetworkConfig } from './types/setup';
 
-export let ALL_NETWORK_AGGREGATION_TYPES_MAP: Map<string, SupportedAggregation> = new Map();
+/**
+ * Map of all available fields with associated aggregation type
+ */
+export let ALL_NETWORK_AGGREGATION_TYPES_MAP: Map<string, keyof typeof SUPPORTED_AGGREGATIONS> =
```
justincorrigible, line 12: `let` seems atypical here. if I didn't know better I'd ask if an AI assistant suggested this here

modules/server/src/gqlServer.ts
```diff
@@ -16,9 +18,9 @@ export type ResolverOutput<T> = T | Promise<T>;
  * @param info - GraphQL info object.
  * @return Returns resolved value;
  */
-export type Resolver<Root = {}, QueryArgs = Object, ReturnValue = undefined> = (
+export type Resolver<Root = Record<string, any>, QueryArgs = Object, ReturnValue = undefined> = (
```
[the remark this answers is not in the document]
ciaranschutte: aliasing type because generic name can't be same as typename

modules/server/src/mapping/resolveAggregations.ts
```diff
@@ -1,40 +1,65 @@
...
 
-export type Aggregation = {
+export type CommonAggregationProperties = {
 	bucket_count: number;
 	buckets: Bucket[];
 };
 
-export type Aggregations = Record<string, Aggregation>;
+// the GQL Aggregations type
+export type Aggregations = CommonAggregationProperties;
```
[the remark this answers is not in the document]
ciaranschutte: Yes, its the shared aggregations fields
example: NumericAggregations would use this as a base

modules/server/src/network/resolvers/aggregations.ts
ciaranschutte: it's a bit overloaded cognitively for sure!
the `fetchGQL` is just an axios call with configured request options - headers, query vars
this `fetchData` calls `fetchGQL` - formats the GQL query string for the requests, handles response errors, returns success/failure objects

modules/server/src/network/resolvers/aggregations.ts
ciaranschutte: might be more confusing... but my understanding:
resolved = success
fulfilled = settled = success | failure

Just so happens that the Promise api uses the word "settled" instead of "fulfilled"

modules/server/src/gqlServer.ts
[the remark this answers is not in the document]
justincorrigible: so maybe the `<Root` here could have a different type name than the shared/exported one?

ciaranschutte:
@justincorrigible thanks for review - I've updated to explicit type imports and alias import mapping

modules/server/src/network/resolvers/fetch.ts
justincorrigible: think the other existing one also does that if I'm not mistaken... maybe I'm thinking of one in the UI components though 🤔

modules/server/src/network/resolvers/aggregations.ts
justincorrigible: sounds like a good reason confirming the need for me to abstract all these into a single shared set of functions later on, once this is all merged

justincorrigible (approved):
changes LGTM, but my scope of business logic is limited

ciaranschutte:
tests failing outside of this branch, can merge into main feature branch while fixing tests

PR #936, opened 2025-04-12
Typescript module server

ciaranschutte:
- converts all files to TS
- ensures `type` is correctly in imports
- uses underscore to denote unused param, supported in TS as unused param eg. `(_req, res)...`
- add `getCaughtErrorMessage`
  - type guards for caught thrown errors 
- moves ES client to module and define consistent type from "new api" folder (this type returns typings from api calls)
  - which also means ES functions require more strict typing overall eg. providing full set of object properties 
- ConfigObject is no longer a partial type 
 - config becomes an extended object which complicates types, leaving most of this as `any` for now
 - config "prop drilling" has a lot of duplicated typing for now, refactoring config has been discussed

Further improvements: 
- location of types. Decide if they are colocated or a distinct "types" folder


modules/server/src/esClient/index.ts
```diff
@@ -0,0 +1,27 @@
...
+			password: esPass,
+		};
+	}
+
+	return new Client(esConfig);
+};
+
+export const buildEsClientViaEnv = (ES_HOST: string, ES_USER: string, ES_PASS: string) => {
+	return buildEsClient(ES_HOST, ES_USER, ES_PASS);
+};
```
ciaranschutte, line 27: move ES client to module

README.md
```diff
@@ -19,7 +19,7 @@ Technical resources for those working with or contributing to the project are av
 ## Development Environment
 
 - [Node.js](https://nodejs.org/) (v22 or higher)
-- [Docker](https://www.docker.com/) (v4.32.0 or higher)
```
justincorrigible, line 22: Is this newer version a requirement? recommended? or merely the latest?

justincorrigible: second thought, seems like something @MitchellShiell  changed.
I'll follow up with him separately...

modules/server/src/esClient/index.ts
```diff
@@ -0,0 +1,27 @@
+import { Client, type ClientOptions } from '@elastic/elasticsearch';
+
+export const buildEsClient = (esHost = '', esUser = '', esPass = '') => {
+	if (!esHost) {
```
justincorrigible, line 4: wouldn't it be fair to provide a default value for this?
ES has a common 9200 port, and localhost is a usual

modules/server/src/graphqlRoutes.ts
```diff
@@ -294,15 +307,26 @@ export const createSchemasFromConfigs = async ({
 			schema: fullSchema,
 		};
 	} catch (error) {
-		const message = error?.message || error;
+		const message = getCaughtErrorMessage(error);
```
justincorrigible, line 310: 👍 one more from my ToDo:s, thanks for handling it!

modules/server/src/graphqlRoutes.ts
```diff
@@ -9,8 +9,12 @@ import { extendColumns, extendFacets, flattenMappingToFields } from './mapping/e
 import { addMappingsToTypes, extendFields, fetchMapping } from './mapping/index.js';
 import makeSchema from './schema/index.js';
 import { createSchemaFromNetworkConfig } from './network/index.js';
+import type { Client } from '@elastic/elasticsearch';
+import type { SQONFilter } from '#sqon/index.js';
+import { getCaughtErrorMessage } from '#utils/caughtError.js';
+import type { GraphQLSchema } from 'graphql';
```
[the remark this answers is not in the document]
ciaranschutte: handy extension thank you 👍  they become "out of sync" with VSCode "import this file" which we now have after ts-ify 👍

PR #944, opened 2025-05-02
Remove data masking

ciaranschutte:
Remove data masking references and additional resolver functions for generating hits based on "masked aggregations"
Removes env var refs, additional response properties, resolver code and type def configs

A snapshot of this work is here for future reference https://github.com/overture-stack/arranger/tree/draft/data_masking


ciaranschutte:
a bunch of conflicts created with latest push to main, fixing these

modules/server/src/network/resolvers/index.ts
```diff
@@ -35,7 +35,7 @@ export const createResolvers = (configs: NodeConfig[]) => {
 		 * Checks validity of SQON
 		 * For now we will pass through the non SQON object to the pipeline
 		 *
-		 * TODO: resolve Arranger / SQONBuilder SQON outer wrapper conflict
+		 * // TODO: resolve Arranger / SQONBuilder SQON outer wrapper conflict
```
justincorrigible, line 38: curious double comment. no action needed, just wondering.

justincorrigible (approved):

modules/server/src/network/resolvers/index.ts
ciaranschutte: Great q. The TODO for vscode plugin picks up `//` but not how it was originally.

PR #947, opened 2025-05-06
Feat/network search config

ciaranschutte:
- add ability to provide federated network aggregations from .env file and not just file input
- added string util to parse json array, in string utilities consistent with other util functions
- update example and boilerplates


modules/server/configTemplates/network.json
```diff
@@ -1,5 +1,3 @@
 {
-	"network": {
-		"servers": []
-	}
+	"network": []
```
ciaranschutte, line 2: reduced this to just servers instead of object

modules/server/src/config/types.ts
```diff
@@ -158,7 +157,7 @@ export interface ConfigObject {
 	[ConfigProperties.INDEX]: string;
 	[ConfigProperties.MATCHBOX]: any[];
 	[ConfigProperties.TABLE]: TableConfigsInterface;
-	[ConfigProperties.NETWORK_AGGREGATION]: NetworkAggregationInterface;
+	[ConfigProperties.NETWORK_AGGREGATION]: any[];
```
ciaranschutte, line 160: matches the loose typing for now, we can tighten later

justincorrigible (approved):

PR #949, opened 2025-06-11
Feature/charts colors

ciaranschutte:
- adds consistent colors for charts using a Map
- cleanup of unused files


modules/charts/src/components/Chart.tsx
```diff
@@ -34,42 +52,36 @@ export const Chart = ({ fieldName, theme, children, headless, DisplayComponent }
...
+		//	const resolveColorFn = wrapWithFieldName;
+		// wrap all theme, instead of passing more props around, keep other interfaces well defined
+
 		return (
 			<ChartContainer>
 				<DisplayComponent
-					data={data.data}
-					config={{ ...config, colors: colors.getColors }}
+					// keep data pretty clean because we might manipulate in the charts
+					// data vs config good seperation anyway, can use functions that take data and resolve
```
justincorrigible, line 81:
```suggestion
					// data vs config good separation anyway, can use functions that take data and resolve
```

modules/charts/src/components/Provider/Colors/colors.ts
```diff
@@ -0,0 +1,44 @@
...
+
+	return {
+		/**
+		 * Creates a color map for charts
+		 *
+		 * @returns Map of keys to assigned colors
+		 */
+		createColorMap: (data: Map<string, any>) => {
+			if (!data) return;
+			if (colorMap.size !== 0) return;
```
justincorrigible, line 23: is this a good synthesis? not sure yet whether each return is meant to become different per case later. either way, best we keep curlies around conditional blocks, for both consistency and to reduce diff bloat.
```suggestion
			if (!data || colorMap.size !== 0) {
				return;
			}
```

modules/charts/src/components/Provider/Provider.tsx
```diff
@@ -0,0 +1,119 @@
...
+		isLoading: boolean;
+		isError: boolean;
+		// TODO: Map<string, Aggregtions | NumericAggregations>
+		data: Map<string, {}> | undefined;
+	};
+};
+
+export const ChartsContext = createContext<ChartContextType | null>(null);
+
+type ChartsProviderProps = React.PropsWithChildren<{
```
justincorrigible, line 21: seems like `PropsWithChildren` is the only bit that relies on the larger React import (which also includes all sorts of other properties, likely including the ones destructured at the same time).

Is there a particular reason why this one is not just destructured like the others?

modules/charts/src/components/Provider/Provider.tsx
```diff
@@ -0,0 +1,119 @@
...
+		resolveColor,
+	};
+
+	return <ChartsContext.Provider value={chartContext}>{children}</ChartsContext.Provider>;
+};
+
+export const useChartsContext = () => {
+	const context = useContext(ChartsContext);
+	if (!context) {
+		throw new Error('context has to be used within <Charts.Provider>');
```
justincorrigible, line 116: this may lead to unhandled error situations as it relies on the consumer side accounting for the lack of a "default" fallback behaviour. e.g. provide an empty context if the expected one is unavailable.

... probably feedback also applicable to the OG Arranger UI components

modules/charts/src/main.tsx
```diff
@@ -2,6 +2,6 @@ import { Barchart } from './components/charts/Barchart/Barchart';
 
 import { Headless } from './components/charts/Headless';
 
-import { ChartsProvider } from './components/Provider';
+import { ChartsProvider } from './components/Provider/Provider';
```
justincorrigible, line 5: not sure how to phrase succinctly without seeming snarky: another "almost" barrel?

modules/charts/src/components/Provider/Provider.tsx
ciaranschutte: I think yelling with a clear error on the consumer side is better than "silent failure" from a context object that doesn't make any sense.

I believe this is a similar context check to `useArrangerData` and `ArrangerDataProvider`

ciaranschutte:
re: "almost barrel" , yep these are barrels, they are intended, removed tests that weren't fully functional so the code was in better PR state. Otherwise there's a slow trickle of more and more things to for what started as a smaller scoped PR.

Alternative is don't use barrel, but then the "intent" in structure is completely lost and we end up with a project with no structure, intended or otherwise

modules/charts/src/components/Provider/Colors/colors.ts
```diff
@@ -0,0 +1,44 @@
...
+			for (const [key, value] of data.entries()) {
+				const buckets = value.buckets; // TODO: numeric agg diff property
+				buckets.forEach((bucket) => {
+					const assignedColor = colors[colorIndex++ % colors.length];
+					colorMap.set(`${key}.${bucket.key}`, assignedColor);
+				});
+			}
+			return colorMap;
```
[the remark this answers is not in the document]
ciaranschutte: Intent was not to be "clever", the array looping was visually more clear in my head than a reduce.

also, and you can add this as a counter argument to our Map talk, using `Map.set` in feels wrong.

modules/charts/src/components/Provider/Provider.tsx
```diff
@@ -0,0 +1,119 @@
...
+			const chartData = chartDataMap?.get(fieldName);
+
+			return {
+				isLoading: false,
+				isError: false,
+				data: chartData,
+			};
+		},
+		resolveColor,
+	};
```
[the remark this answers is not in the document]
ciaranschutte: https://github.com/overture-stack/arranger/pull/949/commits/065d05969cc4428ffe74c7ec5f7766273b60582f

modules/charts/src/components/Provider/Colors/colors.ts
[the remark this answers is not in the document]
justincorrigible: fair point. this is just me trying to wrap my head around the case for using Map here.. that's all

thanks for clarifying the intent 👍

modules/charts/src/components/Provider/Provider.tsx
ciaranschutte: For sure! I'd like your input on the best way to handle it

PR #954, opened 2025-07-03
wild card Emotion to externals

ciaranschutte:
Emotion was being bundled, causing React issues
This PR makes sure it, and all it's many dependants, do not get bundled
also update React peer to 17


ciaranschutte:
closing in favour of https://github.com/overture-stack/arranger/pull/955

PR #955, opened 2025-07-03
make Charts publishable

justincorrigible:
TL;DR: 
- adds server fallback/default configs for Charts and Network so tests pass
- changes manifest to be publishable through the existing Arranger pipeline@Jenkins
- removes the lock, as it is not necessary (published packages don't include one)
      the locking of dependencies happens on build and it's managed in Arranger's root lock (because NPM Workspaces)


modules/server/.env.schema
```diff
@@ -15,3 +17,4 @@ MAX_RESULTS_WINDOW=10000
 PING_MS=2200
 PING_PATH=/ping
 PORT=5050
+ROW_ID_FIELD_NAME=id
```
justincorrigible, line 20:
```suggestion
ROW_ID_FIELD_NAME=id

```

modules/charts/package.json
```diff
@@ -1,26 +1,30 @@
...
 		"storybook": "storybook dev -p 6006",
-		"build-storybook": "storybook build"
+		"test": "vitest"
 	},
 	"peerDependencies": {
 		"@emotion/react": "^11.0.0",
-		"@overture-stack/arranger-components": "^3.0.0-rc.1",
-		"react": "^18.3.1",
-		"react-dom": "^18.3.1"
+		"@overture-stack/arranger-components": "*",
```
ciaranschutte, line 23: This is pointing to the workspace but is invalid as an actual peer dep I think?
Would work in devDep? Because we want a version for development, pointing to local workspace is fine.

ciaranschutte (approved):

PR #957, opened 2025-07-12
Feature/revert configt to frontend

ciaranschutte:
- revert Charts approach to generate on frontend instead of backend


ciaranschutte:
closing in favour of https://github.com/overture-stack/arranger/pull/962

PR #958, opened 2025-07-12
argo data discovery page feedback

ciaranschutte:
- add mouse over cursor hand

### DIFF
--- a/modules/charts/src/components/charts/Barchart/Barchart.tsx
+++ b/modules/charts/src/components/charts/Barchart/Barchart.tsx
@@ -3,18 +3,27 @@ import { ResponsiveBar } from '@nivo/bar';
 import { Chart } from '#components/Chart';
 import { ArrangerChartProps, ArrangerChartTheme } from '#theme/arranger';
 import { arrangerToNivoBarChart } from '#theme/nivo/nivo';
+import { css } from '@emotion/react';
+import { useRef } from 'react';
 
 /**
  * Resolve to a Nivo Bar chart component
  */
 export const BarchartComp = ({ data, theme }: ArrangerChartProps) => {
-	const resolvedTheme = arrangerToNivoBarChart({ data, theme });
+	// create div ref for toggling css style
+	const wrapperRef = useRef(null);
+	const resolvedTheme = arrangerToNivoBarChart({ data, theme, wrapperRef });
 
 	return (
-		<ResponsiveBar
-			data={data}
-			{...resolvedTheme}
-		/>
+		<div
+			css={css({ width: '100%', height: '100%' })}
+			ref={wrapperRef}
+		>
+			<ResponsiveBar
+				data={data}
+				{...resolvedTheme}
+			/>
+		</div>
 	);
 };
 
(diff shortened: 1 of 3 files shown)


modules/charts/src/components/Chart.tsx
```diff
@@ -23,10 +23,23 @@ type ChartProps = {
...
+
+	// consumer
+	if (onDataLoad) {
+		chartData = onDataLoad(chartData);
+	}
+
+	return chartData;
 };
 
 /**
```
ciaranschutte, line 45: Arranger Charts currently only supports "aggregations" field of Arranger Server
There's nothing implemented to grab a total for field, this allows the client to at least grab all the buckets and compute a total. escape hatch until Arranger Charts is feature complete.

ciaranschutte: additionally we will need to send back the typenames for the fields,

justincorrigible (approved):
this branch will try to release a new npm version, but won't without the bump in package.json. is that the intent?

modules/charts/README.md
```diff
@@ -1,4 +1,20 @@
...
-- package.json "files" set what's installed as dependency, package.json automatically included
-- `npm i <this_project_folder>` from consumer project
+# Introduction
+
+## Assumptions
+
+Arranger Charts uses <ArrangerChartsProvider> which internally uses Arranger Components hooks.
+Please make sure <ArrangerChartsProvider> is nested under an instance of <ArrangerDataProvider>
+
+This functionality is needed to query an Arranger server and to update SQON state in an Arranger application.
```
joneubank, line 8: Is there some default behaviour when one (or both) of the Providers are missing? Perhaps it puts an error message intended for the developer to communicate that a provider is missing? Should we document that default behaviour here?

modules/charts/src/components/Chart.tsx
```diff
@@ -23,10 +23,23 @@ type ChartProps = {
...
+		chartData = data.range.buckets;
+	}
+
+	// consumer
+	if (onDataLoad) {
+		chartData = onDataLoad(chartData);
+	}
+
+	return chartData;
 };
```
joneubank, line 43: Unclear what the data type returned here is. I'm sure this is inferred correctly but for a reader its very unclear without function signature.

modules/charts/src/components/Chart.tsx
```diff
@@ -89,16 +102,17 @@ export const Chart = ({ fieldName, theme, headless, children, DisplayComponent }
...
+		const resolvedChartData = resolveData({ data: chartData, onDataLoad: theme.onDataLoad });
+
+		if (isEmpty(resolvedChartData)) {
+			const { EmptyData } = globalTheme.components;
+			return (
+				<ChartContainer>
+					<EmptyData />
+				</ChartContainer>
+			);
+		}
```
joneubank, line 115: 👍

modules/charts/src/theme/nivo/config.tsx
```diff
@@ -0,0 +1,118 @@
+export const defaultNivoConfig = {
```
joneubank, line 1: Missing types for this. Should this be based on Nivo config type, or is this our own interface shape that we should define and export?

joneubank (approved):
Looks pretty nice, have some comments directed to help think about types to make things clearer for future devs.

PR #964, opened 2025-07-23
Feature/charts configs

ciaranschutte:
- chart props proposal, common for all charts
- clear separation between the data and ui parts
- a lot of Arranger theme specifics are not applicable because we're dealing with SVGs not CSS for the most part
- we can still of course use it internally eg. "match the chart colours with Arranger powered portal components"
- good defaults for everything internally, only required field is likely a field name
eg
```tsx
<BarChart data={{fieldNames: ['age_at_diagnosis]}/>
```
---

flux style `onAction` to surface any events to the consumer of the library in a standard way
`action` will be fully typed with action types exported for use

---

- use zod so we can validate the users config, there's potentially a lot of configuration for customization for a chart, good to valiate
- single source of truth for our TS types by generating them from zod

----
I think the main alternative would be to have one large config object called `theme` and stick everything in that.
```tsx
<BarChart fieldNames={['age_at_diagnosis']} theme={{....}}/>
```

### DIFF
--- a/modules/charts/package.json
+++ b/modules/charts/package.json
@@ -27,7 +27,6 @@
 	},
 	"peerDependencies": {
 		"@emotion/react": "^11.0.0",
-		"@overture-stack/arranger-components": "*",
 		"@types/react": "^17.0.0 || ^18.0.0",
 		"@types/react-dom": "^17.0.0 || ^18.0.0",
 		"react": "^17.0.0 || ^18.0.0",
@@ -39,12 +38,10 @@
 		"@nivo/pie": "^0.88.0",
 		"@nivo/sunburst": "^0.88.0",
 		"lodash": "^4.17.21",
-		"react-error-boundary": "^6.0.0",
 		"zod": "^4.0.5"
 	},
 	"devDependencies": {
 		"@eslint/js": "^9.28.0",
-		"@overture-stack/arranger-components": "file:../../../platform-ui/node_modules/@overture-stack/arranger-components",
 		"@storybook/addon-docs": "^9.0.5",
 		"@storybook/addon-onboarding": "^9.0.5",
 		"@storybook/react-vite": "^9.0.5",
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,7 +33,7 @@
 		},
 		"../platform-ui/node_modules/@overture-stack/arranger-components": {
 			"version": "3.0.0-rc.1",
-			"dev": true,
+			"extraneous": true,
 			"dependencies": {
 				"@emotion/react": "^11.0.0",
 				"@emotion/styled": "^11.0.0",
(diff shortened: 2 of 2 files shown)


modules/charts/src/schema.ts
```diff
@@ -0,0 +1,32 @@
...
+			orderBy: z.custom<(param: any) => any>(),
+		}),
+	}),
+	chart: z.object({
+		onAction: z.custom<({ action, data }: { action: ChartActions; data: any }) => any>(),
+		xAxis: AxisSchema,
+		yAxis: AxisSchema,
+	}),
+});
+
```
ciaranschutte, line 31:
```ts
type ChartProps = {
    data: {
        fieldNames: string[];
        onAction: ({ action, data }: {
            action: DataActions;
            data: any;
        }) => any;
        buckets: {
            orderBy: (param: any) => any;
        };
    };
    chart: {
        onAction: ({ action, data }: {
            action: ChartActions;
            data: any;
        }) => any;
        xAxis: {
            ...;
        };
        yAxis: {
            ...;
        };
    };
}
```

ciaranschutte:
```tsx
<BarChart
	data={{ fieldNames: ['age_at_diagnosis'] }}
	chart={{ xAxis: { showAxisTitle: true } }}
/>;
```

justincorrigible (commented):
thinking how to articulate my comment about `data "actions"` being just `chart (internal) "events"`, I'm still for separating data input object from chart configuration.
```tsx
<Chart
  data={[{...},{...}]}
  [config/theme]={{...}}
/>
```

thus, surely my input is useful here though as seeing this code just confirmed my own biases... I'll defer to others' preference and approval for this PR

ciaranschutte:
> posting some thoughts from additional insight gained, after chatting over Slack. but first, here's a 3 point summary of that conversation:
> 
> * "Events" may lead to dissonance like "why is this event not like the HTML spec" (semantic overlap)
> * "actions" come from flux lingo, and as such they're used to handle that dissonance.
> * abstracting that "action handling" under the hood, exposing just the handler options naming them descriptively may be the right approach here.
> 
> now my own conclusion for what could be done here (i.e. may not be the right approach, but it's where I'm leaning):
> 
> * a chart would have the following props (or more): children (?), className, css (emotion support), data (just the stuff to visualize, no bells and whistles), handlers (to be described in the next point, theme, etc.
> * `handlers` grouped together and outside of theme, would group (but not merge into a single function) all the callbacks to be executed upon chart events, like so
> 
> ```js
> handlers={ // functions for each event supported
>   onError,
>   onLoad,
>   onWheelClick,
>   onButtonHover,
>   ...
> }
> ```
> 
> * the actual "actions" and their type would be handled internally within the chart, without exposing the flux paradigm to the user. this allows for more explicit clarity on what the user can do or not, self documenting the code without over-relying on TS acrobatics.

No issues exposing explicit handlers object, it's a good idea.

PR #1016, opened 2026-01-20
Fix uncaught type error in config for non-SSR usage

bioladdev:
Fixes uncaught type error for `process` variable.

@overture-stack/arranger-components can only work in frameworks that support SSR because `process` being available is assumed.

A simple check is added to make sure `process` is available, and if not, defaults to a safe empty object. This allows @overture-stack/arranger-components to work client side without SSR capabilities.


justincorrigible:
think there's some easy actionables here we can implement for the next release. thanks again!

PR #1021, opened 2026-01-29
#1141 Set Up SearchClient :mag:

demariadaniel:
## Summary
Base feature branch for supporting multiple Search Client options in Arranger

## Issues
- https://github.com/nci-hcmi-catalog/portal/issues/1141

## Description of Changes
- Creates a shared SearchClient function for determining configuration and returning appropriate client
- Adds OpenSearch client library
- Initial Type structure for SearchClient
- Prettier & Typescript cleanup across multiple files

## Setup
- Requires Arranger npm install:
```
cd /arranger
npm i
```

- Requires running ElasticSearch / OpenSearch instances
- Basic OpenSearch setup:
https://docs.opensearch.org/latest/getting-started/quickstart/#option-1-try-opensearch-in-one-command

- Requires mock Arranger configs

- ... then run `npm run server`

Successful connection to Arranger server with both OpenSearch and Elastic setups:

OpenSearch:
Screenshot 2026-02-11 at 2 40 03 PM

ElasticSearch:
Screenshot 2026-02-11 at 3 13 08 PM

## Readiness Checklist

- [ ] **Self Review**
- [ ] **Formatting**
- [ ] **Local Testing**
- [ ] **Updated Tests**
- [ ] Documentation


modules/server/src/searchClient/index.ts
```diff
@@ -0,0 +1,23 @@
+import { Client as ElasticClient, type ClientOptions as ESClientOptions } from '@elastic/elasticsearch';
+import { Client as OpenSearchClient, type ClientOptions as OSClientOptions } from '@opensearch-project/opensearch';
+
+import { ENV_CONFIG } from '#config/index.js';
+
+const SearchClient = async (options: ESClientOptions | OSClientOptions) => {
+	const { ES_HOST } = ENV_CONFIG;
+	const searchConfig = await (await fetch(ES_HOST)).json();
+	const { distribution } = searchConfig.version;
```
demariadaniel: See:
https://docs.opensearch.org/latest/api-reference/cluster-api/info/

https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-info

Cluster Info provides a `distribution` field which lets us clearly identify when an OpenSearch instance is running.

This will need to be expanded when investigating ElasticSearch 8. But for OpenSearch hopefully the distribution field will suffice.

joneubank (changes requested):
So great seeing this working end to end. Thanks for the big effort.

I saw quite a few places for improvement in how we are handling the configuration and setup for arranger server now that it supports multiple search services, so that's where this feedback focuses.

We also require some updates to documentation for arranger to communicate to operators about the search supports and how to configure it. @b-f-chan this may fit better into a separate ticket. @demariadaniel for this review, let's just make sure we document somewhere the valid options for the SEARCH_SERVICE_TYPE env var.

There is future work needed to make our SearchClient type strongly typed without all the anys. That will take some dedicated time and effort to get right so I couldn't fit it into this round of feedback comments.

Thanks again Dan!

modules/server/src/admin/index.ts
```diff
@@ -2,100 +2,99 @@
...
+				index: constants.ARRANGER_PROJECT_INDEX,
+			});
+		}
+		return esClient;
+	} catch (err) {
+		setTimeout(async () => {
+			return await initialize(config);
+		}, 1000);
+	}
+};
```
joneubank, line 89: This retry handling is incomplete.
- we are requesting a retry of the initialization attempt, but not awaiting for the result and not returning an esClient from the retry.
- no logging to indicate what errors are occurring and what action the server will take
- there is no handling of the error itself. some errors can't be fixed through a retry. Retry-able errors would be network errors (timeout, connection failure) or similar, but if the error states our request is broken then retrying it is not likely to fix the issue.
- This will retry infinitely if it consistently throws an error. Is that what we want?

modules/server/src/admin/services/elasticsearch/index.ts
```diff
@@ -1,27 +1,29 @@
...
+
+export const createClient = async (esHost: string, esUser: string, esPass: string) => {
+	const esConf = { node: esHost };
+	if (esUser && esPass) {
+		esConf['auth'] = {
+			username: esUser,
+			password: esPass,
+		};
+	}
+	return await getSearchClient(esConf);
```
joneubank: There are type issues here.

`esConf` has an infered type (`{ node: string; }`) so you cant add random properties `esConf['auth']`.

More importantly, getSearchClient does not accept `.auth`, the Input type for getSearchClient looks like:
```ts
{
	node: string;
	user?: string;
	password?: string;
	client?: string;
}
```

modules/server/src/searchClient/index.ts
```diff
@@ -0,0 +1,127 @@
+import { createElasticSearchClient, type ESClientOptions } from './createElasticSearchClient.js';
+import { createOpenSearchClient, type OSClientOptions } from './createOpenSearchClient.js';
+import type { SearchClient, SupportedClientTypes, SearchConfig, SearchConfigWithClient } from './types.js';
+
+export const supportedClientValues: SupportedClientTypes[] = ['opensearch', 'elasticsearch'] as const;
```
joneubank: This `as const` isn't doing anything because you have declared the type as `SupportedClientTypes[]`. If you want this as const with appropriate type checking then use:

```ts
export const supportedClientValues = ['opensearch', 'elasticsearch'] as const satisfies SupportedClientTypes[];
```

This will then have the type: `["opensearch", "elasticsearch"]`

modules/server/src/searchClient/types.ts
```diff
@@ -0,0 +1,60 @@
...
+// Todo: Expected return Type for .search
+export interface SearchResponse extends Record<string, any> {
+	body: {
+		hits: {
+			hits: {
+				_source: any;
+			}[];
+		};
+	};
+}
```
joneubank: Let's remove speculative types from our exports. We can take it out of this PR, we'll add correct types in future PR.

modules/server/src/searchClient/createElasticSearchClient.ts
```diff
@@ -0,0 +1,88 @@
+import { Client, type ClientOptions } from '@elastic/elasticsearch';
+
+import type { SearchClient } from './types.js';
+
+export type ESClientOptions = ClientOptions & {
+	clientType: 'elasticsearch';
+};
+
+export function createElasticSearchClient(options: ESClientOptions) {
```
joneubank: Should declare return type in signature.

modules/server/src/admin/services/elasticsearch/index.ts
joneubank: Just realized this is in the admin path so it doesn't matter. We should still fix the TS issues, especially when the update is very simple.

joneubank (commented):
Couple minor things, and we need to clean up the initialize function retry logic in the catch block.

modules/server/src/admin/index.ts
justincorrigible: nothing in the "admin" folder matters. may we "resolve" this comment here?

demariadaniel: This `retry` is in the admin folder, only changes to this file on my end were cosmetic / formatting changes. So same question, think we can resolve this.

modules/server/src/admin/services/elasticsearch/index.ts
demariadaniel: Type changes pushed, and after this branch is merged we can do a follow up chore branch to remove `/admin`

modules/server/src/admin/index.ts
joneubank: I missed that this was in /admin. Happy to resolve as is.

joneubank (approved):
Approved! Nice work.

PR #1057, opened 2026-04-20
Feat/ SearchClient types

demariadaniel:
## Issues
- #(issue number)
- #(issue number)

## Readiness Checklist

- [ ] **Self Review**
- [ ] **Formatting**
- [ ] **Local Testing**
- [ ] **Updated Tests**
- [ ] Documentation


modules/graphql-router/src/searchClient/types.ts
```diff
@@ -21,29 +28,228 @@ export type SearchConfigWithClient = Omit<SearchConfig, 'clientType'> & {
...
+export type SearchQueryResponse = Record<string, any> & {
+	body: {
+		hits: {
+			hits: {
+				_source: any;
+			}[];
+		};
+	};
+};
+
```
demariadaniel, line 40: Context from Slack:

This file contains a collection of types used for the SearchClient methods, parameters & return types. These types represent the overlap between ElasticSearch & OpenSearch type definitions.

Using the base `Prettify<ElasticType & OpenType>` definition works well enough in many places.
Some of these essentially end up as `Partial<ElasticType & OpenType>`, don’t know if we need to improve these one way or another
Example `IndicesGetMappingParams`

But then there are places where we have conflicts. I've used the name prefix Shared in these places, such as `SharedIndicesCloseResponseBody`.
This is used where the Type overlap has conflicts, & only includes the mandatory fields. So these could be expanded where needed.
These mostly recreate the OpenSearch internal Types which have more specific definitions; where ElasticSearch uses a lot of `Record<string, any>`, at least in this version of ElasticSearch client.
The sharedTypes are also used for Type arguments in the `elasticSearchClient` file.

.gitignore
```diff
@@ -14,4 +14,6 @@ lerna-debug.log
 
 # include, don't ignore
 !.env.test
-!**/.env.schema
\ No newline at end of file
+!**/.env.schema
+
+./modules/graphql-router/src/admin
```
justincorrigible: this isn't necessary
```suggestion
```

PR #1058, opened 2026-04-23
:broom: Chore / Update SearchClient types

demariadaniel:
## Summary
Follow up to https://github.com/overture-stack/arranger/pull/1021
Adds Type definitions for shared Search Client
Defines types which represent the union of OpenSearch and ElasticSearch types
Merges with branch `3.1`

## Issues
- https://github.com/overture-stack/arranger/pull/1021

## Description of Changes
Adds method, input and output type definitions for ElasticSearch and OpenSearch clients
Creates shared SearchClient types to represent shared properties & manage type conflicts

## Readiness Checklist

- [x] **Self Review**
- [x] **Formatting**
- [x] **Local Testing**
- [ ] **Updated Tests**
- [ ] Documentation


modules/graphql-router/src/searchClient/types.ts
```diff
@@ -1,9 +1,16 @@
 import type { Client as ElasticClient, ApiResponse } from '@elastic/elasticsearch';
```
demariadaniel: Context from Slack:

This file contains a collection of types used for the SearchClient methods, parameters & return types. These types represent the overlap between ElasticSearch & OpenSearch type definitions.

Using the base Prettify<ElasticType & OpenType> definition works well enough in many places.
Some of these essentially end up as Partial<ElasticType & OpenType>, don’t know if we need to improve these one way or another
Example IndicesGetMappingParams

But then there are places where we have conflicts. I've used the name prefix Shared in these places, such as SharedIndicesCloseResponseBody.
This is used where the Type overlap has conflicts, & only includes the mandatory fields. So these could be expanded where needed.
These mostly recreate the OpenSearch internal Types which have more specific definitions; where ElasticSearch uses a lot of Record<string, any>, at least in this version of ElasticSearch client.
The sharedTypes are also used for Type arguments in the elasticSearchClient file.

modules/graphql-router/src/searchClient/types.ts
```diff
@@ -1,9 +1,16 @@
...
-import type { Client as OpenSearchClient } from '@opensearch-project/opensearch';
+import type { TransportRequestOptions as ESTransportRequestOptions } from '@elastic/elasticsearch/lib/Transport';
+import type { Client as OpenSearchClient, API } from '@opensearch-project/opensearch';
+import type { TransportRequestOptions as OSTransportRequestOptions } from '@opensearch-project/opensearch/lib/Transport.js';
 
 import type { ESClientOptions } from './createElasticSearchClient.js';
 import type { OSClientOptions } from './createOpenSearchClient.js';
 
+type Prettify<T> = {
```
justincorrigible: this one can be imported from the `@overture-stack/arranger-types` library

justincorrigible: for reference https://github.com/overture-stack/arranger/blob/3.1/modules/types/src/tools/typeFns.ts#L1

demariadaniel: ~~Good to know, however I don't think `arranger-types/tools` is being exported properly, see:
https://github.com/overture-stack/arranger/blob/3.1/modules/types/src/index.ts
(no exports from `/tools`)~~

~~I'm unable to import Prettify as expected in the searchClient file~~

demariadaniel: Resolved here: [Use overture-stack/types import](https://github.com/overture-stack/arranger/pull/1058/commits/c746d118bb0e95914a313f341e5128e3b4fa469a)

yarn.lock
justincorrigible, line 1: this file doesn't belong in this repo.
please remove it

demariadaniel: Thanks for catching that, removed

joneubank (commented):
I am leaving a comment instead of approve/request because I am not certain of the correct advice to give here. These type changes follow the exploratory pattern we were looking at together, but I have some doubts about using them as the published types.

My concern is that we are focusing on ES7 and OS client defined types, while we know we also want to add in ES9 client in the future. Basing our types on these clients may not be sustainable in the future as we support for additional client types. For that reason, I think we may want to define the types for ourselves in a way that is compatible with these clients, but not using their type definitions directly.

I do like the `Pick<Shared, 'index'>` pattern to limit the scope of which variables we are supporting. That works well to expand our supported scope in the future. I trust that you went through the code base and found all the required params - is this list of picked params is the minimum set we need?

modules/graphql-router/src/searchClient/createElasticSearchClient.ts
```diff
@@ -1,85 +1,133 @@
...
+			const output = await elasticSearchClient.index<SearchClientWriteResponseBody>(input, options);
 			return output;
 		},
-		delete: async (input: any, options?: any) => {
-			const output = await elasticSearchClient.delete(input, options);
+		search: async (input, options) => {
+			let _source = input._source;
+			if (typeof _source === 'boolean') {
+				_source = _source.toString();
+			}
```
joneubank, line 124: 👍

modules/graphql-router/src/searchClient/types.ts
```diff
@@ -16,41 +11,85 @@ export type SearchConfig = {
...
+export type SearchClientIndicesExistsParams = { index: string | string[] };
+export type SearchClientIndicesOpenParams = { index: string | string[] };
+export type SearchClientIndicesPutMappingsParams = { index: string; body: Record<string, any> };
+export type SearchClientIndicesPutSettingsParams = { body: Record<string, any> };
+export type SearchClientBulkParams = { body: Record<string, any>[] };
+export type SearchClientCreateParams = { id: string; index: string; body: Record<string, any> };
+export type SearchClientDeleteParams = { index: string; id: string };
+export type SearchClientDeleteByQueryParams = { index: string; body: Record<string, any> };
+export type SearchClientIndexParams = { index: string; body: Record<string, any> };
+export type SearchClientUpdateParams = { id: string; index: string; body: Record<string, any> };
```
joneubank, line 47: The response types do not have the `SearchClient` prefix, for example `IndicesDeleteResponse`. We should make our type naming consistent.

I don't think a type `BulkParams` is clear enough, so removing the prefix seems unsatisfactory. The alternative is to add `SearchClient` to our responses, so `BulkResponse` becomes `SearchClientBulkResponse`.

That would make all the type names long, but we won't need to refer to the types directly frequently so I think it makes more sense to have types that are clear.

TLDR: Can we add `SearchClient` prefix to the response types?

joneubank (changes requested):
This looks really good, and I was about to approve, but I think we want to add a bit more type name consistency. See comment about renaming SearchClient function repsonse types. That'll be the last change needed that I can see.

modules/graphql-router/src/searchClient/types.ts
demariadaniel: Yes good call!

demariadaniel: @joneubank https://github.com/overture-stack/arranger/pull/1058/commits/ec1317c0013b3b273716780497971e2256754cfd

joneubank (approved):
Thanks for this!

PR #1061, opened 2026-05-11
Add network search to a catalog through network config

joneubank:
## Summary

Update the `search-server` so that a `network` aggregation search schema is added to a catalog based on networkConfiguration in the configuration files.

Additionally, network aggregation search can now include the local arranger catalog in the network aggregation search, in addition to the remote arranger servers in the network configuration.

Some updates to types have been contributed in places where the files were still `.js` or the types were very loosely defined. This is an ongoing process of converting everything to a strictly typed TS application - many TODOs were left indicating places where further narrowing the types is beyond the scope of this PR. Example: [graphql-router/src/mapping/addMappingsToTypes.ts`](https://github.com/overture-stack/arranger/pull/1061/changes#diff-18ede485a14e8beddcb0d13613db4a0ffcbb8a4c0f14f435f52f438ba80917c2) was converted from .js to .ts

Some select places have renamed `esClient` to `searchClient`. This has been done in a small number of files, only in newly added files or in functions that are tightly coupled with a specific purpose so that changes to function signatures are not spread far through the code base. Example is [graphql-router/src/mapping/utils/fetchMapping.ts](https://github.com/overture-stack/arranger/pull/1061/changes#diff-fe30499d89f360481d683e6f05dff1d39870c6ff8dc94d43f7fc4a922e9c4e24).

### search-server
- configs
  - network config now is an object with two properties: `remoteNodes` and `localNode`
    - `remoteNodes` contains the details for how to connect to remote arranger servers for use in aggregate search.
    - `localNode` if provided, the network aggregation search will also include the local catalog in the network aggregation results. A display name is required to describe the local node in the network nodes list.
      ```json
      // Network Search Nodes
      "network": {
      	"remoteNodes": [
      		{
      			"graphqlUrl": "http://localhost:5051/graphql",
      			"documentType": "file",
      			"displayName": "Local Port 5051"
      		}
      	],
      	"localNode": { "displayName": "Local Port 5050" }
      },
      ```
  - Remove config feature flag for `ENABLE_NETWORK_AGGREGATION`. This is always available, and will be added to the GQL Schema based on the NetworkConfig provided.

### graphql-router
- If network config is provided, now includes `network` resolver in GQL schema.
- `network` resolver can be given an array of local and remote nodes
  - An arranger server will successfully boot in the case where there are connection errors to the remote nodes on startup.
  - Each local node requires a corresponding localCatalog which includes the resolvers for that catalog. Configs only allow using a single local catalog at the moment, but the code within the resolver is prepared for a multi-catalog setup where multiple local catalogs can be included in the aggregate search.
  - The `network.nodes` resolver now includes information on the nodes that failed to query their schemas on server startup. These nodes are listed in error state and they are not queried in the network aggregate search results.
- Added `mappingToAggregationFields` function that will convert the properties in the search index mapping into array of AggregationField objects. This is similar to `mappingToAggsType`, but does not convert the data into strings for the GQL Schema, it leaves them as JS Objects so that the data can be used within application logic.
- Added `partitionArray` function that will take an array and split it into two arrays based on a predicate function. This is used in several places, for example to separate network nodes that can be connected to from those that fail to connect, and to separate aggregations that are supported from those that aren't.

### types
- network search config was restructured to have both `"remoteNodes` and one `localNode` representing the catalog that the network search is in.
- GetServerSideFilterFn - generic `Context` type is required so that we can define additional properties that are added to the context passed to this filter. The context must extend `ArrangerBaseContext` which always includes the `esClient: SearchClient` property.

## Future Work
Need to improve documentation on this feature:
- for administrators that want to setup network aggregation search
- for end users that are interacting with the GQL Schema to understand what the `network` property does
- for developers to navigate through the code for network resolvers

## Readiness Checklist

- [x] **Self Review**
- [x] **Formatting**
- [x] **Local Testing**
- [ ] **Updated Tests**
- [ ] Documentation


modules/graphql-router/src/types.ts
```diff
@@ -0,0 +1,38 @@
...
+	esClient: SearchClient;
+};
+export type GraphQLEndpointMiddleware<TSource = any, TContext = any, TArgs = any> =
+	| IMiddleware<TSource, TContext, TArgs>
+	| IMiddlewareGenerator<TSource, TContext, TArgs>;
+
+export type GraphQLEndpointOptions<Context extends ArrangerBaseContext> = {
+	context?: Context | ((req: Request, res: Response, connection: any) => Context);
+	middleware?: GraphQLEndpointMiddleware[];
+} & Record<string, unknown>;
```
joneubank, line 38: Its in `options.context` that a service can provide additional properties to the `Context` type being used in the arranger resolvers. Its important that if the Server Side Filter function requires some additional properties in the Context that this context property be provided those extra properties.

This is not new behaviour, but hopefully these types move us one step closer to clarifying how this works and making sure that users of the graphql-router have these behaviours enforced through the type system. The types may need some additional work to make this enforcement consistent (for example, context being optional should only be allowed if the default context is used).

modules/graphql-router/src/mapping/resolveAggregations.ts
```diff
@@ -36,35 +37,54 @@ export type NumericAggregations = { stats: Stats } & CommonAggregationProperties
...
 
+/**
+ * Replaces all `.` symbols in the keys of an Aggregation Map with `__`.
+ * GraphQL fields cannot use the `.` symbol, they may only be alphanumeric or underscores.
+ *
+ * For example, `donor.age` will become `donor__age`.
+ */
 const toGraphqlField = (acc: AllAggregationsMap, [a, b]: [string, CommonAggregationProperties]) => ({
 	...acc,
 	[a.replace(/\./g, '__')]: b,
```
justincorrigible, line 61: this could also reuse `convertNameForGraphql`

modules/graphql-router/src/mapping/utils/convertNameForGraphql.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Makes a field name usable in a GraphQL Schema.
+ *
+ * This will replace all `.` characters with `__`, which is intended to be used to indicate
+ * a nested relationship since GraphQL schemas cannot use `.` characters.
+ */
+const convertNameForGraphql = (name: string): string => name.split('.').join('__');
```
justincorrigible, line 7: ... and this could be more declaratively expressed as `.replace(/\./g, '__')`

modules/graphql-router/src/utils/partitionArray.ts
```diff
@@ -0,0 +1,16 @@
+const partitionArray = <A, B>(
```
justincorrigible, line 1: just curious what this function is used for...
`naming things is hard` lol

justincorrigible (approved):
this is soooo ready to be Squashed and Merged! :shipit: 
great work, @joneubank

thanks! 🙏

justincorrigible:
merged it myself to continue with the integration work

modules/graphql-router/src/utils/partitionArray.ts
joneubank: Naming this was so hard that I used the internet to find what the common name for this behaviour is. I'll add meaningful docs to describe it.

tldr: elements of one array separated into 2 arrays based on weather each element passes a predicate. it's like filter but doesn't discard the elements that return false form the provided function, it keeps them in a the second array

PR #1066, opened 2026-05-20
Customize Headers in network search remote requests

joneubank:
## Summary

Introduces a mechanism to customize the network requests made to remote nodes during network search. With this initial setup, we have added the ability to add custom headers to outgoing requests to each remote node sent during network aggregation searches.

When the arranger `graphql-router` is created, an optional function can be provided as part of the network config. This CustomizeRemoteRequestFn will receive the Arranger context (including incoming request headers), as well as the config for a remote node. This function can then determine whether to inject headers into the outgoing request.

The `search-server` has used this to enable a new network configuration: `passthroughHeaders`. If the network configuration file includes a `passthroughHeaders` property (`string[]`), then headers of the given name will be taken from the incoming search query and added to the outgoing request to every remote node.

This functionality is requested to enable transporting Authorization credentials from the requesting user through to the other nodes in the network which have implemented a custom authorization scheme in front of the arranger servers.

### modules/graphql-router
- Adds to the ArrangerBaseContext information about the incoming request
  - headers are made available with this change. Future additions could expand this.
- Added to graphqlRouter the `customizeRemoteRequest` optional config property.
  - This takes a function that can add properties to each outgoing request to a remote node in a network query.
  - The output of this function can supply headers to add to the outgoing request
  - This function is called once for each remote node, so that custom logic can be provided for modifying each node's request.

### apps/search-server
- `network` config now accepts a `remoteRequests` property which can specify custom properties to add to remote requests:
  - `headers` will add to outgoing remote network requests any headers from the incoming requests with names that match the values in this string array
- Each `remoteNode` config object can be given a `requests` property that matches the `remoteRequests` customization object. These properties apply to only the one node they are set for, and will override values set in the top network level `remoteRequests` config.
- The custom handling of these configs is in the `normalize.ts` file, where the `customizeRemoteRequests` function is defined. These properties are not used by the `graphql-router`
- Updated the example config files to demonstrate the usage of these new properties

## Readiness Checklist

- [x] **Self Review**
- [x] **Formatting**
- [x] **Local Testing**
- [ ] **Updated Tests**
- [x] Documentation


.dev/roadmap.md
```diff
@@ -17,7 +17,7 @@ The scope is wider than just swapping a client library. It includes the `SearchC
 
 The `SearchClient` abstraction already exists as the right boundary — the migration should align the types and default configuration to OpenSearch while preserving compatibility for ES users.
 
-**What's already done:** The integration test suite (`integration-tests/server`) already supports both engines via a `SEARCH_ENGINE` env var (defaults to `'elasticsearch'`). `buildSearchClient` accepts a `client` type parameter mapped to `SupportedSearchClients`. The architecture is ready; the missing pieces are the OpenSearch client dependency and a running OpenSearch instance in CI.
+**What's already done:** The integration test suite (`integration-tests/server`) already supports both engines via a `SEARCH_ENGINE` env var (defaults to `'elasticsearch'`). `buildSearchClient` accepts a `client` type parameter mapped to `SupportedClientTypes`. The architecture is ready; the missing pieces are the OpenSearch client dependency and a running OpenSearch instance in CI.
```
joneubank, line 20: There is no type named `SupportedSearchClients`, I believe this is leftover before a name change. I have replaced all instances if this with `SupportedClientTypes` and this SHOULD work as intended. If needed, suggest the desired name and we can make this consistent before merging (or resolve in a separate PR).

apps/search-server/src/configs/fromFiles/normalize.ts
```diff
@@ -22,6 +95,22 @@ const normalize = (fileDataJSON: any) => {
...
+ * Normalizes raw config file JSON for use by the search server. Takes the raw JSON read from any provided config files
+ * and formats them as a valid arranger ConfigObject
+ *
+ * - Normalize Network Config:
+ *   - Create customRemoteRequestFn based on `passthroughHeaders` value
+ * - Normalize Table Config:
+ *   - Ensures every entry in `table.defaultSorting` has `desc` explicitly set,
+ *     defaulting to `false` when the field is absent.
+ */
+const normalize = (fileDataJSON: any) => {
```
joneubank, line 116: I don't think this is the best function name, but I'm trying to refactor only lightly as I add features.

apps/search-server/src/configs/types/index.ts
```diff
@@ -21,6 +23,11 @@ export type BaseServerConfigs = {
 	[serverConfigProperties.SERVER_PORT]: number;
 } & RuntimeFeatureFlagConfigs;
 
+export type ExternalNetworkConfig = Partial<{
+	[serverNetworkConfigExtendedProperties.PASSTHROUGH_HEADERS]: string[];
+}> &
+	NetworkConfig<ArrangerBaseContext>;
```
joneubank, line 61: Config file types and `graphql-router` types are diverging, which is to be expected. It would be worthwhile doing a dedicated task on separating the types for these two contexts, and to provide validation schemas (zod) for the config files.

.dev/roadmap.md
justincorrigible: I strongly prefer the more descriptive name I had applied (this needs to be read as a configuration option, not a code compliance thing), but we can redo that change once this is merged (along with a couple other name changes decided elsewhere), so your PR's scope doesn't grow. 👍

apps/search-server/configTemplates/configs.json.schema
```diff
@@ -104,6 +104,7 @@
 
 	// Network Search Nodes
 	"network": {
+		"passthroughHeaders": ["Authorization"],
```
justincorrigible: I see what you're doing here, and could use some documentation (even if only as a comment in the line above)

apps/search-server/configTemplates/network.json
```diff
@@ -1,3 +1,13 @@
 {
-	"network": []
+	"network": {
```
justincorrigible, line 2: 🎉

apps/search-server/src/configs/fromFiles/normalize.ts
```diff
@@ -1,14 +1,87 @@
...
 import { configRootProperties, tableProperties } from '@overture-stack/arranger-types/configs/constants';
 import { merge } from 'lodash-es';
 
+import { serverNetworkConfigExtendedProperties } from '#configs/types/constants.js';
+import type { ExternalNetworkConfig } from '#configs/types/index.js';
+
 /**
- * Normalizes raw config file JSON for use by the search server.
+ * This will assume the network config provided has the expected format, it is not fully validated.
+ * TODO: Fully validate config file structure
```
justincorrigible, line 20: nice! added an item to this effect in the roadmap, to do it both her and in the core module, with proper error feedback, and to add the corresponding tests

apps/search-server/src/configs/fromFiles/normalize.ts
```diff
@@ -1,14 +1,87 @@
...
+	return fileDataJSON[configRootProperties.NETWORK_AGGREGATION] as NetworkConfig<ArrangerBaseContext>;
+};
+
+const getNetworkPassthroughHeaders = (networkConfig: ExternalNetworkConfig): string[] => {
+	const passthroughHeaders = networkConfig[serverNetworkConfigExtendedProperties.PASSTHROUGH_HEADERS] ?? [];
+
+	// If passthroughHeaders value is provided, check that the passthroughHeaders value is an array of strings.
+	// We want to error out if the provided file config is not a usable format.
+	if (
+		passthroughHeaders !== undefined &&
```
justincorrigible: this check is probably unnecessary, given the `?? []` above
```suggestion
```

apps/search-server/src/configs/fromFiles/normalize.ts
```diff
@@ -22,6 +95,22 @@ const normalize = (fileDataJSON: any) => {
...
+ * - Normalize Network Config:
+ *   - Create customRemoteRequestFn based on `passthroughHeaders` value
+ * - Normalize Table Config:
+ *   - Ensures every entry in `table.defaultSorting` has `desc` explicitly set,
+ *     defaulting to `false` when the field is absent.
+ */
+const normalize = (fileDataJSON: any) => {
+	normalizeTableConfig(fileDataJSON);
+
+	normalizeNetworkConfig(fileDataJSON);
```
justincorrigible, line 119:
```suggestion
	// TODO: this mutational logic is brittle
	normalizeTableConfig(fileDataJSON);
	normalizeNetworkConfig(fileDataJSON);
```

apps/search-server/configTemplates/network.json
```diff
@@ -1,3 +1,13 @@
...
+		"passthroughHeaders": [""],
```
justincorrigible: this array should be empty. anyone using the template suggestion as is, will find the logic attempts to pass through a header with an empty name on every request.
```suggestion
		"passthroughHeaders": [],
```

integration-tests/import/package.json
```diff
@@ -9,7 +9,6 @@
 	},
 	"private": "true",
 	"scripts": {
-		"test": "NODE_OPTIONS=\"$NODE_OPTIONS --experimental-vm-modules\" jest --config=jest.config.ts"
-	},
-	"type": "module"
+		"test": "jest --config=jest.config.ts"
```
justincorrigible: did you test this? I think it breaks compatibility on some imports, hence why I added it. seems beyond the scope of this PR tbh

modules/types/src/configs/index.ts
```diff
@@ -147,9 +174,7 @@ export type ConfigsObject<Context extends object> = {
 		getServerSideFilter: GetServerSideFilterFn<Context>;
 		[configOptionalProperties.CATALOG_ID]: string;
 		[configOptionalProperties.ES_HOST]: string;
-		[configOptionalProperties.GRAPHQL_MAX_ALIASES]: number;
-		[configOptionalProperties.GRAPHQL_MAX_DEPTH]: number;
```
justincorrigible, line 151: please readd these. looks like merge conflict mistake

justincorrigible: may I ask to confirm the related logic wasn't also removed?

modules/graphql-router/src/graphqlRoutes.ts
```diff
@@ -246,13 +247,25 @@ export const createEndpoint = async <Context extends ArrangerBaseContext>({
...
+				const headers = new Headers();
+				for (const [key, value] of Object.entries(req.headers)) {
+					if (value !== undefined) {
+						const valueAsString = Array.isArray(value) ? value.join(', ') : value;
+						headers.set(key, valueAsString);
+					}
+				}
+				const request: RequestContextProps = { headers };
+
+				// Add to context based on external parameters
```
justincorrigible, line 260: while I of course may be misreading the passthrough logic overall, it looks like this applies the same headers to every remote node in the network. i.e. there's no per-node control. 

for instance If remote nodes have different trust levels, e.g. one is internal, another is a third-party node, forwarding `Authorization` to all of them may leak credentials. 

... either the docs/config schema should explicitly warn that this affects all nodes equally, or better yet, this should be an opt-in kind of deal so we don't go against `OWASP A01: Broken Access Control` recommendations.

modules/graphql-router/src/network/gql.ts
```diff
@@ -22,7 +25,11 @@ export const fetchGql = ({
 	const axiosOptions: AxiosRequestConfig = {
 		url,
 		method: 'POST',
-		headers: { 'Content-Type': 'application/json', 'X-REQUEST-TYPE': 'GraphQL' },
+		headers: {
+			'Content-Type': 'application/json',
+			'X-REQUEST-TYPE': 'GraphQL',
+			...(customRequestProps?.headers || {}),
```
justincorrigible, line 31: hm yeah, here it is... the passthrough thing is non-discriminate and could leak credentials. thinking the configuration should state explicitly which nodes to send which headers to, somehow (granted, that sounds complex)

modules/graphql-router/src/network/setup/fields.ts
```diff
@@ -1,14 +1,13 @@
...
-import type { AggregationField, NetworkRemoteNode } from '../types/setup.js';
 import partitionArray from '../../utils/partitionArray.js';
+import type { AggregationField, NetworkRemoteNode } from '../types/setup.js';
 
 export type SupportedAggregationField = AggregationField & { type: (typeof SUPPORTED_AGGREGATIONS)['Aggregations'] };
 
-const isFieldAggregationSupported = (fieldObject: AggregationField): fieldObject is SupportedAggregationField => {
+export const isFieldAggregationSupported = (
+	fieldObject: AggregationField,
+): fieldObject is SupportedAggregationField => {
```
justincorrigible, line 10: i really need to wrap my head around this "is" thing. I forget it exists and can be so useful

modules/graphql-router/src/network/resolvers/aggregations.ts
```diff
@@ -76,7 +87,8 @@ export const aggregationPipeline = async <Context extends ArrangerBaseContext>(p
 
 			if (response.success) {
 				const documentName = config.documentType;
-				const responseData = response.data[documentName];
+				// TODO: Response content is not validated, we expect the return structure based on the GraphQL query we requested
```
justincorrigible, line 90: this sounds easier than it may actually be, from a model-agnostic perspective, considering an aggregator may not be fully aware of the nodes' model and harmonization (or at least some mapping) may be necessary

modules/graphql-router/src/config/constants.ts
```diff
@@ -39,7 +35,6 @@ const fallbackCatalogConfigs = {
 		[facetsProperties.AGGS]: [],
 	},
 	[configRootProperties.MATCHBOX]: [],
-	[configRootProperties.NETWORK_AGGREGATION]: { [configArrangerNetworkProperties.REMOTE_NODES]: [] },
```
justincorrigible, line 42: surprised at first, but i guess it makes sense in the decoupling of "core" from "aggregator" types 👍

justincorrigible (changes requested):
Great work overall (as usual)! 👏 
two items of "concern" I'd like to surface here:
- the OWASP a01 thing I pointed out, passing headers to all nodes.
- the possible regression on GraphQL security functionality (thanks for putting up with the Git shenanigans)

apps/search-server/configTemplates/configs.json.schema
joneubank: Yeup, I owe better (read: some) documentation for this feature

integration-tests/import/package.json
joneubank: Another merge conflict issue, will restore the script from main.

modules/graphql-router/src/network/setup/fields.ts
joneubank: I actually strongly discourage using type predicates now. TS will infer the predicate logic if the function returns boolean and narrows the type of the input variable. If TS can't calculate that the type is actually narrowed correctly then manually declaring the type predicate is about as risky as using any type assertion.

modules/graphql-router/src/config/constants.ts
joneubank: This is because I updated the config type to make the remoteNodes property optional so there is no need to set an empty list by default.

apps/search-server/configTemplates/configs.json.schema
```diff
@@ -41,8 +41,8 @@
 			"rangeStep": 1,
 			"unit": null,
 			"type": "object" // from the ES mapping
-		},
-		...
+		}
+		// ...
```
justincorrigible, line 45: this is a conceptual schema, not a template for reuse
unsure how this change helps here

justincorrigible (approved):
changes look good! 
`squash and merge` away!
:shipit:

PR #1074, opened 2026-06-09
✨ Feat - Charts - Add optional suppression handling for Bar chart fields with a zero value

anncatton:
## Summary

Adds handling for zero values in the Bar Chart component. Required for the UI implementation portion of https://github.com/OHCRN/platform/issues/1917. **This is a custom use case for OHCRN**, as we need to suppress any Aggregation values below a configured cell size, so we have added extra handling in our backend arch that will return a zero value instead for any field that does not meet this criteria.

**Note**: Includes https://github.com/overture-stack/arranger/pull/1075

## Issues
[OHCRN Platform #1917](https://github.com/OHCRN/platform/issues/1917)

### Charts Module
Adds handling for fields with a zero value in the Bar Chart:
- maps through the data array passed to the Bar Chart, and for any value that is `0`, adds the `SUPPRESSION_INCREMENT_VALUE`. This forces the chart to display a shortened bar and allows a tooltip to be displayed on hover.
- **any non-zero values will be rendered normally**. Since other use cases for Aggregations should not be providing zero values for any field, this behaviour should be ignored for those cases.
- adds an optional boolean prop `suppressed` to the Tooltip props. This will trigger custom text `Too few` for the tooltip display instead of the numeric value. This value is set to `false` in the Sunburst tooltip props, so should not affect that component.

## Readiness Checklist

- [x] **Self Review**
- [x] **Formatting**
- [x] **Local Testing**
- [ ] **Updated Tests**
- [ ] Documentation

### DIFF
--- a/modules/charts/src/components/charts/Bar/View.tsx
+++ b/modules/charts/src/components/charts/Bar/View.tsx
@@ -6,6 +6,8 @@ import { useColorMap } from '#hooks/useColorMap';
 import { BarChartProps } from './BarChart';
 import { arrangerToNivoBarChart } from './nivo/config';
 
+const SUPPRESSION_INCREMENT_VALUE = 0.2;
+
 interface BarChartViewProps {
 	data: any;
 	handlers: BarChartProps['handlers'];
@@ -13,6 +15,12 @@ interface BarChartViewProps {
 	maxBars: BarChartProps['maxBars'];
 	colorMapRef: React.RefObject<Map<string, string>>;
 }
+
+type BarData = {
+	key: string;
+	value: number;
+	label: string;
+};
 
 /**
  * Creates a chart color map for consistent colors across charts
@@ -61,16 +69,24 @@ export const BarChartView = ({ data, handlers, theme, maxBars, colorMapRef, fiel
 		[theme, colorMap, handlers],
 	);
 
+	// Values that are 0 will be incremented by the SUPPRESSION_INCREMENT_VALUE, in order for the Bar Chart to render a bar
+	// with a Tooltip, and will add custom text to the Tooltip.
+	// Any non-zero values will be rendered normally.
+	const dataWithSuppressedValues = data.map((d: BarData) => {
+		const suppressed = d.value === 0;
+		return { ...d, suppressed, value: suppressed ? d.value + SUPPRESSION_INCREMENT_VALUE : d.value };
+	});
+
 	// 1) custom sort order or ascending (from axis)
 	// 2) limit by maxRows
 	// 3) reverse order for display
 	const barData = theme.sortByKey
 		? theme.sortByKey
-				.map((label) => data.find((bar) => bar.key === label))
+				.map((label) => dataWithSuppressedValues.find((bar: BarData) => bar.key === label))
 				.filter(Boolean)
 				.slice(0, maxBars)
-		: data
-				.toSorted((a, b) => b.value - a.value)
+		: dataWithSuppressedValues
+				.toSorted((a: BarData, b: BarData) => b.value - a.value)
 				.slice(0, maxBars)
 				.reverse();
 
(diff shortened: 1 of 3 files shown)


modules/charts/src/components/charts/Bar/View.tsx
```diff
@@ -6,6 +6,8 @@ import { useColorMap } from '#hooks/useColorMap';
 import { BarChartProps } from './BarChart';
 import { arrangerToNivoBarChart } from './nivo/config';
 
+const SUPPRESSION_INCREMENT_VALUE = 0.2;
```
justincorrigible, line 11: adding comment to do this later, not needed in current scope
```suggestion
// TODO: abstract as a customizable value, using this as default
const SUPPRESSION_INCREMENT_VALUE = 0.2;
```

modules/charts/src/components/charts/Tooltip.tsx
```diff
@@ -1,37 +1,43 @@
...
-						<span className="tooltip-data-source">: Donor</span>
-					</span>
-					<span>{value > 1 ? 's' : ''}</span>
+					{suppressed ? (
+						<span>Too few</span>
+					) : (
+						<>
+							<span>{value}</span>
+							<span className="tooltip-data-source-wrapper">
+								<span className="tooltip-data-source">: Donor</span>
```
justincorrigible: "Donor" is not a standard/generic term
this needs to be passed in from your app, while the default could be something like "Record"?
unsure what this tooltip reflects, so I'm open to ideas

anncatton: Updated! 🙏

samrichca (approved):
Looks good to me - this preserves the class names for us to be able to customize "donor"/"record" to "participant"

justincorrigible (approved):

modules/charts/src/components/charts/Bar/View.tsx
```diff
@@ -35,6 +46,64 @@ const colorMapResolver = ({ chartData, savedMap, colors }) => {
 	return colorMap;
 };
 
+/**
+ * Creates a numeric array of even multiples.
+ *
+ * @remarks The start value is expected to be higher than the stop value.
```
justincorrigible, line 52: fun, hadn't seen this one. will need to do some reading on it 👍

modules/charts/src/components/charts/Bar/View.tsx
```diff
@@ -35,6 +46,64 @@ const colorMapResolver = ({ chartData, savedMap, colors }) => {
...
+ * @param start starting value of array
+ * @param stop final value of the array
+ * @param step value for the multiple of each element in the array
+ * @returns an array of numbers
+ */
+const range = (start: number, stop: number, step: number): number[] => {
+	if (start >= stop || stop - start < step) {
+		return [start, stop];
+	}
+	return Array.from({ length: Math.ceil((stop - start) / step) }, (_, i) => start + i * step);
```
justincorrigible: missed this single letter variable name in the previous review
```suggestion
	return Array.from({ length: Math.ceil((stop - start) / step) }, (_, index) => start + index * step);
```

modules/charts/src/components/charts/Tooltip.tsx
```diff
@@ -1,37 +1,43 @@
...
-					<span>{value > 1 ? 's' : ''}</span>
+					{suppressed ? (
+						<span>Too few</span>
+					) : (
+						<>
+							<span>{value}</span>
+							<span className="tooltip-data-source-wrapper">
+								<span className="tooltip-data-source">: Record</span>
+							</span>
+							<span>{value > 1 ? 's' : ''}</span>
```
justincorrigible, line 39: `0 Record`? I'd use `pluralize` [like so (an npm package already in the monorepo, at Components)](https://github.com/overture-stack/arranger/blob/main/modules/components/src/Table/CountDisplay/CountDisplay.tsx#L39) and call it a day

anncatton: Does not work when the label is applied via css `content`; added `TODO`

PR #1076, opened 2026-06-15
Add NetworkAggregation mode to Charts

joneubank:
## Summary

Adds a mechanism to have the Chart components perform a network aggregation search and display that network aggregation data.

Each of the existing chart types (`BarChart` and `SunburstChart`) have a new property `isNetworkAggregation` that if set to `true` will switch their functionality to use the network aggregation search query.

A new chart type `<NetworkNodesChart />` is also added to query specifically the hits on each node in the Arranger search network. This chart is a BarChart with data handling dedicated to the total hits on each node in the network.

Lastly, a mechanism was added to the graphql-router to filter the network query to only include a subset of the available nodes. This mechanism is a `nodesFilter` value that is provided to the `network` type in the GQL query. It filters based on the included `nodeId`s. To support this change, the configuration of the arranger server now requires each node in the network to have a `nodeId` and the arranger server application was updated to read this value in its configs.

## Readiness Checklist

- [x] **Self Review**
- [x] **Formatting**
- [x] **Local Testing**
- [ ] **Updated Tests**
- [x] Documentation


joneubank:
Branch was originally made off of a previous PR and so it has a long list of old commits that I haven't cleared out. New changes begin with [58f714a](https://github.com/overture-stack/arranger/pull/1076/commits/58f714afef39fba2c553e5f296ee60d059af49b8) .

modules/charts/src/components/Provider/Provider.tsx
```diff
@@ -54,75 +58,123 @@ const createChartDataMap = (data): GQLDataMap | null => {
...
 	}, []);
 
-	const chartContext: ChartContextType = {
+	const chartContext: ChartContext = {
 		registerChart,
 		deregisterChart,
 		getChartData,
+		getNetworkChartData,
+		getNetworkNodesData,
+		requireNetworkSearch,
```
joneubank, line 177: Use these new functions exported from the provider to access network search data, or require the network search to run.

If a chart uses network search, this can be specified as an additional property in the existing `registerChart` function.

modules/charts/src/components/charts/Bar/BarChart.tsx
```diff
@@ -22,6 +23,8 @@ interface SupportedNivo {
 
 export interface BarChartProps {
 	fieldName: string;
+	isNetworkAggregation?: boolean;
```
joneubank, line 26: This new property is the key addition, this will be set to `true` to indicate that the BarChart should the data from the network aggregation search instead of the default local catalog aggregation data.

modules/charts/src/components/charts/Bar/BarChart.tsx
```diff
@@ -57,37 +63,37 @@ export const BarChart = ({
...
-		[fieldName, variables, extendedMapping],
-	);
-
+	// TODO: Validate for network fields and determine gqlType (no extended mapping guarantee atm for network aggregations)
+	const validationResult = isNetworkAggregation
+		? success({ fieldName, gqlTypename: networkAggregationType, isNetworkAggregation })
+		: useMemo(
+				() => validateQueryProps({ fieldName, variables, extendedMapping }),
+				[fieldName, variables, extendedMapping],
+			);
```
justincorrigible: ### MUST for this PR.
think this will error out as a Rules of Hooks validation. Hooks should be called unconditionally. 

odd, I thought we had React ESLint to help flag these out in this repo. TODO for myself to make sure they're working.

modules/charts/src/components/charts/Sunburst/SunburstChart.tsx
```diff
@@ -44,26 +50,31 @@ export const SunburstChart = ({
...
-	const validationResult = useMemo(
-		() => validateQueryProps({ fieldName, extendedMapping }),
-		[fieldName, extendedMapping],
-	);
+	const validationResult = isNetworkAggregation
+		? success({ fieldName, gqlTypename: aggregationsTypenames.Aggregations, isNetworkAggregation: true })
+		: useMemo(
+				() => validateQueryProps({ fieldName, variables: {}, extendedMapping }),
+				[fieldName, extendedMapping],
+			);
```
justincorrigible: ### MUST for this PR.
same note as in barchart about Rules of Hooks conditional call error.

apps/search-server/src/configs/fromFiles/normalize.ts
```diff
@@ -49,13 +49,11 @@ const normalizeNetworkConfig = (configFilesJson: any) => {
...
-				// If there are multiple matches, we can't differentiate, we will use the first match from the configs
-				const matchingConfig = remoteNodeExtendedConfigs.find(
-					(node) =>
-						node.graphqlUrl === params.remoteNode.graphqlUrl &&
-						node.displayName === params.remoteNode.displayName,
-				);
+				// Match by nodeId.
+				// If there are nodes with duplicate nodeId then we can't differentiate, we will use the first match from the configs
+				const matchingConfig = remoteNodeExtendedConfigs.find((node) => {
+					return node.nodeId === params.remoteNode.nodeId;
```
justincorrigible, line 61: ### MUST for this PR.
if `nodeId` is ever absent on both sides, then `undefined === undefined` => true, so anyone with multiple remote nodes and custom per-node headers but no `nodeId` in their config will have the first node's headers applied to all nodes.

I think this is a regression from your work in #1066, and a quick fix would be to make `nodeId` required in RemoteNodeConfig (which makes it easier to detect on boot), or to keep the old `graphqlUrl + displayName` match as fallback when `nodeId` is absent. 

your call 👍

modules/charts/tsconfig.app.json
```diff
@@ -22,6 +22,7 @@
 		"noUnusedParameters": true,
 		"noFallthroughCasesInSwitch": true,
 		"noUncheckedSideEffectImports": true,
+		"noImplicitAny": false,
```
justincorrigible: ### COULD for this PR. Let's talk
wouldn't it be better if we forced adding `any` rather than allowing implicit ones that make things harder to refactor or debug later?
I have a feeling this is for the nivo configs, which could be handled with e.g. `Partial<BarSvgProps>` or `@nivo/bar`

modules/charts/src/components/charts/NetworkNodes/NetworkNodesChart.tsx
```diff
@@ -0,0 +1,95 @@
...
+ * @returns JSX element with complete bar chart or null if field validation fails
+ */
+export const NetworkNodesChart = ({
+	disableTopBarsCount = false,
+	handlers,
+	maxBars = Infinity,
+	theme,
+}: NetworkNodesChartProps) => {
+	// ensure maxBars is provided
+	if (!maxBars) {
```
justincorrigible: ### COULD for this PR. Let's talk
this defaults to infinity, so it will (almost?) never trigger. this condition/error handler can be removed and teh default documented in JSDoc instead

modules/charts/src/components/Provider/useQueryFieldNames.tsx
```diff
@@ -1,70 +1,106 @@
...
-			}
+	const removeQuery = useCallback((fieldName: string, isNetworkAggregation: boolean) => {
+		// TODO: Need to keep unique records for every registration and only remove them one at a time. The gql generation can de-duplicate the fields.
+		if (isNetworkAggregation) {
+			setNetworkAggregationQueries((prev) => {
+				// There is no current reason to check if this chart has been registered before, we can just set it again
+				// Need to rebuild the Map to ensure we get the state update
+				const newValuesMap = new Map(prev);
+				newValuesMap.delete(fieldName);
+				logger.debug(`Registered Chart for Network Aggregation on field '${fieldName}'`);
```
justincorrigible: ### COULD for this PR. maybe I'm misreading it
looks like this log comes up in the `removeQuery` so maybe it should say `Deregistered` instead?

modules/charts/tsconfig.app.json
```diff
@@ -1,9 +1,9 @@
 {
 	"compilerOptions": {
 		"tsBuildInfoFile": "./node_modules/.tmp/tsconfig.app.tsbuildinfo",
-		"target": "ES2020",
+		"target": "ES2023",
```
justincorrigible, line 4: ### TODO for another PR.
I think this was to enable `Array.toSorted`? 
it's raising the minimum required runtime, which I'm fine with but could use being documented both in changelog and docs.

modules/types/src/configs/index.ts
```diff
@@ -113,6 +113,7 @@ export type TableConfigs = {
 
 export type BaseNodeConfig = {
 	[baseNodeProperties.DISPLAY_NAME]: string;
+	[baseNodeProperties.NODE_ID]?: string;
```
justincorrigible, line 116: ### TODO for another PR. perhaps
While making this optional is honest to implementation, but basically required for NetworkNodesChart and the filtering to work, it may make sense to at least give a warning on boot that a missing nodeId may break things or something like that

modules/graphql-router/src/network/typeDefs/aggregations.ts
```diff
@@ -95,9 +96,10 @@ export const createNetworkAggregationTypeDefs = (networkFieldTypes: SupportedAgg
 			network: {
 				type: networkType,
 				args: {
-					filters: { type: GraphQLJSONObject },
+					filters: { type: GraphQLJSON },
```
justincorrigible, line 99: ### NO Actionable here.

lol quiet bug fix? nice catch!
curious: what's the difference in practice?

modules/charts/src/components/charts/Bar/BarChart.tsx
```diff
@@ -46,6 +50,8 @@ export const BarChart = ({
 	disableTopBarsCount = false,
 	fieldName,
 	handlers,
+	isNetworkAggregation = false,
+	networkAggregationType = aggregationsTypenames.Aggregations,
```
justincorrigible, line 54: ### NO Actionable. just design debate
these two could be unified into a single one, where since networkAggregationType is already optional, whenever present it's as if the `isNetworkAggregation == true`

modules/charts/src/components/charts/Bar/BarChart.tsx
```diff
@@ -57,37 +63,37 @@ export const BarChart = ({
...
-		const result = validateQueryProps({
-			extendedMapping,
-			fieldName,
-			variables,
-		});
-
 		if (!validationResult.success) {
 			logger.log(validationResult.message);
-			return null;
+			return undefined;
```
justincorrigible, line 84: ### COULD for this PR. implementation doubt

I recall returns in a useEffect had a React-specific purpose -> "cleanup", and this may be breaking that convention.
perhaps things have changed, and this actually works as you intend it?

justincorrigible: follow-up: it's still convention `Your setup function may also optionally return a cleanup function`
https://react.dev/reference/react/useEffect#parameters

justincorrigible: hm and I can see now that is what you're trying to do here... so resolving my own comments

modules/charts/src/components/charts/Bar/BarChart.tsx
joneubank: True for the moment. I have ambition to have network search properly support numeric aggregations as well.

modules/graphql-router/src/network/typeDefs/aggregations.ts
joneubank: `GraphQLJSONObject` type does not actually exist ;)

modules/charts/tsconfig.app.json
joneubank: Do we currently have this documented in the docs? Just looked it up and this would put the minimum Node version to 18... but this runs in the browser so its compatible with all modern browsers either way and will likely go through a bundling process.

Yes this was for `.toSorted` to remove the TS error being called out in the `Bar` chart where it was already used, and I reused it in the `NetworkNodes` chart as well.

modules/graphql-router/src/network/resolvers/index.ts
```diff
@@ -92,17 +95,21 @@ export const createResolvers = <Context extends ArrangerBaseContext>(params: {
 				throw new Error(`Provided filter is not a valid sqon: ${result.data}`);
 			}
```
justincorrigible: just found, weeks later: is this logic correct? suggesting reversal to `if (!result.success)`
PS: TODO above may be resolved with latest implementation of SQONbuilder tools in `modules/sqon`


---
EDIT because GH review comments are being stupid

joneubank: Will review but looks like a mistake, ya.

modules/charts/tsconfig.app.json
joneubank: Don't think I meant to commit this. Am reviewing.

apps/search-server/src/configs/fromFiles/normalize.ts
joneubank: Using original match if the provided params have an undefined `nodeId` value.

modules/charts/tsconfig.app.json
joneubank: There's are hundreds of warnings caused by implicit any. It would take significant type repairs to fix this. It looks like it is common in other Arranger modules to have this property set, likely for the same reason.

joneubank: My concern with adding in explicit `any` is that it would indicate we "intend" the type to be `any`, when I do not want it to be that, its just that its been left in an untyped state.

joneubank: I have decided to remove this change. The charts build logs many errors, but it still builds. This is the preferable state while it is untyped: does not explicitly state `any` where types are not provided, and highlights the errors in those cases, without preventing the build.

modules/charts/src/components/charts/NetworkNodes/NetworkNodesChart.tsx
joneubank: Removing this check, was copy pasted from before the default. For this chart, defaulting to Infinity should be safe here since it requires a large number of remote nodes be configured to have a large number of bars.

modules/charts/src/components/Provider/useQueryFieldNames.tsx
justincorrigible:
```suggestion
				logger.debug(`Deregistered Chart for Network Aggregation on field '${fieldName}'`);
```

PR #1077, opened 2026-06-15
✨ Add `execute-query` MCP Tool (overture-stack/admin#177)

mistryrn:
## Summary

Adds `execute-query` tool to MCP Server, which enables making SQON-filtered GraphQL requests to Arranger.

## Issues

- https://github.com/overture-stack/admin/issues/177

### MCP Server
- Added `modules/sqon` (`@overture-stack/sqon`) as a dependency to `mcp-server`
- Updated `mcp-server` Dockerfile build target to copy `modules/sqon` dependencies
- Renamed `ArrangerIntrospectionClient` to `ArrangerClient` to reflect full capabilities (more than just introspection, now allows executing queries)
- Added `executeQuery` to `ArrangerClient`, enabling making `POST` requests containing GQL queries to Arranger
- Added `arranger/queryBuilder.ts` to MCP Server, containing functions for building parameterized GraphQL requests for Arranger catalogue queries, and accompanying unit tests
- Added `arranger/queryValidation.ts` to MCP Server, containing functions for validating the different parts of Arranger GraphQL queries (`sqon`, `hits` fields, `aggregations` fields, `sort` fields), and accompanying unit tests
- Added `arranger/queryResults.ts` to MCP Server, which takes an Arranger GQL hits response and turns it into flat, plain documents that are more easily read by an LLM.
- Registered new `execute-query` tool to MCP Server, which builds, validates, confirms, and executes a SQON-filtered GQL query against one Arranger catalogue

### Integration Tests
- Updated existing MCP Server integration tests to account for new `execute-query` tool
- Added new integration tests for MCP Server `execute-query` tool

### Special Instructions

Before running these changes, you will need to:

1. Rebuild the shared modules:
```bash
# from project root
npm run modules:build
```

2. Install dependencies (to add `modules/sqon` to `mcp-server`):
```bash
# from project root
npm ci
```

## Readiness Checklist

- [x] **Self Review**
  - I have performed a self review of code
  - I have run the application locally and manually tested the feature
  - I have checked all updates to correct typos and misspellings
- [x] **Formatting**
  - Code follows the project style guide
  - Autmated code formatters (ie. Prettier) have been run
- [x] **Local Testing**
  - Successfully built all packages locally
  - Successfully ran all test suites, all unit and integration tests pass
- [x] **Updated Tests**
  - Unit and integration tests have been added that describe the bug that was fixed or the features that were added
- [x] Documentation
  - All new environment variables added to `.env.schema` file and documented in the README
[...]

PR #1078, opened 2026-06-25
⚡️ Add few-shot examples and a "SQON Cheat Sheet" to improve LLM SQON Generation

mistryrn:
## Summary

Enhancements to PR #1077 to improve the likelihood that smaller LLMs such as `gemma-4-e4b` can successfully generate valid SQON for use with the `execute-query` tool. Adds few-shot examples of valid SQON, as well as clear instructions on how to build SQON from a user's query.

## Issues

- https://github.com/overture-stack/admin/issues/177

## Description of Changes

Initial testing of the `execute-query` tool with a small local LLM (`gemma-4-e4b`) revealed that despite having access to the full SQON schema from the `get-sqon-schema` tool, small LLMs still failed to produce valid SQON. An analysis led by a Claude thinking model determined 5 potential root causes for the failures, summarized below:

1. Prior training for the model (based on outdated/inaccurate Arranger docs) conflicts with the schema
2. `get-sqon-schema` returns a validation artifact, not a generation guide
3. The published JSON schema contains dangling `$ref`s
4. The `sqon` input parameter to `execute-query` is `Zod.unknown()`, which provides zero structure at the protocol level
5. Error responses for invalid SQON provide little helpful detail for LLM self-correction

The docs have since been updated (thanks, @justincorrigible !), so this PR focuses on issues 1 and 2 as follows:

1. Provide few-shot examples of valid SQON at every touchpoint the LLM has in the `execute-query`/SQON-generation flow, aiming to take precedence over any prior training.
2. Update the `text` response of `get-sqon-schema` to be a SQON generation guide, while keeping the `structuredContent` as the full JSON schema.

Local re-testing with `gemma-4-e4b` following these 2 enhancements has shown promising results so far, allowing the model to generate valid SQON for some basic example queries.

### MCP Server

* Added few-shot examples of valid SQON to the `sqon` parameter description in the `execute-query` tool
* Added a "SQON Cheat Sheet" as the `text` response from the `get-sqon-schema` tool, while still returning the full schema response in `structuredContent`
  * The cheat sheet is structured to help LLMs understand how to structure SQON and build a query, with examples to reinforce best practices and override any erroneous/outdated training data
* Updated the `execute-query` tool description to reinforce utilizing the "SQON cheat sheet"
* Updated `get-sqon-schema` integration tests to reflect new response structure

### Special Instructions

There are no additional steps required to test these changes.

## Readiness Checklist

- [x] **Self Review**
- [x] **Formatting**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation

### DIFF
--- a/.dev/tech-debt.md
+++ b/.dev/tech-debt.md
@@ -159,15 +159,6 @@ When Arranger Server (`apps/search-server`) is updated to use `catalogue`, the M
 **Fix:** (a) Docs/schema comments pass: update README.md:13, docs/usage/02-arranger-components.md (title + line 29), configs.json.schema:28, and console strings in configs/index.ts. (b) Identifier rename pass (separate commit): buildCatalogsFromFolder -> buildCatalogsFromDirectory, folderName -> directoryName in apps/search-server/src/configs/. (c) Cross-references: add pointer to docs/concepts.md early in docs/usage/02-arranger-components.md; introduce "filter clause" for leaf nodes in docs/sqon/03-sqon-in-detail.md.
 **Standalone:** yes; (a) is docs-only; (b) is a mechanical rename; (c) is a docs addition. All three independent.
 
-### `docs/concepts.md` SQON examples use `field` instead of `fieldName`
-
-**File:** `docs/concepts.md:44,56-57`
-**Severity:** medium (canonical concepts doc contradicts the runtime schema; actively trains LLMs and readers into the #1 invalid-SQON mistake)
-**Kind:** stale documentation
-**Issue:** The SQON examples in `docs/concepts.md` use `"field"` as the filter-clause key, but the runtime (`modules/sqon` `SqonLeafSchema`, and `buildQuery` in `graphql-router` which reads `content.fieldName`) requires `"fieldName"`. `field` is the legacy/GDC-era spelling and is exactly the key local LLMs keep producing when generating SQON for the MCP `execute-query` tool — the project's own docs reinforce the wrong prior. `docs/usage/00-query-processing.md` already uses `fieldName` correctly.
-**Fix:** Update the examples in `concepts.md` to `fieldName`; grep docs for any other `"field"` usages in SQON (not ES query) context.
-**Standalone:** yes — docs-only change
-
 ### `setup.md` references `.env.arrangerDev` which no longer exists in the repo
 
 **File:** `docs/setup.md`; step 2 of "Running the Arranger-Server"
--- a/apps/mcp-server/src/mcp/executeQueryTool.ts
+++ b/apps/mcp-server/src/mcp/executeQueryTool.ts
@@ -235,7 +235,7 @@ export const registerExecuteQueryTool = (server: McpServer, { client }: McpServe
 				'2. call get-catalogue-fields to discover valid field names and per-type SQON operators. ' +
 				'3. call get-sqon-schema to learn how to construct valid SQON and see worked examples. ' +
 				'DO NOT guess field names, you MUST call get-catalogue-fields. ' +
-				'DO NOT construct "sqon" without callling get-sqon-schema. ' +
+				'DO NOT construct "sqon" without calling get-sqon-schema. ' +
 				'The user is asked to review and confirm the generated GraphQL query before it runs (when the client supports elicitation).',
 			inputSchema,
 			outputSchema,
--- a/apps/mcp-server/src/mcp/executeQueryTool.ts
+++ b/apps/mcp-server/src/mcp/executeQueryTool.ts
@@ -33,7 +33,15 @@ const inputSchema = {
 	sqon: zod
 		.unknown()
 		.describe(
-			'SQON filter for the query. Required. For an unfiltered query ("show me everything"), pass the empty root SQON {"op": "and", "content": []} — never null. Use get-sqon-schema for the SQON structure and get-catalogue-fields for valid fields and per-type operators.',
+			'SQON filter for the query (required). A SQON is a tree of two node kinds: group and leaf. ' +
+				'A GROUP combines conditions: {"op": "and" | "or" | "not", "content": [ ...child nodes... ]} (content is an ARRAY). ' +
+				'A LEAF is one field condition: {"op": "in" | "gt" | ..., "content": {"fieldName": "<field>", "value": <scalar or array>}} (content is an OBJECT). ' +
+				'Key rule: in a leaf, "fieldName" and "value" go TOGETHER inside "content", "op" stays outside, and the key is "fieldName", never "field". ' +
+				'CORRECT (donors that are Female): {"op":"and","content":[{"op":"in","content":{"fieldName":"donors.gender","value":["Female"]}}]}. ' +
+				'WRONG (rejected): {"op":"and","content":[{"fieldName":"donors.gender","op":"in","content":{"value":["Female"]}}]} (fieldName must be inside content, not beside op). ' +
+				'WRONG (rejected): {"op":"and","content":[{"field":"donors.gender","op":"in","value":["Female"]}]} ("field" should be "fieldName", and the condition is not nested). ' +
+				'Always use a group as the root, even for a single condition. For an unfiltered query ("show me everything") pass {"op":"and","content":[]}, never null. ' +
+				'Call get-sqon-schema for the cheat sheet and get-catalogue-fields for valid field names and per-type operators.',
 		),
 	queryType: zod
 		.enum(['hits', 'aggregations', 'both'])
@@ -45,7 +53,7 @@ const inputSchema = {
 		.array(zod.string().min(1))
 		.optional()
 		.describe(
-			'Dot-notation document fields to return for each hit (e.g. "donor.age_at_diagnosis"). Do not guess field names — use get-catalogue-fields first. Omit to return only the total hit count.',
+			'Dot-notation document fields to return for each hit (e.g. "donor.age_at_diagnosis"). Do not guess field names; use get-catalogue-fields first. Omit to return only the total hit count.',
 		),
 	first: zod
 		.number()
@@ -68,7 +76,7 @@ const inputSchema = {
 		.array(zod.string().min(1))
 		.optional()
 		.describe(
-			'Fields to aggregate. Nested properties use double underscores (e.g. "donor__age_at_diagnosis"); dot notation is also accepted. Do not guess field names — use get-catalogue-fields first. Required when queryType is "aggregations" or "both".',
+			'Fields to aggregate. Nested properties use double underscores (e.g. "donor__age_at_diagnosis"); dot notation is also accepted. Do not guess field names; use get-catalogue-fields first. Required when queryType is "aggregations" or "both".',
 		),
 	includeMissing: zod
 		.boolean()
@@ -168,7 +176,7 @@ const validateRequest = ({
 /**
  * Asks the user to review and confirm the generated GraphQL request before it is executed,
  * using MCP elicitation. When the connected client does not advertise the elicitation
- * capability, confirmation is skipped and the query proceeds — the executed request is
+ * capability, confirmation is skipped and the query proceeds; the executed request is
  * always echoed in the tool response for transparency.
  * @returns `true` when execution may proceed, `false` when the user declined or cancelled.
  */
@@ -228,7 +236,8 @@ export const registerExecuteQueryTool = (server: McpServer, { client }: McpServe
 			title: 'Execute Arranger Query',
 			description:
 				'Execute a SQON-filtered query against one Arranger catalogue and return matching documents (hits), per-field aggregation summaries, or both. ' +
-				'Before calling this tool: use list-catalogues to find the catalogue, then get-catalogue-fields to discover valid field names and per-type SQON operators — never guess field names. ' +
+				'Translate the user request into a SQON filter tree for the "sqon" argument; if unsure of the structure, call get-sqon-schema for the SQON cheat sheet. ' +
+				'Before calling this tool: use list-catalogues to find the catalogue, then get-catalogue-fields to discover valid field names and per-type SQON operators; never guess field names. ' +
 				'The user is asked to review and confirm the generated GraphQL query before it runs (when the client supports elicitation).',
 			inputSchema,
 			outputSchema,
--- a/apps/mcp-server/src/mcp/executeQueryTool.ts
+++ b/apps/mcp-server/src/mcp/executeQueryTool.ts
@@ -33,15 +33,9 @@ const inputSchema = {
 	sqon: zod
 		.unknown()
 		.describe(
-			'SQON filter for the query (required). A SQON is a tree of two node kinds: group and leaf. ' +
-				'A GROUP combines conditions: {"op": "and" | "or" | "not", "content": [ ...child nodes... ]} (content is an ARRAY). ' +
-				'A LEAF is one field condition: {"op": "in" | "gt" | ..., "content": {"fieldName": "<field>", "value": <scalar or array>}} (content is an OBJECT). ' +
-				'Key rule: in a leaf, "fieldName" and "value" go TOGETHER inside "content", "op" stays outside, and the key is "fieldName", never "field". ' +
-				'CORRECT (donors that are Female): {"op":"and","content":[{"op":"in","content":{"fieldName":"donors.gender","value":["Female"]}}]}. ' +
-				'WRONG (rejected): {"op":"and","content":[{"fieldName":"donors.gender","op":"in","content":{"value":["Female"]}}]} (fieldName must be inside content, not beside op). ' +
-				'WRONG (rejected): {"op":"and","content":[{"field":"donors.gender","op":"in","value":["Female"]}]} ("field" should be "fieldName", and the condition is not nested). ' +
-				'Always use a group as the root, even for a single condition. For an unfiltered query ("show me everything") pass {"op":"and","content":[]}, never null. ' +
-				'Call get-sqon-schema for the cheat sheet and get-catalogue-fields for valid field names and per-type operators.',
+			'SQON filter for the query (required). ' +
+				'Call get-sqon-schema before using execute-query for SQON grammar, operators, and worked examples. ' +
+				'For an unfiltered query ("show me everything") pass {"op":"and","content":[]}, never null.',
 		),
 	queryType: zod
 		.enum(['hits', 'aggregations', 'both'])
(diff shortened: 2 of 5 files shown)


.dev/tech-debt.md
```diff
@@ -138,15 +138,6 @@ When Arranger Server (`apps/search-server`) is updated to use `catalogue`, the M
 **Fix:** (a) Docs/schema comments pass: update README.md:13, docs/usage/02-arranger-components.md (title + line 29), configs.json.schema:28, and console strings in configs/index.ts. (b) Identifier rename pass (separate commit): buildCatalogsFromFolder -> buildCatalogsFromDirectory, folderName -> directoryName in apps/search-server/src/configs/. (c) Cross-references: add pointer to docs/concepts.md early in docs/usage/02-arranger-components.md; introduce "filter clause" for leaf nodes in docs/sqon/03-sqon-in-detail.md.
 **Standalone:** yes; (a) is docs-only; (b) is a mechanical rename; (c) is a docs addition. All three independent.
 
-### `docs/concepts.md` SQON examples use `field` instead of `fieldName`
```
mistryrn, line 162: Removing this from tech debt as the docs-in-question have been corrected as of https://github.com/overture-stack/arranger/commit/834ad943ef0c9c467ae4e53dc4660bcc0099420e#diff-c282aa25f1dabaaba19197006c4aa6bab00e8d2b2100375c42b41b7cf4bc55cb

apps/mcp-server/src/mcp/tools.ts
```diff
@@ -10,6 +10,62 @@ import {
...
+ * is the generation guide, the structuredContent is the validation artifact.
+ *
+ * Smaller LLMs pattern-match examples far more reliably than they "execute" a recursive JSON
+ * Schema, so this leads with the single most common mistake (fieldName/value placed beside "op"
+ * instead of inside "content"), then gives the Group/Leaf shapes, a copyable leaf template, a
+ * build recipe, and natural-language-to-SQON worked examples. Keep it in sync by hand with the
+ * SQON schema in `@overture-stack/sqon` (modules/sqon/src/schema): the operator names and node
+ * shapes below are derived from it. Every example here is checked against `SqonSchema`.
+ */
+const SQON_CHEAT_SHEET = `SQON cheat sheet (Serializable Query Object Notation).
```
mistryrn, line 26: This is the result of a few revisions of a "SQON Cheat Sheet". Intended to be returned alongside the full JSON schema response (that is `structuredContent`, this gets returned as `text`) to guide the LLM in generating valid SQON. It can certainly be refined further, but so far the results are promising 🙏

apps/mcp-server/src/mcp/tools.ts
```diff
@@ -10,6 +10,62 @@ import {
...
+
+Field operators (the "op" of a Leaf):
+  in, not-in, some-not-in, all   match value(s); any field type; "value" is a scalar or array
+  gt, gte, lt, lte               numeric or date fields; "value" is a single number or date string
+  between                        numeric or date fields; "value" is a 2-element array, [min, max]
+  filter                         fuzzy text match; SPECIAL shape: {"op":"filter","content":{"fieldNames":["<field>"],"value":"<text>"}}  ("fieldNames" is a plural ARRAY; "value" is one string)
+
+How to build a SQON:
+  1. Find each field's path and its allowed operators with get-catalogue-fields.
+  2. Write every condition as a Leaf: {"op": <op>, "content": {"fieldName": <field>, "value": <value>}}.
```
justincorrigible: a tiny adjustment to help this tool provide its intended effect
```suggestion
  2. Write every condition as a Leaf: {"op": <op>, "content": {"fieldName": "<field>", "value": <value>}}, except for `filter`, where "fieldNames" is an array of field paths instead. (See operator table below.)
```

apps/mcp-server/src/mcp/executeQueryTool.ts
```diff
@@ -33,7 +33,15 @@ const inputSchema = {
 	sqon: zod
 		.unknown()
 		.describe(
```
justincorrigible, line 35: the inline grammar (Group/Leaf shapes) and both "WRONG/CORRECT" examples duplicate SQON_CHEAT_SHEET, and having two copies of the same instructions means two places to maintain if the examples change, which may lead to confusing the LLM.

that said, the hint at point-of-use is definitely valuable here, but the full grammar could be trimmed to just the key rule ("fieldName and value go together inside content") and the empty-root hint; the rest can live in the cheat sheet.

integration-tests/mcp-server/test/readTools.ts
```diff
@@ -31,17 +31,25 @@ export default ({ getClient, configuredCatalogues, expectedFieldsByCatalogue }:
...
-		assert.equal(typeof data.title, 'string');
-		assert.ok(data.schema);
-		assert.ok(data.operators);
-		assert.ok(Array.isArray(data.operators.combination));
-		assert.ok(Array.isArray(data.operators.field));
+		assert.ok(text.includes('SQON'), `expected a SQON cheat sheet, got: ${text}`);
+		assert.ok(text.includes('fieldName'), 'expected the cheat sheet to mention "fieldName"');
+
+		// The full machine-readable schema and operator metadata move to structuredContent.
+		const data = (result as { structuredContent?: Record<string, unknown> }).structuredContent;
```
justincorrigible: (disclosure: needed "help" because I haven't used this SDK)

`structuredContent` is already on the `CallToolResult` type returned by `callTool()` (it's the modern-protocol branch of the SDK union — see client/index.d.ts)... that makes this type cast is unnecessary. 

meaning you can just write `result.structuredContent` directly and TypeScript will (should?) type it as `Record<string, unknown> | undefined` without any assertion.
```suggestion
		const data = result.structuredContent;
```

apps/mcp-server/src/mcp/tools.ts
```diff
@@ -10,6 +10,62 @@ import {
...
+Grammar:
+  SQON  = Group | Leaf
+  Group = {"op": "and" | "or" | "not", "content": [ SQON, ... ]}      (content is an ARRAY of child nodes; combines conditions)
+  Leaf  = {"op": <field op>, "content": {"fieldName": "<dot.path>", "value": <scalar or array>}}   (content is an OBJECT; one condition)
+  Leaf template to copy and fill in:  {"op": "<field op>", "content": {"fieldName": "<field>", "value": <value>}}
+
+Field operators (the "op" of a Leaf):
+  in, not-in, some-not-in, all   match value(s); any field type; "value" is a scalar or array
+  gt, gte, lt, lte               numeric or date fields; "value" is a single number or date string
+  between                        numeric or date fields; "value" is a 2-element array, [min, max]
```
justincorrigible, line 38: this description surfaced a SQON schema gap, fixed [here](https://github.com/overture-stack/arranger/commit/4849e756ba8246fceb269e4ee7fa1ec12e4e4a84).
`BetweenFilterSchema` was using `min(2)`, so [18, 65, 100] passed validation and silently produced `gte: 18, lte: 100` which is fine but "weird".
... thanks to this I was also able to find other cases where not making `min(n)` requirements was causing silent failures and broken ES queries. 👏

apps/mcp-server/src/mcp/tools.ts
justincorrigible: follow up (July 1):  while "filter" remains supported for backwards compatibility, this operation will be renamed to "wildcard" on the next versions of SQON.

further reference, https://github.com/overture-stack/arranger/blob/main/docs/usage/03-sqon-in-detail.md#wildcard-operator

apps/mcp-server/src/mcp/tools.ts
```diff
@@ -10,6 +10,62 @@ import {
...
+
+Natural language to SQON examples:
+  Female donors:           {"op":"and","content":[{"op":"in","content":{"fieldName":"donors.gender","value":["Female"]}}]}
+  Age over 50:             {"op":"and","content":[{"op":"gt","content":{"fieldName":"donors.age_at_diagnosis","value":50}}]}
+  Female and age >= 18:    {"op":"and","content":[{"op":"in","content":{"fieldName":"donors.gender","value":["Female"]}},{"op":"gte","content":{"fieldName":"donors.age_at_diagnosis","value":18}}]}
+  Female or age over 65:   {"op":"or","content":[{"op":"in","content":{"fieldName":"donors.gender","value":["Female"]}},{"op":"gt","content":{"fieldName":"donors.age_at_diagnosis","value":65}}]}
+  Exclude deceased:        {"op":"and","content":[{"op":"not-in","content":{"fieldName":"donors.vital_status","value":["Deceased"]}}]}
+  Age between 18 and 65:   {"op":"and","content":[{"op":"between","content":{"fieldName":"donors.age_at_diagnosis","value":[18,65]}}]}
+  Fuzzy text match:        {"op":"and","content":[{"op":"filter","content":{"fieldNames":["donors.primary_site"],"value":"brain"}}]}
```
justincorrigible:
```suggestion
  Wildcard text match:        {"op":"and","content":[{"op":"wildcard","content":{"fieldNames":["donors.primary_site"],"value":"brain"}}]}
```

apps/mcp-server/src/mcp/tools.ts
```diff
@@ -10,6 +10,62 @@ import {
...
+  SQON  = Group | Leaf
+  Group = {"op": "and" | "or" | "not", "content": [ SQON, ... ]}      (content is an ARRAY of child nodes; combines conditions)
+  Leaf  = {"op": <field op>, "content": {"fieldName": "<dot.path>", "value": <scalar or array>}}   (content is an OBJECT; one condition)
+  Leaf template to copy and fill in:  {"op": "<field op>", "content": {"fieldName": "<field>", "value": <value>}}
+
+Field operators (the "op" of a Leaf):
+  in, not-in, some-not-in, all   match value(s); any field type; "value" is a scalar or array
+  gt, gte, lt, lte               numeric or date fields; "value" is a single number or date string
+  between                        numeric or date fields; "value" is a 2-element array, [min, max]
+  filter                         fuzzy text match; SPECIAL shape: {"op":"filter","content":{"fieldNames":["<field>"],"value":"<text>"}}  ("fieldNames" is a plural ARRAY; "value" is one string)
```
justincorrigible:
```suggestion
  wildcard                         case-insensitive substring match; SPECIAL shape: {"op":"wildcard","content":{"fieldNames":["<field>"],"value":"<text>"}}  ("fieldNames" is a plural ARRAY; "value" is one string)
```

integration-tests/mcp-server/test/readTools.ts
mistryrn: Good catch! Looks like the current type cast is load-bearing, as writing just `data = result.structuredContent` works for that line but causes the assertions in the following lines (`data?.version`, `data?.title`, `data?.schema` etc) to fail.

Seems related to `result` being a union of two shapes:

- the normal result: `{ ...; structuredContent?: Record<string, unknown>; isError?; content }`
- a compatibility result: `{ [x: string]: unknown; toolResult: unknown }`

The `assert.ok(data?...)` only narrows unknown to {}, so `data?.version`, `data?.title`, `data?.schema`, etc all fail with `TS2339: Property does not exist on type 'unknown'`

Instead of casting the whole `result` object, we can cast just the property we need though:
```suggestion
		const data = result.structuredContent as Record<string, unknown> | undefined;
```

Would that be acceptable in this case?

apps/mcp-server/src/mcp/tools.ts
```diff
@@ -10,6 +10,62 @@ import {
...
+
+Field operators (the "op" of a Leaf):
+  in, not-in, some-not-in, all   match value(s); any field type; "value" is a scalar or array
+  gt, gte, lt, lte               numeric or date fields; "value" is a single number or date string
+  between                        numeric or date fields; "value" is a 2-element array, [min, max]
+  wildcard                         case-insensitive substring match; SPECIAL shape: {"op":"wildcard","content":{"fieldNames":["<field>"],"value":"<text>"}}  ("fieldNames" is a plural ARRAY; "value" is one string)
+
+How to build a SQON:
+  1. Find each field's path and its allowed operators with get-catalogue-fields.
+  2. Write every condition as a Leaf: {"op": <op>, "content": {"fieldName": "<field>", "value": <value>}}, except for "filter", where "fieldNames" is an array of field paths instead. (See operator table below.)
```
justincorrigible:
```suggestion
  2. Write every condition as a Leaf: {"op": <op>, "content": {"fieldName": "<field>", "value": <value>}}, except for "wildcard", where "fieldNames" is an array of field paths instead. (See operator table below.)
```

integration-tests/mcp-server/test/readTools.ts
justincorrigible: that narrower cast sounds like the right call (even though I'm not a fan of casting myself), because casting only the property you're reading rather than the whole result is cleaner 👍

apps/mcp-server/src/mcp/tools.ts
```diff
@@ -10,6 +10,62 @@ import {
...
+  Always use a Group as the root, even for a single condition.
+
+THE MISTAKE TO AVOID (this is what usually fails):
+  In a LEAF, "fieldName" and "value" go TOGETHER inside "content"; "op" stays outside; the key is "fieldName", never "field".
+  CORRECT: {"op":"in","content":{"fieldName":"donors.gender","value":["Female"]}}
+  WRONG:   {"fieldName":"donors.gender","op":"in","content":{"value":["Female"]}}   (fieldName must be INSIDE content, not beside op)
+  WRONG:   {"field":"donors.gender","op":"in","value":["Female"]}                   ("field" should be "fieldName", and nothing is nested in content)
+
+Notes:
+  - "value" items are strings or numbers. Send booleans as the strings "true" or "false".
```
justincorrigible: in a recent update to the SQON module, we made `SqonScalarValueSchema` include `zod.boolean()`, so native JSON booleans are now accepted, so the the LLM doesn't need to stringify them anymore. 

```suggestion
  - "value" items are strings, numbers, or booleans. 
```

apps/mcp-server/src/mcp/executeQueryTool.ts
```diff
@@ -33,7 +33,15 @@ const inputSchema = {
...
+				'A LEAF is one field condition: {"op": "in" | "gt" | ..., "content": {"fieldName": "<field>", "value": <scalar or array>}} (content is an OBJECT). ' +
+				'Key rule: in a leaf, "fieldName" and "value" go TOGETHER inside "content", "op" stays outside, and the key is "fieldName", never "field". ' +
+				'CORRECT (donors that are Female): {"op":"and","content":[{"op":"in","content":{"fieldName":"donors.gender","value":["Female"]}}]}. ' +
+				'WRONG (rejected): {"op":"and","content":[{"fieldName":"donors.gender","op":"in","content":{"value":["Female"]}}]} (fieldName must be inside content, not beside op). ' +
+				'WRONG (rejected): {"op":"and","content":[{"field":"donors.gender","op":"in","value":["Female"]}]} ("field" should be "fieldName", and the condition is not nested). ' +
+				'Always use a group as the root, even for a single condition. For an unfiltered query ("show me everything") pass {"op":"and","content":[]}, never null. ' +
+				'Call get-sqon-schema for the cheat sheet and get-catalogue-fields for valid field names and per-type operators.',
```
justincorrigible: this is more or less what I meant in the comment above (please pardon if there are formatting issues. hard to edit here)
```suggestion
			'SQON filter for the query (required). ' +
			'Key rule: in a leaf, "fieldName" and "value" go TOGETHER inside "content"; "op" stays outside, and the key is "fieldName", never "field". ' +
			'Always wrap conditions in a group root, even for a single condition. ' +
			'For an unfiltered query ("show me everything") pass {"op":"and","content":[]}, never null. ' +
			'Call get-sqon-schema for the full cheat sheet (grammar, operators, worked examples) and get-catalogue-fields for valid field names.',
```

justincorrigible:
@mistryrn, as usual, great work on the cheat sheet iteration and the verification process!
the `text/structuredContent` split is the right call, because it puts the generation guide where LLMs actually read and keeps the validation artifact accessible for schema-constrained clients 👍 

one thing worth noting for context: the structural root causes (4 and 5 from the analysis: `zod.unknown()` input and unhelpful validation errors) are addressed in the `build_sqon` design at `.dev/docs/build-sqon-tool.md`, which means the cheat sheet is the right interim measure; yet `build_sqon` is the longer-term fix that removes SQON synthesis from the LLM entirely... worth cross-referencing there so the trajectory is visible to anyone picking up root causes 4 and 5 next.

also! worth logging as tech-debt, i.e. work not due for this PR: the cheat sheet is hand-maintained in `tools.ts` and needs to track operator changes in `modules/sqon`. as mentioned over Slack re. the `build-sqon` design doc, the operator section in particular could instead be generated from `getSqonFieldOperatorDetails()` at startup to remove the drift risk.

apps/mcp-server/src/mcp/executeQueryTool.ts
```diff
@@ -228,7 +230,12 @@ export const registerExecuteQueryTool = (server: McpServer, { client }: McpServe
 			title: 'Execute Arranger Query',
 			description:
 				'Execute a SQON-filtered query against one Arranger catalogue and return matching documents (hits), per-field aggregation summaries, or both. ' +
-				'Before calling this tool: use list-catalogues to find the catalogue, then get-catalogue-fields to discover valid field names and per-type SQON operators — never guess field names. ' +
+				'Before calling this tool you MUST: ' +
+				'1. use list-catalogues to find the catalogue. ' +
+				'2. call get-catalogue-fields to discover valid field names and per-type SQON operators. ' +
+				'3. call get-sqon-schema to learn how to construct valid SQON and see worked examples. ' +
+				'DO NOT guess field names, you MUST call get-catalogue-fields. ' +
+				'DO NOT construct "sqon" without callling get-sqon-schema. ' +
```
justincorrigible:
```suggestion
				'DO NOT construct "sqon" without calling get-sqon-schema. ' +
```

justincorrigible:
Solid improvement over the previous version! The cheat sheet is well structured: leading with "THE MISTAKE TO AVOID" and showing the CORRECT and WRONG leaf shapes side by side is exactly the right approach for smaller models that pattern-match examples rather than parse a recursive JSON Schema. The `wildcard` shape exception (`fieldNames` as a plural array) is called out explicitly, which was the other common point of failure. The execute-query tool description now mandates calling `get-sqon-schema` first via numbered steps, so the workflow is self-evident to the model.

One piece of context for the hand-maintained cheat sheet: the project's roadmap has a follow-on item (titled "MCP surface unification") to make `get-sqon-schema` "derive" its operator list automatically from `getSqonFieldOperatorDetails()` in `modules/sqon`, so the two stay in sync without a manual update. The SQON_CHEAT_SHEET JSDoc comment already flags this; just want to make sure you're aware the automation path exists when the time comes. For now, hand-maintenance is the right call given that the few-shot examples cannot be auto-generated anyway.

ready to squash and merge!! 👍

justincorrigible (approved):

PR #1079, opened 2026-06-25
add flag to disable GraphQL introspection with tests and related docs updates

justincorrigible:
PR for visibility and to allow testing before merging


justincorrigible:
updated branch to resolved conflicts from latest "main" changes

PR #1081, opened 2026-07-02
WIP: 🕸️ Bug / Support AWS OpenSearch Auth

demariadaniel:
## Summary

Adds AWS SDK Client Auth support for OpenSearch client

## Issues
https://github.com/nci-hcmi-catalog/portal/pull/1145

## Description of Changes
Adds AWS SDK & authentication boilerplate
Adds additional config for indicating connection with AWS

## Readiness Checklist

- [ ] **Self Review**
- [ ] **Formatting**
- [ ] **Local Testing**
- [ ] **Updated Tests**
- [ ] Documentation


apps/search-server/pnpm-lock.yaml
justincorrigible, line 1: until we migrate this codebase to use pnpm, please refrain from using the library or committing this file

thanks

modules/graphql-router/src/admin/schemas/ProjectSchema/utils.ts
justincorrigible, line 1: please exclude the modules/graphql-router/src/admin files out of this and all PRs


thanks

modules/graphql-router/src/router.ts
```diff
@@ -76,7 +75,7 @@ const arrangerRouter = async <Context extends ArrangerBaseContext>({
 			}));
 
 		const mappingFromIndex = await getIndexMapping({
-			enableDebug,
+			enableDebug: !!enableDebug,
```
justincorrigible: how is this change necessary?

modules/graphql-router/src/searchClient/index.ts
```diff
@@ -243,6 +250,28 @@ const createSearchClient = (clientConfig: SearchConfigWithClient): SearchClient
 			return createOpenSearchClient(options);
 		}
 
+		case 'opensearch-aws': {
+			const options: OSClientOptions = {
+				...clientConfig,
+				clientType,
+				...AwsSigv4Signer({
+					region: 'us-east-1',
```
justincorrigible: hard coding this would limit the library's usefulness to a specific subset of users who would otherwise choose a different mirror/server location

apps/search-server/pnpm-lock.yaml
demariadaniel: Sorry this was created by mistake, removed

modules/graphql-router/src/router.ts
demariadaniel: This is a Type error, enableDebug at line 58 is `boolean | undefined`, and `getIndexMapping` expects only a boolean.
We had to leverage `enableDebug` to troubleshoot in HCMI Dev.

modules/graphql-router/src/admin/schemas/ProjectSchema/utils.ts
demariadaniel: There are a lot of distracting Type errors in this folder, maybe we can remove it or gitignore? I'm often using Type errors to validate that my new changes haven't broken anything, but frequently end up impacting code in /admin

Suggested gitignore change:
https://github.com/overture-stack/arranger/pull/1081/changes#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947

I will revert the changes in this file, but that's the context / source of repeat issues

modules/graphql-router/src/searchClient/index.ts
demariadaniel: Yes this should be dynamic, I'm not sure yet what other features also may need to be included. 

This is still in draft until I can validate what other changes will need to be made to work with the context of HCMI. The AWS-ArrangerSearchClient config still needs improvement, this is only set up for testing.

modules/graphql-router/src/searchClient/types.ts
```diff
@@ -40,8 +44,13 @@ export type SearchClientIndicesOpenParams = { index: string | string[] };
 export type SearchClientIndicesPutMappingsParams = { index: string; body: Record<string, any> };
 export type SearchClientIndicesPutSettingsParams = { body: Record<string, any> };
 export type SearchClientBulkParams = { body: Record<string, any>[] };
-export type SearchClientCreateParams = { id: string; index: string; body: Record<string, any> };
-export type SearchClientDeleteParams = { index: string; id: string };
+export type SearchClientCreateParams = {
+	id: string;
+	index: string;
+	body: Record<string, any>;
+	refresh?: boolean | 'false' | 'true' | 'wait_for';
```
joneubank: This list of types is peculiar, is this inherited from some third party library or did we decide these types? If we decided them, lets simplify to a list of string literals and add some documentation for why we chose these. If they are inherited, could we simplify them in a similar way?

modules/graphql-router/src/searchClient/types.ts
```diff
@@ -40,8 +44,13 @@ export type SearchClientIndicesOpenParams = { index: string | string[] };
...
+};
+export type SearchClientDeleteParams = { index: string; id: string; refresh?: boolean | 'false' | 'true' | 'wait_for' };
```
joneubank: Once we settle on this `refresh` property type, maybe define it once and references it in multiple places since its a compound type.

modules/graphql-router/src/router.ts
joneubank: Alternately we could change the `getIndexMapping` signature to make `enableDebug` an optional argument and be `false` by default.

justincorrigible: sounds like a type problem in the function's definition, rather than something to be coerced in this manner 🤔 

@joneubank thoughts on this?

apps/search-server/package.json
```diff
@@ -21,9 +21,10 @@
 		"test": "tsx --test ./src/**/*.test.ts"
 	},
 	"dependencies": {
+		"@aws-sdk/credential-provider-node": "^3.972.60",
```
justincorrigible: this is not necessary. as a separate dependency (more to maintain on our end), and as it is a transient of the Opensearch module, you may be able to just leverage `@opensearch-project/opensearch`'s `AwsSigv4Signer.js` internal dynamic import instead (which you're already using)...

justincorrigible: on further review, this would be imported in `graphql-router` (and likely why your changes are working). the `search-server` is not the right place for this
```suggestion
```

modules/graphql-router/src/searchClient/index.ts
```diff
@@ -1,3 +1,6 @@
+import { defaultProvider } from '@aws-sdk/credential-provider-node';
+import { AwsSigv4Signer } from '@opensearch-project/opensearch/aws';
```
justincorrigible, line 1: not a request for change but just double checking: 
`@opensearch-project/opensearch/aws` is the v2 SDK path, while the v3 equivalent is `@opensearch-project/opensearch/aws-v3` (the current standard. v2 is "old")

justincorrigible:
To summarize my previous comments, as confirmed by reading the source of the installed package.

**`@aws-sdk/credential-provider-node` is not needed and the `getCredentials` wrapper can be removed entirely.**
`AwsSigv4Signer` from `@opensearch-project/opensearch/aws` already does the dynamic import internally:

```js
// seen in node_modules/@opensearch-project/opensearch/lib/aws/AwsSigv4Signer.js
const getAwsSDKCredentialsProvider = async () => {
  try {
    const awsV3 = await import('@aws-sdk/credential-provider-node');
    if (typeof awsV3.defaultProvider === 'function') {
      return awsV3.defaultProvider();
    }
  } catch (err) { /* falls back to aws-sdk v2 */ }
  // ...
};
```

The explicit `defaultProvider` import and the `getCredentials` wrapper in the `AwsSigv4Signer` call are duplicating this exactly. The fix is to drop both:

```ts
// remove: import { defaultProvider } from '@aws-sdk/credential-provider-node';

...AwsSigv4Signer({
  region: auth.region || '',
  service: auth.service,
  // no getCredentials - AwsSigv4Signer resolves credentials automatically
}),
```

And drop `@aws-sdk/credential-provider-node` from `apps/search-server/package.json` (and it should not be added to `modules/graphql-router` either... it is not needed there at all).

**Runtime note:** on AWS infrastructure (ECS task role, EC2 instance profile), \`@aws-sdk/credential-provider-node\` is already present via the SDK layer. Deployments that need it will have it; those that don't will get an explicit error from the OpenSearch client on startup.

modules/graphql-router/src/searchClient/index.ts
demariadaniel: That's interesting, the docs don't make this distinction. But I see what you're saying when I dig into the library.

demariadaniel: Referring to: https://docs.opensearch.org/latest/clients/javascript/index#authenticating-with-amazon-opensearch-service-aws-signature-version-4

package-lock.json
```diff
@@ -65,38 +65,20 @@
...
-			"resolved": "https://registry.npmjs.org/undici-types/-/undici-types-7.24.6.tgz",
-			"integrity": "sha512-WRNW+sJgj5OBN4/0JpHFqtqzhpbnV0GuB+OozA9gCL7a993SmU+1JBZCzLNxYsbMfIeDL+lTsphD5jN5N+n0zg==",
-			"dev": true,
-			"license": "MIT"
-		},
 		"apps/search-server": {
 			"name": "@overture-stack/arranger-search-server",
 			"version": "0.0.0-dev",
 			"dependencies": {
+				"@aws-sdk/credential-provider-node": "^3.972.60",
```
justincorrigible: this file seems to be out of date now

modules/graphql-router/src/searchClient/index.ts
justincorrigible: Ok, upon further review, a few things to clarify from my earlier comment:

`/aws is correct`, so please disregard the part about v2 vs v3... as your link shows, the official docs use /aws for SDK v3 too, and /aws-v3 is undocumented, so we can leave the import as is. unclear why their code doesn't match the docs but not an "us" problem today.

apps/search-server/.env.schema
joneubank, line 1: We aren't adding any code to consume these env variables, in fact we make no code changes to the server. Let's not add new env variables to our schema that are unused.

modules/graphql-router/src/searchClient/types.ts
```diff
@@ -1,19 +1,47 @@
...
+		type: 'standard';
+	}
+>;
+export type AWSAuthConfig = Prettify<
+	BaseAuthConfig & {
+		type: 'AWS';
+		region: string;
+		service?: 'es' | 'aoss';
+	}
+>;
```
joneubank: No reason to extend a base config.


```suggestion

export type StandardAuthConfig = {
	type: 'standard';
	password: string;
	username: string;
}

export type AWSAuthConfig = {
	type: 'AWS';
	password: string;
	username: string;
	region: string;
	service?: 'es' | 'aoss';
}
```

modules/graphql-router/src/searchClient/types.ts
```diff
@@ -1,19 +1,47 @@
...
+export type DefaultAuthConfig = Prettify<
+	BaseAuthConfig & {
+		type: 'standard';
+	}
+>;
+export type AWSAuthConfig = Prettify<
+	BaseAuthConfig & {
+		type: 'AWS';
+		region: string;
+		service?: 'es' | 'aoss';
```
joneubank: These constant names are so ambiguous, but I read the README and see that `es` means `OpenSearch`, while the other version is serverless.

It may be worth introducing some type aliases:
```ts
type AWS_SERVICE_OPENSEARCH = 'es';
type AWS_SERVICE_OPENSEARCH_SERVERLESS = 'aoss';
type AwsService = AWS_SERVICE_OPENSEARCH | AWS_SERVICE_OPENSEARCH_SERVERLESS;
```

modules/graphql-router/src/searchClient/types.ts
```diff
@@ -1,19 +1,47 @@
...
+	password: string;
+	type?: AuthTypes;
+	username: string;
+};
+export type DefaultAuthConfig = Prettify<
+	BaseAuthConfig & {
+		type: 'standard';
+	}
+>;
+export type AWSAuthConfig = Prettify<
```
joneubank: CamelCase standard for acronyms is to only capitalize the first character: `AwsAuthConfig`

... Its possible this has been inconsistently applied but we can move towards consistency with new names.

joneubank (changes requested):
Thanks for adopting the new configuration pattern.

modules/graphql-router/src/searchClient/types.ts
demariadaniel: I see the problem too. Just to be clear, this is comment points to an outdated change, the latest version does not contain string literals for 'false' or 'true'. These are definitions that reflect the expected types for the OS/ES libraries.

apps/search-server/.env.schema
demariadaniel: I wanted get a start on documenting the Auth changes and new config values, then updated the client side without revisiting. Will revise.

modules/graphql-router/src/searchClient/types.ts
demariadaniel: Yeah I saw that BaseAuthConfig just ends up cluttering Type things downstream. This is updated using StandardAuth. Would `BasicAuth` be fitting var name you think?

modules/graphql-router/src/searchClient/types.ts
demariadaniel: Both elasticsearch and opensearch are using a `boolean | 'wait_for'` definition, I've added a shared type for these

PR #1085, opened 2026-07-17
add tooltip classnames

samrichca:
## Summary

Adds tooltip classnames to Arranger Charts to support styling in OHCRN.

## Issues
- https://github.com/OHCRN/platform/issues/1769

### Arranger Charts
- Adds classnames to all divs in the Tooltip component so they can be styled.

## Readiness Checklist

- [x] **Self Review**
- [x] **Formatting**
- [x] **Local Testing**
- [ ] **Updated Tests**
- [ ] Documentation

### DIFF
--- a/modules/charts/src/components/TooltipContainer.tsx
+++ b/modules/charts/src/components/TooltipContainer.tsx
@@ -4,6 +4,7 @@ import { PropsWithChildren } from 'react';
 export const TooltipContainer = ({ children }: PropsWithChildren<{}>) => {
 	return (
 		<div
+			className="tooltip-container"
 			css={css({
 				fontFamily: 'Work Sans, sans-serif',
 				fontSize: '11px',
--- a/modules/charts/src/components/charts/Tooltip.tsx
+++ b/modules/charts/src/components/charts/Tooltip.tsx
@@ -24,19 +24,22 @@ export const Tooltip = (tooltipData: Bar | SunburstSegment) => {
 	const { value, label, suppressed } = normalizeTooltipProps(tooltipData);
 	return (
 		<TooltipContainer>
-			<div>
-				<div>{`${label}`}</div>
-				<div>
+			<div className="tooltip-wrapper">
+				<div
+					className="tooltip-label"
+					data-label={label} // use this as a CSS selector, to select labels by their content
+				>{`${label}`}</div>
+				<div className="tooltip-data">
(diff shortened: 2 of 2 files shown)


anncatton (commented):
looks good to me, will leave final approval with @justincorrigible

modules/charts/src/components/charts/Tooltip.tsx
```diff
@@ -24,19 +24,22 @@ export const Tooltip = (tooltipData: Bar | SunburstSegment) => {
 	const { value, label, suppressed } = normalizeTooltipProps(tooltipData);
 	return (
 		<TooltipContainer>
-			<div>
-				<div>{`${label}`}</div>
-				<div>
+			<div className="tooltip-wrapper">
+				<div
+					className="tooltip-label"
+					data-label={label} // use this as a CSS selector, to select labels by their content
```
justincorrigible: not an urgent issue, but a risk worth bringing up, since the whole point of that line is enabling exactly this kind of CSS targeting: `this will break for labels containing quotes`.

a label comes straight from the underlying facet/data value (`normalizeTooltipProps`, typed `string` without sanitization), so a label like `O'Brien` would need `CSS.escape()` or an attribute-substring selector `e.g. [data-label*="..."]`, rather than a naive `[data-label="O'Brien"]`, or else it will silently not match.

modules/charts/src/components/TooltipContainer.tsx
```diff
@@ -4,6 +4,7 @@ import { PropsWithChildren } from 'react';
 export const TooltipContainer = ({ children }: PropsWithChildren<{}>) => {
 	return (
 		<div
+			className="tooltip-container"
```
justincorrigible, line 7: a thought just occurred to me, non-blocking but would be useful: since this PR exists specifically so an external consumer can find and use these hooks, a short "stable CSS classes" section in the Readme, listing them, would help the next consumer avoid having to scour the source to discover what's "stylable"

modules/charts/src/components/charts/Tooltip.tsx
justincorrigible:
```suggestion
					data-label={label} // CSS selector hook for per-label styling; if building the selector dynamically in JS, escape with CSS.escape() first

```

anncatton (approved):
🚀

PR #1088, opened 2026-07-23
🚚 Chore: Rename MCP Resources and Tools to Use `snake_case` Instead of `kebab-case`

mistryrn:
## Summary

Renamed all MCP tool and resource identifiers from `kebab-case` to `snake_case`, to match the naming convention used by the majority of actively-maintained TypeScript MCP servers.

## Issues
- N/A

## Description of Changes

**Context:** an early reference example for this MCP server used kebab-case, supported by discussions supporting both as acceptable for TS MCP servers. However, a recent examination of popular actively maintained repos revealed that 90% use snake_case. These changes bring Arranger's MCP implementation in-line with the industry standard.

### MCP Server
- Tools renamed: `list-catalogues` → `list_catalogues`, `get-sqon-schema` → `get_sqon_schema`, `get-catalogue-fields` → `get_catalogue_fields`, `execute-query` → `execute_query`.
- Resources renamed: `arranger-server-introspection` → `arranger_server_introspection`, `arranger-sqon-schema` → `arranger_sqon_schema`, `arranger-catalogue-fields` → `arranger_catalogue_fields`.
- Updated and corrected outdated documentation relating to the MCP server.

### Special Instructions
No additional steps are required to test these changes.

## Readiness Checklist

- [x] **Self Review**
  - I have performed a self review of code
  - I have run the application locally and manually tested the feature
  - I have checked all updates to correct typos and misspellings
- [x] **Formatting**
  - Code follows the project style guide
  - Automated code formatters (ie. Prettier) have been run
- [x] **Local Testing**
  - Successfully built all packages locally
  - Successfully ran all test suites, all unit and integration tests pass
- [x] **Updated Tests**
  - Unit and integration tests have been added that describe the bug that was fixed or the features that were added
- [x] **Documentation**
  - All new environment variables added to `.env.schema` file and documented in the README
  - All changes to server HTTP endpoints have open-api documentation
  - All new functions exported from their module have TSDoc comment documentation

PR #1087, opened 2026-07-23
✨ Add `query-arranger` System Prompt and MCP Server `instructions` (overture-stack/admin#184)

mistryrn:
## Summary
This PR take's @MitchellShiell 's original [POC system prompt PR](https://github.com/overture-stack/arranger/pull/1068), rebases it against the latest `main`, and refactors it while making a few additions (tests, updated SQON reference, MCP server `instructions`).

### `query_arranger` Prompt
The prompt, when invoked by a client/host that supports it, accepts a single `goal` from the user describing what the researcher wants. It then returns a set of messages containing:

1. Workflow instructions for the LLM to follow (order of operations; instructions to classify the goal as "answerable", "unanswerable", "ambiguous", or "improper"; how to respond based on the classification)
2. A SQON reference
3. The original goal from the researcher

After consuming these messages, the LLM should be able to reliably respond to the user's query in a predictable manner.

### MCP Server `instructions`
The `instructions` are a static string returned as part of the MCP Server's `initialize` response, informing the client:

1. What the server is
2. Instructing the client on "what not to do" (i.e. common pitfalls to avoid, such as guessing catalogue/field names)
3. Providing the client with a general workflow for translating user intent into an Arranger query

## Issues
- https://github.com/overture-stack/admin/issues/184

### MCP Server
- Added `query_arranger` prompt from Mitchell's POC
- Refactored `prompts.ts`
- Pulled `SQON_CHEAT_SHEET` out of `tools.ts` and into its own module for ease of reference (and later, ease of deprecation)
- Added server-level `instructions` that get returned to client as part of MCP server `initialize` response

### Integration Tests
- Added integration tests for MCP Prompts, under `readPrompts.ts`
- Updated existing MCP Server integration tests to use `node:assert/strict` as per repo conventions
- Updated existing MCP Server initialization tests to account for `instructions`

### Docs
- Updated `06-ai-and-automation.md` usage docs to include reference to Prompts and Instructions

### Special Instructions
No additional instructions are required to test these changes. Simply run the MCP server or tests as usual:
```bash
# from project root

# start dev mcp server
npm run mcp-server:dev

# or, run new tests (make sure ES/OS is running first)
npm run test
```

## Readiness Checklist
- [x] **Self Review**
  - I have performed a self review of code
  - I have run the application locally and manually tested the feature
  - I have checked all updates to correct typos and misspellings
- [x] **Formatting**
  - Code follows the project style guide
  - Autmated code formatters (ie. Prettier) have been run
[...]

PR #1089, opened 2026-07-30
Overture docs update

MitchellShiell:
Documentation for the [Overture docs site](https://github.com/overture-stack/docs), plus factual corrections to two in-repo READMEs. No source code changes.

Three new pages, a restructure of `docs/usage/` into `docs/reference/`, and three corrections to the `graphql-router` README's network-search section.

## New pages

| Page | Sidebar | Source of truth |
| --- | --- | --- |
| `docs/reference/graphql-api.md` | GraphQL API (`sidebar_position: 2.5`) | the generated per-catalogue schema: the single `POST /graphql` endpoint, SQON and field selection, pagination, sorting, and a pointer to the Introspection API for runtime field discovery |
| `docs/charts.md` | Arranger Charts (`sidebar_position: 4`) | ported from `modules/charts/README.md`, covering `ChartsProvider` batching every registered chart into one GraphQL query, plus three screenshots |
| `docs/federated-search.md` | Arranger Federated search (`sidebar_position: 6`) | written from `modules/graphql-router/src/network/`, not from the README |

The federated search page is new coverage, not a port. A sweep of all 74 remote branches found no dedicated page anywhere, only scattered mentions plus an unmerged 2024 design doc on `origin/search-docs` whose design has since diverged from what shipped. The page states up front that federation returns aggregate counts, not documents (there is no federated `hits` list), and records that the feature appears under three names: federated search, network search, and network aggregation, with `network` being what configuration and GraphQL field names use.

## Restructure

- `docs/usage/*` becomes `docs/reference/*`, and `usage.mdx` becomes `reference.mdx`.
- `docs/usage/06-ai-and-automation.md` becomes `docs/mcp-server.md`, promoted out of the reference section to a top-level "Arranger MCP server" page.
- `docs/migration/v3.1.md` becomes `docs/reference/08-Migration/v3.1.md`.
- Renumbered as fallout: `07-defaults-and-limits` to `06`, `08-feature-flags` to `07`.
- Cross-links repointed across `overview.md`, `setup.md`, `reference.mdx`, `02-query-processing.md`, `06-defaults-and-limits.md` and `graphql-api.md`.

`docs/overview.md` also gains a Key Features bullet and a federation pathway; `01-arranger-configs.md` notes the optional fifth config file (`network.json`); `05-introspection.md`, `07-feature-flags.md` and `charts.md` gain federation cross-links.

## Corrections to `modules/graphql-router/README.md`, Network search

Three statements were wrong against the code in `src/network/`:

1. **Field merging was described as an intersection.** The code takes a union; nodes lacking a field contribute an `___aggregation_not_available___` sentinel bucket rather than dropping the field.
2. **`remoteRequests.headers` and `remoteNodes[].requests.headers` were presented as `arrangerRouter` options.** They are `search-server` config-file properties, normalized into a `customizeRemoteRequest` function, so passing them to the library directly is a silent no-op. Per-node headers also replace the global list rather than merging with it.
3. **The example used `documentType: 'FileAggs'`,** which resolves to `FileAggsAggregations` and fails discovery.

`apps/mcp-server/README.md` is updated for the docs-site URL change.

## Rebased onto `main`

Current with `main`, no conflicts. 11 commits absorbed, four resolved:

- **MCP tool names.** #1088 renamed them to snake_case on `main` while this branch still had kebab-case, so `main` wins on both the tool list and the fuller `execute_query` signature, verified against `apps/mcp-server/src`.
- **`docs/mcp-server.md`** was combined, keeping `main`'s new `query_arranger` Prompts block from #1087 alongside this branch's heading promotion out of `reference/`.
- **Two `.dev/` files** follow this branch's own `chore: drop .dev changes` commit, so `.dev/` matches `main` exactly. The exception is `tech-debt.md`, where both entries are kept because they are additive: this branch's OpenSearch entry records why its Makefile change removes the `start-os` target.

`main` also landed eight `graphql-router` changes around catalogue availability. `docs/reference/05-introspection.md` already documents the resulting `healthy`/`degraded`/`unhealthy` and `available`/`failed` values and all six error codes, each confirmed present in the code, so that page needed no update.

## Before merging

- Docs site build was green against the pre-rebase branch tip and has not been re-run since.


apps/mcp-server/README.md
```diff
@@ -4,6 +4,17 @@ This app is an MCP server that learns how to talk to Arranger by consuming Arran
 
 The current scaffold implements the Streamable HTTP MCP transport using **v1.x** of the official [MCP TypeScript SDK](https://github.com/modelcontextprotocol/typescript-sdk/tree/v1.x).
 
+## Tools
```
justincorrigible, line 7: these tool names are stale: the MCP server registers them as `snake_case`, not `kebab-case`.

- list-catalogues should be list_catalogues
- get-sqon-schema should be get_sqon_schema
- get-catalogue-fields should be get_catalogue_fields
- execute-query should be execute_query

these were renamed to snake_case in #1088. The PR description says this branch already rebased past that change, so it looks like this new table just didn't pick up the rename when it was written

README.md
```diff
@@ -12,9 +12,9 @@ Arranger is a versatile, model-agnostic data discovery API for OpenSearch and El
 
 Technical resources for those working with or contributing to the project are available from our official documentation site, the following content can also be read and updated within the `/docs` folder of this repository.
 
-- **[Arranger Overview](https://docs.overture.bio/docs/core-software/Arranger/overview)**
-- [**Setting up the Development Enviornment**](https://docs.overture.bio/docs/core-software/Arranger/setup)
-- [**Common Usage Docs**](https://docs.overture.bio/docs/core-software/Arranger/setup)
+- **[Arranger Overview](https://docs.overture.bio/develop/Arranger/overview)**
+- [**Setting up the Development Environment**](https://docs.overture.bio/develop/Arranger/setup)
+- [**Reference Docs**](https://docs.overture.bio/develop/Arranger/reference)
```
justincorrigible: since this content already lives in the same repo, three of these links could point at it directly instead of the docs site:
- Arranger Overview (line 15) to ./docs/overview.md
- Setting up the Development Environment (line 16) to ./docs/setup.md
- Reference Docs (line 17) to ./docs/reference/reference.mdx.
**one tradeoff worth knowing**: docs/reference/reference.mdx and the pages it links to use Docusaurus-specific syntax (the DocCardList component, ::: admonition blocks), which renders as plain text on GitHub rather than the docs site's styled callouts, still fully readable, just less polished

... your call on these.

README.md
```diff
@@ -27,7 +27,7 @@ Technical resources for those working with or contributing to the project are av
 ## Support & Contributions
 
 - For support, feature requests, and bug reports, please see our [Support Guide](https://docs.overture.bio/community/support).
-- For detailed information on how to contribute to this project, please see our [Contributing Guide](https://docs.overture.bio/docs/contribution).
+- For detailed information on how to contribute to this project, please see our [Contributing Guide](https://docs.overture.bio/develop/contributing).
```
justincorrigible: on the other hand, this Contributing Guide link (line 30) could similarly point at `./CONTRIBUTING.md`, which already exists at the repo root with matching content, and would be directly relevant to Arranger's own specific contributions rather than a generic Overture.
I dare say that should be the priority for this specific link, across all repos, because each repo has nuances that are not "generic" as non-devs may imagine can be.

...but **importantly**: GitHub also gives root-level CONTRIBUTING.md special treatment (it's auto-surfaced on new issues and PRs), so linking to it directly is **arguably more useful** than the external link, not just an equivalent swap at all. 

for contrast, the Support Guide and quickstart guides links are fine left as-is, since that content is org-wide or platform-wide and doesn't live in this repo

justincorrigible (changes requested):

apps/mcp-server/README.md
MitchellShiell: fixed!

README.md
MitchellShiell: updated, went with all three in-repo

README.md
MitchellShiell: Keeping it to Arranger in this PR but will update across the board

MitchellShiell: fixed

PR #1091, opened 2026-08-11
✨ #1080: Add `build_sqon` MCP Tool

mistryrn:
## Summary

Implements **[v1](https://github.com/overture-stack/arranger/blob/main/.dev/docs/build-sqon-tool.md#v1-scalar-operators-one-combinator)** of the `build_sqon` MCP tool, which returns a validated SQON and a plain-English summary when provided the following input:
- a catalogue ID,
- a combinator (`and` or `or`),
- a list of clauses to filter on (`fieldName`, `operator`, `value`, optional `negate`),
- (optionally) the SQON from a previous execution
 
This tool eliminates the need for LLMs to generate their own SQON for the `execute_query` tool, which significantly reduces context window usage.

**Note:** [v1](https://github.com/overture-stack/arranger/blob/main/.dev/docs/build-sqon-tool.md#v1-scalar-operators-one-combinator) of this tool only covers single-field scalar operators, with one combinator per call. Text operators, aliases, and mixed AND/OR are currently rejected by the input schema.

## Issues
- #1080 

### MCP Server
* Added `mcp/buildSqonTool.ts`, registering the new `build_sqon` tool
  * Given a catalogueId, combination (`and`/`or`), clauses, and an optional existingSqon, this tool:
    * **(1)** Resolves the provided catalogueId
    * **(2)** Fetches the available fields and operators for that catalogue from its introspection
    * **(3)** Validates any provided existingSqon against the catalogue
    * **(4)** Validates the provided clauses
    * **(5)** Builds, folds, and normalizes the clauses into a valid SQON object, combined with any existing SQON
    * **(6)** Compares the final filter count to the input clause count
    * **(7)** Returns the final SQON, a plain-English summary of what the SQON does, the submitted clause count, the final filter count, and an optional note explaining the difference in count
  * For v1, it only works with scalar operators (`in`, `not-in`, `gt`, `gte`, `lt`, `lte`, `between`) and ONE `combination (`and` or `or`) for the whole batch
* Added `arranger/clauseValidation.ts`, which validates every clause in a batch against catalogue introspection and reports one error per bad clause, so the whole batch can be fixed in one resubmission
* Added `arranger/sqonSummary.ts`, which renders the built SQON as one plain-English line using the catalogue's display names
* Updated the guidance surfaces to route SQON construction through `build_sqon`:
  * `executeQueryTool.ts`: tool description, `sqon` param description, and GraphQL error recovery text now point at `build_sqon` instead of `get_sqon_schema`
[...]

Issue #1080, opened 2026-06-25
AIM1-BO8: build_sqon tool


MitchellShiell:
## Problem

MCP-driven SQON generation produces inconsistent results. LLMs infer SQON structure from training data that is incomplete or out of date relative to the current spec. Operator names, value schemas, combination nesting rules, and property names like `fieldName` drift between spec versions and what models were trained on.

A concrete example: the current SQON spec uses `content.fieldName` as the property name for field references. An LLM trained on older Arranger documentation may generate `content.field` instead: a silent mismatch that produces invalid queries with no clear error at the generation site. This kind of training-data staleness cannot be corrected by prompting alone; it reappears every time the spec evolves.

## Solution

Add a `build_sqon` MCP tool that delegates SQON construction to a deterministic code path (`SqonBuilder`), removing the LLM from the JSON synthesis loop.

The LLM selects the field name, operator, and value from the available catalogue schema. The tool produces and validates the SQON. The LLM is the selector; the tool is the generator.

Example:
```
build_sqon(fieldName: "age", operator: "gt", value: 30)
→ { "op": "gt", "content": { "fieldName": "age", "value": 30 } }
```

This makes SQON construction:
- **Deterministic**: the same inputs always produce the same valid SQON regardless of model version or prompt variation
- **Validated at construction**: the builder rejects impossible operator/field-type combinations before they reach Elasticsearch
- **Spec-tracked**: as the SQON schema evolves, only the tool implementation changes; the LLM does not need retraining or re-prompting
- **Token-efficient**: cheaper than injecting SQON documentation into every prompt or routing through the `/introspection` endpoint as a runtime guide

## Scope

1. **`build_sqon` tool** in `apps/mcp-server/src/mcp/tools.ts`: accepts `fieldName`, `operator`, `value`, and optional `combination` type; calls `SqonBuilder` internally; returns a validated SQON object
2. **Agent-optimized tool description**: the tool schema carries descriptions that guide the LLM toward correct parameter selection: valid operators per field type and value type constraints; the tool's own description is the spec, no SQON documentation needed in the system prompt
3. **Versioned changelog**: as the SQON schema evolves, the tool interface version and changelog are machine-readable so agent behaviour decouples from model training data

## Prerequisites

[...]

PR #1093, opened 2026-08-25
#1080: ✨ Implement `build_sqon` v2

mistryrn:
## Summary

Implements v2 of the `build_sqon` MCP tool, bringing it to full operator parity with what is currently available in `modules/sqon`. Updates unit tests, integration tests, and documentation to match, and addresses a few tech-debt items which surfaced while planning this work.

Also fixes a bug in the SQON module's `reduceSqon` functionality where date-range bounds would get merged to `null`, caused by an unchecked `as number` cast.

## Issues
- #1080 

### MCP Server

Implemented v2 of the `build_sqon` tool: the `wildcard` text operator with the `fieldNames` (plural) clause shape, plus `all` and `some-not-in`, which `modules/sqon` already supported but v1 withheld.
- Added `checkFieldOperator` to `queryValidation.ts`, the shared field-and-operator check now used by both clause validations in `existingSqon`/`execute_query`
- Rejected `*` inside `in-like` values, which Arranger silently runs as a regular expression rather than an exact match, directing callers to use `wildcard` instead
- Stopped advertising unrestricted operators as applying to "any field type", which contradicted the per-type operator lists the catalogue actually enforces
- Summarized multi-field text clauses with display names joined by "or"
- Fixed a bug where `existingSqon` errors were not batched with clause errors, forcing multiple re-submissions to fix one call
- Aligned the capitalization of "unknown" across emitted messages, per PR #1091 feedback
- Update unit tests to reflect the new functionality: from 199 tests before, to 249 tests now

### SQON Module

- Fixed a bug in `reduceSqon` which corrupted merged date-range bounds to `null`, caused by `Math.max`/`Math.min` behind an unchecked `as number` cast. Bounds now compare numerically, parsed by timestamp, or lexicographically; unorderable bounds are kept as two clauses rather than merged.

### Integration Tests

- Added coverage for the v2 operators against real Arranger and ES: substring matching, any-field matches across `fieldNames`, negated wildcard, the `all` operator, and a multi-field wildcard through the aggregations path
- Added coverage for the asterisk `*` rejection and for `fuzzy` being refused (since it is not implemented)
- Added a case covering an invalid clause AND an unusable `existingSqon` reported in one response

### Docs

- Updated `docs/mcp-server.md`, `apps/mcp-server/README.md`, and `CHANGELOG.md` for the new operator coverage, the `fieldNames` shape, and the asterisk `*` rejection's deliberate asymmetry with `execute_query`'s raw `sqon` path
- Add tech-debt entries for the `applicableTo` divergence between `modules/sqon` and catalogue introspection, and for `integration-tests/mcp-server`'s tsconfig never typechecking app sources cleanly
 
### Special Instructions
Before running these changes, you should rebuild the `sqon` module to ensure you have the latest dist:
```bash
# from root
npm run sqon:build
```

## Readiness Checklist

- [x] **Self Review**
- [x] **Formatting**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] **Documentation**


apps/mcp-server/src/mcp/buildSqonTool.ts
```diff
@@ -102,6 +146,24 @@ const clauseSchema = () =>
...
 				.describe('Exactly two bounds, [min, max], ascending and inclusive at both ends.'),
 		}),
+		zod.object({
+			...clauseBase(),
+			operator: zod.literal(ALL_OPERATOR).describe(describeOperators([ALL_OPERATOR])),
+			value: zod
+				.array(scalarValue())
+				.min(1)
+				.describe('Every value the field must contain. An array even for one value, never a bare scalar.'),
+		}),
```
justincorrigible, line 156: **no change needed in this PR**
a follow-up PR stacked on this one adds a check here: refuse, or ask the caller to confirm, an `all` clause carrying more than one value unless the field's `isArray` is `true`. Flagging it now so it's visible before that PR is up.

---
context:
`all` can silently match nothing on a field that only ever holds one value, because it requires the field to contain every listed value at once, which is only possible if the field can actually hold more than one value at a time. If it can't, `all` with two or more values will never match anything, and this returns an empty result the same way a normal "no matches" query would, not an error...

rejoice! here's now a way to tell those two cases apart. my added merge resolving the conflicts with `main` brought in a new field-metadata property, `isArray`, reported per field by catalogue introspection: `true` means it can hold more than one value, `false` means it's declared to hold exactly one, and `null` means nothing declared either way. this clause here doesn't check it yet.

apps/mcp-server/src/mcp/buildSqonTool.ts
```diff
@@ -315,20 +464,36 @@ const foldClauses = ({
...
-			operator: clause.operator,
-			value: clause.value,
-		};
-
-		// `addFilterClause` dispatches scalar operators through a switch with no default, and
-		// returns undefined for anything outside it: a text operator, an operator alias, or an
-		// operator added to modules/sqon but not to buildScalarClause.
-		const next = addFilterClause(params);
-		// for v1 of build_sqon, this is unreachable. This guard exists as a failsafe for v2
+		const shared = { combination, existing: sqon, negate: clause.negate ?? false };
```
justincorrigible, line 467: **note from merge conflict resolution**
main's union-merge fix grafted onto this branch's dispatch

the pre-check just above (ending in its own `continue`) is `main`'s fix for the same issue `reduceSqon` used to over-merge: a plain `in` clause matching an existing field is "unioned" (aka "or") into it before this branch's wildcard/scalar dispatch below runs, so both changes apply rather than one overwriting the other. Verified: 250/250 apps/mcp-server tests pass, including this branch's wildcard/all/resolveExistingSqon coverage and main's union-merge tests together.

modules/sqon/src/builder/reduce.ts
```diff
@@ -66,23 +66,76 @@ const deduplicateValues = (node: SqonNode): SqonNode => {
...
+
+	return undefined;
+};
+
+/**
+ * Returns a new node that merges `incoming` into `existing` per the applicable reduction rule, or
+ * `undefined` when the rule cannot be applied because the two range bounds are not orderable. Only
+ * the range rules can decline; the value-merge rules concatenate and always apply.
+ */
+const mergeIntoExisting = (
```
justincorrigible, line 106: **note from merge conflict resolution**
both sides combined here, not one replacing the other

the op-selection logic, `foldIntoOutput`, and pivot-aware matching come from main's already-merged fix (`in` no longer merges under `and`; nothing merges under `not`). the date-aware bound comparison (`compareBounds`, above) and `mergeIntoExisting` returning `undefined` for un-orderable bounds are this branch's addition, carried forward unchanged. 

Verified: 165/165 modules/sqon tests pass against this combination, including the property-based `reduceSqon` tests.

CHANGELOG.md
```diff
@@ -66,8 +66,8 @@ See [docs/reference/08-Migration/v3.1.md](docs/reference/08-Migration/v3.1.md) f
...
+- **`build_sqon` merges two same-field `in` clauses by combining their value lists**: `status in ['active']` submitted alongside `status in ['pending']` builds `status in ['active', 'pending']`, meaning "either", and `notes` reports the merge so the lower filter count is explained rather than silent. "Either" is the correct reading on a single-valued field, where no document could satisfy both clauses at once. **The merge is not yet conditional on the field's `isArray`**, so on a field that can hold several values at once (`isArray: true`, or `null` where nothing declared it) the competing reading, "every one of these must be present", is equally legitimate and the merge picks "either" regardless. Use the `all` operator directly when you need that reading. Tracked in the repo's tech-debt notes.
```
justincorrigible, line 70: **no change needed in this PR**
the follow-up PR stacked on this one updates this wording once `all` itself enforces that check.

---

"Use `all` directly" needs the same caveat this bullet already gives the `in` merge

this bullet is from `main`, kept as part of resolving this PR's conflict, not something you wrote. it points to your `all` operator as the fix for the `in` merge's ambiguity; but `all` only reliably matches when the field can hold more than one value, a fact reported by the `isArray` field metadata this same merge brought in (`true` multi-valued, `false` single-valued, `null` undeclared). 

on a field that's `false` or `null`, `all` with more than one value can't match either, the identical failure this bullet is warning about for `in`, just on the operator it now recommends.

docs/mcp-server.md
```diff
@@ -37,14 +37,18 @@ The server returns a short set of usage instructions that most clients fold into
...
+An asterisk inside an `in`, `not-in`, `some-not-in`, or `all` value is rejected, because Arranger runs such a value as a regular expression rather than matching it literally: use `wildcard` instead. `execute_query`'s raw `sqon` parameter still accepts it, so an asterisk-bearing keyword value is reachable there but not through `build_sqon`.
+
+Two `in` clauses on the same field also merge, by combining their value lists: `status in ['active']` together with `status in ['pending']` becomes `status in ['active', 'pending']`, meaning "either". That is the correct reading on a single-valued field, where no document could satisfy both clauses at once. It is not conditional on the field's `isArray` yet, so on a field that can hold several values at once (`isArray: true`, or `null` where nothing declared it) the other reading, "every one of these must be present", is equally legitimate and the merge silently picks "either" regardless. Use the `all` operator directly when you need that reading.
```
justincorrigible, line 49: **no change needed in this PR**
the follow-up PR stacked on this one updates this wording once `all` itself enforces that check.

---

"Use `all` directly" needs the same caveat this paragraph already gives the `in` merge (same as the CHANGELOG.md entry)

this paragraph is from `main`, kept as part of resolving this PR's conflict, not something you wrote. it points to your `all` operator as the fix for the `in` merge's ambiguity; but `all` only reliably matches when the field can hold more than one value, a fact reported by the `isArray` field metadata this same merge brought in (`true` multi-valued, `false` single-valued, `null` undeclared). 

on a field that's `false` or `null`, `all` with more than one value can't match either, the identical failure this paragraph is warning about for `in`, just on the operator it now recommends.

justincorrigible (approved):
ready to squash. took the liberty to re-sync with `main` after some changes I made earlier, and then left comments here to illustrate how that rippled into your branch.

will open a new PR tomorrow with the rest of the related work, so that we can land this v2 part of the `build_sqon` 👍

PR #1096, opened 2026-09-02
chore: Baseline the SQON Contract Before Migrating to Zod 4 (Zod Upgrade Pt 1)

mistryrn:
## Summary

**PR 1 of 3** for the Zod 4 migration.
- PR 2: #1097 
- PR 3: #1098 

Pins the `GET /introspection/sqon` response with fixtures and tests so the Zod 4 migration in [PR 2](https://github.com/overture-stack/arranger/pull/1097) has something to diff against, and drops three dependencies `apps/mcp-server` never used. No behaviour changes.

**No breaking changes.** This PR is additive plus a dependency removal, and is safe to merge on its own even if PRs 2 and 3 stall.

## Issues
- N/A

### Search Server

* Added two fixtures and four tests covering the `GET /introspection/sqon` response so that changes to the published contract are caught and handled deliberately
  * The endpoint is read directly by REST clients and returned verbatim by the MCP Server's `get_sqon_schema` tool, but nothing validated its shape previously
  * Split across `introspectionSqon.schema.json` and `introspectionSqon.metadata.json` so that schema changes and operator-metadata changes stay out of each other's diffs; PR 2 causes changes to schema only
  * The version is stored as a placeholder and keys are sorted, so neither a release nor a key reordering churns the files
* Added `generateIntrospectionSqonFixtures.mts` and a `fixtures:introspection-sqon` npm script to regenerate them, with a `scripts/README.md` covering when to run it and why the fixtures exist
  * `buildSqonDetails()` is a zero-argument pure function, so regenerating needs no server, Elasticsearch, or network

### MCP Server

* Removed unused `express`, `cors`, and `@types/cors` from `dependencies` so that the manifest reflects what the app actually uses
  * Nothing imports either at runtime: the only reference is a type-only `express` import in `src/http/app.ts`, and the serving app is built by the SDK's own bundled Express 5
  * Both stay resolvable through `apps/search-server` and `modules/graphql-router`, and `@types/express` stays declared because `src/http/app.ts` needs `Request` and `Response`

### Docs

* Corrected the tech-debt entry on dangling `$ref` pointers to describe a latent defect rather than a live one
  * Every pointer in the published schema currently resolves, and only because of `$defs` declaration order; reordering the map is enough to break one
* Reframed the Express entry around the type/runtime skew that remains, now that the duplicate Express 4 copy is gone

### Special Instructions
Before running these changes locally, you should re-install dependencies and rebuild the SQON module:
```sh
npm ci                 # dependency removals changed package-lock.json
npm run modules:build
```

## Readiness Checklist

- [x] **Self Review**
- [x] **Formatting**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] **Documentation**

### DIFF
--- a/apps/mcp-server/package.json
+++ b/apps/mcp-server/package.json
@@ -25,9 +25,7 @@
 		"@modelcontextprotocol/sdk": "^1.29.0",
 		"@overture-stack/arranger-types": "file:../../modules/types",
 		"@overture-stack/sqon": "file:../../modules/sqon",
-		"cors": "^2.8.6",
 		"dotenv": "^16.6.1",
-		"express": "^4.21.2",
 		"pino": "^10.3.1",
 		"pino-pretty": "^13.1.3",
 		"tsx": "^4.21.0",
@@ -35,7 +33,6 @@
 	},
 	"devDependencies": {
 		"@tsconfig/node22": "^22.0.5",
-		"@types/cors": "^2.8.12",
 		"@types/express": "^4.17.14",
 		"@types/node": "^25.6.2",
 		"typescript": "^5.8.3"
--- a/apps/search-server/scripts/generateIntrospectionSqonFixtures.mts
+++ b/apps/search-server/scripts/generateIntrospectionSqonFixtures.mts
@@ -1,0 +1,77 @@
+#!/usr/bin/env tsx
+/**
+ * Regenerates the committed `GET /introspection/sqon` fixtures.
+ *
+ * Usage: npm run fixtures:introspection-sqon -w apps/search-server
+ *
+ * See ./README.md for when to run this, and for why the fixtures exist at all. Read the resulting
+ * `git diff`: it is a change to a published contract, so decide whether SQON_SCHEMA_VERSION needs a
+ * bump before committing.
+ *
+ * `buildSqonDetails()` is a zero-argument pure function, so no server, Elasticsearch, or network is
+ * involved. Three details the test depends on: the response is split in two so schema and operator
+ * changes stay out of each other's diffs, the version is stored as a placeholder so releases do not
+ * churn the files, and keys are sorted then Prettier-formatted so the diff stays readable and
+ * `scripts/format-all.sh` cannot silently reformat them.
+ */
+import { execFileSync } from 'node:child_process';
+import { writeFileSync } from 'node:fs';
+import { dirname, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+import { SQON_SCHEMA_VERSION } from '@overture-stack/sqon';
+
+import {
+	METADATA_FIXTURE_FILENAME,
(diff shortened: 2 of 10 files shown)


apps/mcp-server/package.json
```diff
@@ -25,17 +25,14 @@
 		"@modelcontextprotocol/sdk": "^1.29.0",
 		"@overture-stack/arranger-types": "file:../../modules/types",
 		"@overture-stack/sqon": "file:../../modules/sqon",
-		"cors": "^2.8.6",
```
mistryrn, line 28: Dropped some unused deps from the MCP Server ahead of the Zod upgrade in [PR 2](https://github.com/overture-stack/arranger/pull/1097).

apps/search-server/scripts/generateIntrospectionSqonFixtures.mts
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env tsx
+/**
+ * Regenerates the committed `GET /introspection/sqon` fixtures.
```
mistryrn, line 3: This script is explained in more detail in the accompanying README, but the TL;DR behind it all is:
- Zod 4 will change how the SQON schema JSON is generated
- this script, the fixtures it generates, and the accompanying tests will be used to verify that the Zod 4 upgrade produces an equivalent SQON introspection response

apps/search-server/src/introspection/introspectionSqon.metadata.json
mistryrn, line 1: Fixture that can/should be regenerated by the accompanying script in the event of an **operator/metadata** change that causes the related tests to fail.

apps/search-server/src/introspection/introspectionSqon.schema.json
mistryrn, line 1: Fixture that can/should be regenerated by the accompanying script in the event of a **schema** change that causes the related tests to fail.

justincorrigible (commented):
s-q-u-a-s-h 🐛

apps/search-server/src/introspection/introspectionSqonFixtures.ts
```diff
@@ -0,0 +1,61 @@
...
+			.sort()
+			.map((key) => [key, sortKeysDeep((value as Record<string, unknown>)[key])]),
+	);
+};
+
+/**
+ * Reads a fixture, substituting the live schema version back in.
+ *
+ * `readFileSync` rather than a JSON import: imported JSON is one cached object shared by every
+ * importer, and the substitution below mutates.
```
justincorrigible, line 47: note to self: this not `mutation`, strictly speaking: `replaceAll` runs on the file string pre-parse and `JSON.parse` returns a fresh object...  and the real reason `readFileSync` is needed is that placeholder substitution is textual and must happen before parsing.

will update myself once merged to ensure clarity for posterity

justincorrigible (approved):

PR #1097, opened 2026-09-02
feat: Migrate to Zod 4 (Zod Upgrade Pt 2)

mistryrn:
## Summary

**PR 2 of 3** for the Zod 4 migration.
- PR 1: #1096 
- PR 3: #1098 

Upgrades `modules/sqon` and `apps/mcp-server` to Zod 4 (which MCP v2 will require), and rewrites the SQON JSON Schema generator using Zod 4's native `toJSONSchema`. Also fixes a crash reachable from an unauthenticated MCP input.

**Three published-contract changes**, all recorded in `CHANGELOG.md` under `[3.1.0] - Unreleased`:
  1. `@overture-stack/sqon` and `@overture-stack/arranger-types` now require Zod 4 as a **peer dependency**. Consumers on Zod 3 must upgrade, since these packages export Zod schema instances and two Zod majors in one tree do not interoperate.
  2. The published SQON JSON Schema changed _shape_ due to differences in Zod 4's `toJSONSchema` output. It **accepts exactly the same SQONs**; only how it is _written_ changed.
  3. `execute_query` now rejects a missing `sqon` at the schema layer rather than in the handler.

~**🚨 `SQON_MAX_DEPTH` was set to 500 and needs a team decision.** It carries a `TODO(review before merge)` in `modules/sqon/src/schema/depth.ts` with the case for 128 written out. Agreeing a final number is the one thing in this PR explicitly waiting on the team.~ **✅ UPDATE 2026-09-02:** `128` was chosen as the `SQON_MAX_DEPTH` as it consistently passed across environments while still providing more than enough depth.

**❓ Open question for @justincorrigible :** [`modules/types/src/sqon.ts`](https://github.com/overture-stack/arranger/blob/main/modules/types/src/sqon.ts) is a bare `export * from '@overture-stack/sqon'`, so `@overture-stack/arranger-types` republishes the whole sqon surface including Zod instances. Nothing in this repo consumes it that way, and its origin looks like a compatibility shim whose companion line was deleted 11 days later. This PR takes the conservative route (same peer declared on `modules/types`), which is correct either way; the answer only decides whether the re-export can be narrowed later.

## Issues
- N/A

### modules/sqon

* Bumped `zod` to `^4.2.0` and moved it to `peerDependencies` plus a devDependency, so that a consumer on a different major cannot end up with two copies of a package that exports Zod schema instances
* Rewrote `getSqonJsonSchema` using `zod.toJSONSchema` and dropped [`zod-to-json-schema`](https://github.com/StefanTerdell/zod-to-json-schema), which is unmaintained since November 2025 and cannot consume v4 schemas
  * `additionalProperties: true` became the equivalent `{}`, shared value subschemas inline instead of pointing through another definition's path, and canonical-plus-alias operators emit as a two-branch `oneOf` instead of one flat `enum`
  * Kept the `anyOf` to `oneOf` rename so the published keyword does not move twice in one release; the case for `anyOf` is logged as its own tech-debt entry
* Capped SQON nesting depth so that an over-deep payload fails validation instead of throwing an uncaught `RangeError` out of `safeParse`
  * `safeParse` is documented never to throw, and the recursion overflowed the stack on about 25KB of JSON reachable from unauthenticated MCP input
  * The check runs through a pipe so it fires before the recursion; confirmed no throw at 20,000 nested combinations
  * `SQON_MAX_DEPTH` and `checkSqonDepth` are exported so callers can apply a stricter limit
* Narrowed `SqonFieldFilter` from `SqonNode &` to `SqonLeaf &` so that negative narrowing keeps working, and to match what `isFieldFilter` already does at runtime
  * Zod 4's flatter inference made TypeScript collapse `Exclude<SqonNode, SqonFieldFilter>` to `never`, breaking every downstream narrowing in `buildSqonTool.ts`

### modules/types

* Declared `zod^4.2.0` as a peer, since it re-exports the entire sqon surface and previously carried no `zod` of its own
  * Without this, a consumer installing only `@overture-stack/arranger-types` would have an unsatisfied peer chain and `SqonSchema`'s type would stop resolving

### MCP Server

* Fixed the breaks Zod 4 introduces: eight single-argument `zod.record` sites, which fail at runtime rather than returning a validation error, and `ZodError.errors` becoming `.issues`
* Kept the actionable error text when `sqon` is omitted so that an agent is still told to call `build_sqon`, or to pass an empty root SQON for an unfiltered query
  * Zod 4 makes an `unknown()` key required, so the SDK now rejects before the handler runs, and the default message is a bare "expected nonoptional" that names no fix
  * `.nonoptional({ error })` carries the existing text into the schema error; `SQON_REQUIRED_MESSAGE` is shared so the schema path and `validateSqon`'s explicit-`null` path cannot drift

### Search Server

* Regenerated `introspectionSqon.schema.json`. The metadata fixture is untouched, which is the split from PR 1 doing its job.

### Integration Tests

* No source changes. The MCP suite caught the `execute_query` error-text regression above and now passes unmodified, which is the signal that the agent-visible contract is preserved.

### Special Instructions

Before running these changes locally, you should re-install dependencies and rebuild the SQON module:
```sh
npm ci                   # zod moved to peerDependencies; lockfile changed
npm run modules:build    # modules/sqon must be rebuilt before search-server or mcp-server tests
```

## Readiness Checklist

- [x] **Self Review**
- [x] **Formatting**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] **Documentation**


modules/sqon/src/schema/depth.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Measured at default stack size: 750 combinations parse (18.8KB), 1000 throws `RangeError` out of
+ * `safeParse`.
+ *
+ * TODO(review before merge): confirm this number as a team. 500 is deliberately permissive so the
+ * guard cannot reject an existing caller while the limit is still open. 128 is defensible too:
+ * still 10x any real query, but an order of magnitude below a ceiling that is stack-dependent
+ * rather than constant, and so only accidentally safe at 500.
+ */
+export const SQON_MAX_DEPTH = 500;
```
mistryrn: Open question for the team: **what do we want to set as the max depth for a SQON?** See the code comment above this for full context, but around a depth of 750-1000 causes a crash.

Arbitrarily set to 500 for this PR as a very permissive upper limit without getting too close to the "danger zone"

mistryrn: Commenting here for posterity: `128` was ultimately chosen as it passed consistently across environments while still providing more than enough depth.

modules/sqon/src/schema/index.ts
```diff
@@ -83,16 +85,40 @@ export const SqonLeafSchema = zod.union([
 	WildcardFilterSchema,
 ]);
 
-export const SqonCombinationSchema: zod.ZodType<SqonCombination> = zod.lazy(() =>
+export const SqonCombinationSchema: zod.ZodType<SqonCombination, SqonCombination> = zod.lazy(() =>
```
justincorrigible: `SqonCombinationSchema` is exported from the package root at `src/index.ts:29`, and it does not go through `checkSqonDepth`. `SqonNodeSchema` on the other hand, just below this block, has the same gap but is "private" and says so in its comment. The exported one has neither.

I built a SQON nested 20,000 levels deep and ran it through both:
```
SqonSchema.safeParse(nest(20000))             -> success=false
SqonCombinationSchema.safeParse(nest(20000))  -> throws RangeError
```
The throw is the thing. `safeParse` is documented never to throw, so a caller who wrapped it the way the docs say still crashes. That is the exact defect this PR fixes for `SqonSchema`, and this export is the one path where the issue survives.

To be clear on urgency: this is not a regression... `main` already exports the same schema and has no guard anywhere, so both schemas throw there today and this PR strictly improves that. It just leaves one door open. Worth doing now only because the change is tiny and you are already in this file 👍

What limits the impact is the type. It is `ZodType<SqonCombination, SqonCombination>`, so a TypeScript caller cannot hand it raw input without casting first. One reaches the throw from JavaScript, or by casting on purpose. That is why this is "worth finishing" rather than "urgent".

Three options, your pick:

- Remove `SqonCombinationSchema` from `src/index.ts:29`. Looks safe: the only in-repo user is `jsonSchema/runtime.ts`, and it imports from `../schema/index.js` rather than the package root, so nothing internal breaks.
- Give it the same warning comment `SqonNodeSchema` has, and leave the behavior alone.
- Pipe it through `checkSqonDepth` the way `SqonSchema` does, so it stops throwing.

modules/sqon/src/schema/index.ts
```diff
@@ -83,16 +85,40 @@ export const SqonLeafSchema = zod.union([
...
+ * `RangeError`. SQONs arrive as untrusted request content, so that is a crash reachable with about
+ * 25KB of JSON. The pipe short-circuits: an over-deep value fails validation and never recurses.
+ */
+export const SqonSchema: zod.ZodType<SqonNode, unknown> = zod
+	.unknown()
+	.superRefine((value, ctx) => {
+		if (!checkSqonDepth(value)) {
+			ctx.addIssue({
+				code: 'custom',
+				message: `SQON nests deeper than the maximum JSON nesting depth of ${SQON_MAX_DEPTH}.`,
```
justincorrigible: The message names "JSON nesting depth", and a caller reading it counts combinations. Converting the number for them is tempting but does not work, because the combination limit moves with the shape of the innermost leaf.

Your `nest()` builds `value: ['a']`, and for that leaf, 62 is exactly right. An array adds a JSON level of its own, so a scalar-valued leaf gets one more combination:

```
{ op: 'in', content: { fieldName: 'f', value: ['a'] } }  -> 62 combinations
{ op: 'in', content: { fieldName: 'f', value: 'a'   } }  -> 63
```
Not a problem in the code, and the number in `depth.ts` and the CHANGELOG is right for the common case. It only matters if the message tries to name a combination count, since that count would then be wrong for some inputs while the JSON depth number stays exact/correct.

Suggestion: we keep the JSON depth number and just need to say somehow that is the unit, with the rough conversion alongside. e.g. The comment in `depth.ts` already phrases it that way, with "roughly twice the combination count".

modules/sqon/src/schema/depth.ts
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @param value - Any value, typically an unvalidated SQON from a request.
+ * @param maxDepth - Defaults to {@link SQON_MAX_DEPTH}.
+ * @returns `true` when `value` nests no deeper than `maxDepth`.
+ */
+export const checkSqonDepth = (value: unknown, maxDepth: number = SQON_MAX_DEPTH): boolean => {
+	const stack: { node: unknown; depth: number }[] = [{ node: value, depth: 0 }];
+
+	while (stack.length > 0) {
+		const { node, depth } = stack.pop() as { node: unknown; depth: number };
```
justincorrigible: Tiny one, feel free to drop it.

`stack.pop()` is typed `T | undefined`, and the `while (stack.length > 0)` right above already rules the `undefined` case out, so the cast is safe. It just means TypeScript is not checking something it could check if the code were shaped a little differently:

```
const item = stack.pop();
if (!item) { break; }
const { node, depth } = item;
```

Only raising it because `depth.ts` is new in this PR and it is a guard, so it is the one spot where a silenced type actually costs something. There is plenty of `as` elsewhere in `modules/sqon` (my mistake as a TS n00b, pushing to rush through things) and I am not suggesting a sweep.

justincorrigible (changes requested):
heaps of improvements to the codebase with this upgrade! nice work @mistryrn

modules/sqon/src/schema/index.ts
mistryrn: Agreed! Replacing the message with the following:
```ts
`SQON exceeds the maximum nesting depth of ${SQON_MAX_DEPTH},
counted in JSON levels (roughly twice the number of nested
filter combinations).`
```
Marked as "roughly" twice the number of nested filter combinations, with no combination number stated as fact. Same phrasing as `depth.ts` 🙏

modules/sqon/src/schema/depth.ts
mistryrn: Great catch! Updated accordingly 🙇

modules/sqon/src/schema/index.ts
mistryrn: Great catch! Went with option 3, slightly modified. Piping the guard onto `SqonCombinationSchema` directly would have re-run it at _every_ nested level, since that schema is what the recursion re-enters.

Instead, I split it into `SqonCombinationNodeSchema` (structure and recursion) and `SqonCombinationSchema` (the "NodeSchema" wrapped in a shared `withDepthGuard` function), mirroring the existing `SqonSchema`/`SqonNodeSchema` pair. Both exported schemas now take unknown, which also closes the cast route you mentioned, so this is safe from TypeScript too.

One thing to note: `jsonSchema/runtime.ts` has to register the **unguarded** schemas, because registering the piped one emits a malformed `$ref`. As a result `Group.content.items` is now `{"$ref": "#/$defs/SQON"}` rather than an inlined union. It resolves, and the tests and `CHANGELOG` are updated.

justincorrigible (approved):
👏 (silent standing ovation)

PR #1099, opened 2026-09-05
feat: Upgrade `apps/mcp-server` to MCP SDK v2 and protocol revision `2026-07-28`

mistryrn:
## Summary

Upgrades `apps/mcp-server` to MCP SDK v2 and protocol revision `2026-07-28`, which removes protocol sessions, the `initialize` handshake, and server-initiated requests, so the transport and the confirm-before-execute flow both had to be rebuilt rather than ported. **Breaking for consumers:** the endpoint serves `2026-07-28` only, and every SDK client negotiates the 2025 era by default, so a host must opt into modern negotiation explicitly.

Please refer to [`.dev/docs/mcp-sdk-v2-changes.md`](https://github.com/overture-stack/arranger/blob/d321d8982e9ab8cd1770dbba7c08b5c1a14d00d8/.dev/docs/mcp-sdk-v2-changes.md) for a short summary of the key changes in MCP v2 and how they impacted our MCP Server.

## Issues
- https://github.com/overture-stack/admin/issues/276

### MCP Server

- Upgraded to `@modelcontextprotocol/{server,node}@2` and protocol revision `2026-07-28`
- Dropped support for 2025-era clients, because that path cannot carry `execute_query`'s confirmation: the SDK's legacy shim reads capabilities declared at `initialize`, which per-request serving never sees
- Replaced Express with plain `node:http`, keeping the SDK's Host and Origin guards and re-adding the `100kb` body cap `express.json()` used to provide
- Rebuilt confirm-before-execute on the multi-round-trip flow: `execute_query` returns an `input_required` result and the client re-invokes the tool with the answer, since servers can no longer initiate requests
- Refused clients that do not declare `elicitation`, rather than executing unconfirmed, which was the last route to running a query nobody approved
- Bound the approval to the query it approved, by sealing an HMAC-signed digest of the built query into `requestState`, which travels through the client and returns as untrusted input
- Published `ttlMs` and `cacheScope` on the six cacheable results, and read `serverInfo.version` from the package manifest now that the revision stamps it onto every result
- Deleted `http/app.ts` and `utils/inMemoryEventStore.ts`, both of which existed only to manage sessions the revision removed

### Integration Tests

- Pinned the test client to `2026-07-28`, which no SDK client negotiates by default, so the suite would otherwise have covered the wrong wire era silently
- Replaced the `ping` liveness probe with `server/discover`

### Documentation

- Added [`.dev/docs/mcp-sdk-v2-changes.md`](https://github.com/overture-stack/arranger/blob/d321d8982e9ab8cd1770dbba7c08b5c1a14d00d8/.dev/docs/mcp-sdk-v2-changes.md), a short read on what changed and why, and [`mcp-sdk-v2-upgrade-plan.md`](https://github.com/overture-stack/arranger/blob/d321d8982e9ab8cd1770dbba7c08b5c1a14d00d8/.dev/docs/mcp-sdk-v2-upgrade-plan.md), the implementation record behind each decision
- Updated `apps/mcp-server/README.md` and the root `CHANGELOG.md`, and set `"protocolEra": "modern"` in `mcp-inspector.json`, without which the MCP Inspector negotiates 2025-era and is refused

### Special Instructions
Before running these changes, you will need to install the latest dependencies and rebuild your local modules:

```sh
npm ci                 # dependencies changed: the v1 SDK and Express are gone
npm run modules:build  # rebuild workspace modules
```

You will also need to update your environment variables, including generating an `MCP_REQUEST_STATE_SECRET`:

```sh
# from root
npm run mcp-server:generate-secret
```

Any MCP client used to test this branch must pin protocol revision `2026-07-28`; the endpoint refuses anything else.

### New Environment Variables

- **`MCP_ALLOWED_HOSTS`**: hostnames clients use to reach the server, matched against `Host` for DNS rebinding protection. Required whenever `MCP_HOST` is not loopback, which the `0.0.0.0` default is not, or the server exits at startup.
- `MCP_ALLOWED_ORIGINS`: browser origins allowed to call the server.
- `MCP_MAX_BODY_BYTES`: largest request body accepted, default `102_400`.
- **`MCP_REQUEST_STATE_SECRET`**: signs query confirmations. Required when `MCP_HOST` is not loopback.

For full details refer to the environment variable table in [`apps/mcp-server/README.md`](https://github.com/overture-stack/arranger/blob/d321d8982e9ab8cd1770dbba7c08b5c1a14d00d8/apps/mcp-server/README.md#environment-variables) and [`apps/mcp-server/.env.schema`](https://github.com/overture-stack/arranger/blob/d321d8982e9ab8cd1770dbba7c08b5c1a14d00d8/apps/mcp-server/.env.schema).

## Readiness Checklist

- [x] **Self Review**
- [x] **Formatting**
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] **Documentation**


apps/mcp-server/src/http/server.ts
```diff
@@ -0,0 +1,127 @@
...
+		void (async () => {
+			try {
+				// Before anything reads the body: a refused request should never cost us the payload.
+				for (const guard of guards) {
+					if (!guard(req, res)) {
+						return;
+					}
+				}
+
+				const { pathname } = new URL(req.url ?? '/', `http://${host}`);
```
justincorrigible, line 88: This throws for an unbracketed IPv6 host. 
`MCP_HOST=::1` is documented in the README as a supported loopback value, but `new URL('/x', 'http://::1')` throws `Invalid URL` (IPv6 literals need brackets in a URL authority: `http://[::1]`). Every request lands in the outer catch and gets a 500, so the server can't serve a single request on this bind. None of the current tests catch it since they only bind to `127.0.0.1`.

Suggested fix: bracket the host when it's a bare IPv6 literal before building the base URL, e.g. `` `http://${host.includes(':') && !host.startsWith('[') ? `[${host}]` : host}` ``, or reuse whatever bracketing helper the SDK may already expose for this (if one exists 🤞).

apps/mcp-server/src/http/requestBody.ts
```diff
@@ -0,0 +1,86 @@
...
+		}
+		chunks.push(buffer);
+	}
+
+	if (received === 0) {
+		return {};
+	}
+
+	try {
+		return { body: JSON.parse(Buffer.concat(chunks).toString('utf8')) };
```
justincorrigible, line 77: The old transport went through Express' `json()` middleware, which only parses bodies with `Content-Type: application/json` by its default filter; anything else was left as an empty body and rejected downstream. `readCappedJsonBody` here parses every request body as JSON regardless of Content-Type, so that gate is gone.

The `Origin` guard still covers CSRF for browser traffic when `MCP_ALLOWED_ORIGINS` is set, but a non-browser caller, or a deployment where an upstream proxy doesn't cleanly pass `Origin` through, loses a layer of protection that previously existed as a side effect of the old parser. 

Was dropping this deliberate, or worth adding back explicitly (i.e. checking `Content-Type` before attempting to parse)?

apps/mcp-server/src/mcp/executeQueryTool.ts
```diff
@@ -177,45 +220,106 @@ const validateRequest = ({
...
-	variables: Record<string, unknown>;
-}): Promise<boolean> => {
-	if (!server.server.getClientCapabilities()?.elicitation) {
-		return true;
-	}
+const clientCanElicit = (ctx: ServerContext): boolean => {
+	const envelope = ctx.mcpReq.envelope as Record<string, unknown> | undefined;
+	const capabilities = envelope?.[CLIENT_CAPABILITIES_META_KEY] as ClientCapabilities | undefined;
+	return capabilities?.elicitation !== undefined;
+};
```
justincorrigible, line 234: Two related things on this function:

1. **Behavioural change from the old transport.** The old `confirmExecution()` read `getClientCapabilities()`, negotiated once at `initialize` and cached for the session: a client declared elicitation support once, and one that lacked it got a documented graceful fallback (query runs unconfirmed but is echoed back). Looks like protocol revision `2026-07-28` has no session-level state to fall back on, so `clientCanElicit()` now requires the capability to be re-asserted in the `_meta` envelope of every `tools/call`. A client whose implementation only populates capabilities at `initialize` (a plausible habit carried over from the session-based era) now gets a silent, permanent refusal on every `execute_query` call instead of graceful degradation. Is this an accepted consequence of the protocol change, or worth a fallback/warning path for that class of client?

2. **Unchecked casts on peer-controlled input.** The `_meta` envelope is narrowed via `as Record<string, unknown>` and `as ClientCapabilities` rather than a Zod parse, even though this file validates every other piece of external input with a schema. A malformed envelope (e.g. `elicitation: 0` instead of an object) is currently read as "supports elicitation" by the `!== undefined` check purely because the cast tells TypeScript to trust the shape. Low blast radius today (the round trip just fails downstream), but worth the same schema treatment as everything else here so a shape mismatch surfaces as a validation error instead of failing silently later.

apps/mcp-server/src/mcp/requestState.ts
```diff
@@ -0,0 +1,87 @@
...
+const resolveKey = (secret: string | undefined): string | Uint8Array => {
+	if (secret !== undefined) {
+		return secret;
+	}
+	logger.warn(
+		'MCP_REQUEST_STATE_SECRET is not set: query confirmations are signed with a key generated for this ' +
+			'process. That is the intended default at a single replica. The key is not shared, so confirmations ' +
+			'issued before a restart stop being answerable, and every confirmation fails across multiple replicas. ' +
+			'Set MCP_REQUEST_STATE_SECRET when running more than one.',
+	);
```
justincorrigible, line 63: Read with concurrency and horizontal scalability in mind:
When `MCP_REQUEST_STATE_SECRET` is unset, confirmations are signed with a `randomBytes(32)` key generated fresh per process, with a warning logged (the message itself says "Set MCP_REQUEST_STATE_SECRET when running more than one"). Behind a load balancer with more than one replica, round one of a confirmation can be minted on replica A and routed to replica B for round two, which can't verify it and reports it as unbound. That's the default deployment shape here (multi-replica behind a load balancer), and there's currently no startup-time signal, unlike the equivalent `MCP_HOST` routable-without-`MCP_ALLOWED_HOSTS` case, which does hard-fail via `superRefine`. Worth the same treatment: fail startup (or at least fail loudly) when the host is non-loopback-only and the secret isn't set, rather than a log line an operator has to go looking for.

Two smaller things nearby, low priority: the `bind()` comment at line 83 says the method+clientId tag "today only stops state minted for one method being replayed against another," but `clientId` is always empty and `method` is always `'tools/call'` in the current call graph, so the bind check is currently a no-op rather than an active defence, plus the tag is built with a single `\0`-delimited concatenation (`${method}\0${clientId}`), which isn't collision-free once `clientId` becomes real, attacker-influenced text. Neither is exploitable today (nothing calls this with varying method/clientId yet), but worth tightening the comment's wording and, whenever a second bindable dimension actually shows up, using a length-prefixed or JSON-encoded bind instead of raw concatenation.

apps/mcp-server/src/server.ts
```diff
@@ -31,13 +43,13 @@ export const startServer = async (): Promise<void> => {
 	const client = createArrangerClient(config);
 	await validateArrangerConnection(config, client);
 
-	const deps: McpServerDeps = { config, client };
-	const { app, closeAllSessions } = createHttpApp(config, () => createMcpServer(deps));
+	const deps: McpServerDeps = { config, client, requestStateCodec: createConfirmationCodec(config) };
+	// One instance per request: the handler serves each request independently, so nothing is held
+	// between them and there is no session map to reap on shutdown.
+	const { close } = await startMcpHttpServer(config, () => createMcpServer(deps));
```
justincorrigible, line 49: The comment above this line explains why the server is built fresh per request, so my comment isn't asking whether that was intentional: it clearly was. However, the follow-on cost the comment doesn't mention: `_createRegisteredTool` eagerly runs `standardSchemaToJsonSchema` on every tool's input schema at registration time (confirmed directly in the SDK source), and `createMcpServer` reruns `registerResources`/`registerTools`/`registerPrompts` in full on every single incoming JSON-RPC request, not just tool calls, so that conversion (five tools including `build_sqon`'s multi-branch union, three resources, one prompt) is redone from scratch on every request.

Worth being precise about what the SDK does and doesn't already handle here, since it's clearly aware of this cost shape: `McpServer` memoizes each tool's converted schema in `_toolInputSchemaJson` specifically so the scan at registration time, and the pre-dispatch validation step share one conversion "instead of paying it twice per request under the per-request-factory `createMcpHandler` model" (the SDK's own words)... But that cache lives on the `McpServer` instance, and a fresh instance is built per request here, so we'll only remove a duplication within the same request, not the across-request one: every request still pays the full conversion cost once, whereas a session-scoped server would pay it once total.

Worth confirming this per-request cost was weighed against the lifecycle simplification, since the two are separable: a schema-conversion cache keyed by schema identity and shared across instances would keep the per-request server lifecycle this comment argues for while removing the repeated conversion work.

apps/mcp-server/src/mcp/executeQueryTool.ts
```diff
@@ -245,22 +349,37 @@ export const registerExecuteQueryTool = (server: McpServer, { client }: McpServe
...
+			if (!clientCanElicit(ctx)) {
+				return errorResult(
+					'execute_query requires a client that supports elicitation, because the generated query must be ' +
+						'confirmed before it runs, and this client did not declare the "elicitation" capability. ' +
+						'Reconnect with elicitation support, or use build_sqon to inspect the filter without executing it.',
+				);
+			}
+
 			try {
 				const serverIntrospection = serverIntrospectionSchema.parse(await client.getServerIntrospection());
```
justincorrigible, line 384: easy win: `client.getServerIntrospection()` and `client.getCatalogueIntrospection(catalogueId)` are awaited sequentially, but the second call only needs `catalogueId` from the tool's own input, not anything from the first call's result. `Promise.all`-ing them would save one Arranger round trip on a path the design doc already flags as costly, without changing the number of round trips overall.

apps/mcp-server/src/http/server.ts
```diff
@@ -0,0 +1,127 @@
...
+					return;
+				}
+
+				await serve(req, res, body);
+			} catch (error) {
+				logger.error({ err: error }, 'Unhandled error serving MCP request');
+				if (!res.headersSent) {
+					writeJsonRpcError(res, 500, -32603, 'Internal server error');
+				}
+				res.end();
```
justincorrigible, line 106: Nit (ignore if you wish): 
the catch block's `res.end()` runs unconditionally after `writeJsonRpcError` (which already ends the response) whenever `!res.headersSent`. Currently harmless (Node no-ops a second `end()`), but the trailing `res.end()` reads as though it's meant only for the `headersSent === true` branch and should probably be in an `else`.

apps/mcp-server/src/http/server.ts
```diff
@@ -0,0 +1,127 @@
...
+	await new Promise<void>((resolve, reject) => {
+		httpServer.once('error', reject);
+		httpServer.listen(port, host, () => {
+			httpServer.removeListener('error', reject);
+			resolve();
+		});
+	});
+
+	const close = async () => {
+		await handler.close();
```
justincorrigible, line 120: Nit (ignore if you wish):
`close()` awaits `handler.close()` before `httpServer.close()`. In the window between the two, a request can still land, pass the guards, and reach `handler.fetch()`, which throws "This MCP handler has been closed," answered as a generic 500 instead of a clean shutdown-style refusal. 

Reversing the order would avoid answering a pre-shutdown-eligible request with a fault response.

apps/mcp-server/src/http/server.ts
```diff
@@ -0,0 +1,127 @@
+import http, { type IncomingMessage, type Server, type ServerResponse } from 'node:http';
+
+import { hostHeaderValidation, originValidation, toNodeHandler } from '@modelcontextprotocol/node';
+import { createMcpHandler, type McpServerFactory } from '@modelcontextprotocol/server';
+
+import { readCappedJsonBody } from '#http/requestBody.js';
+import { type ArrangerMcpConfig } from '#utils/config.js';
+import logger from '#utils/logger.js';
```
justincorrigible, line 8: Nit (ignore if you wish):
this file imports the default `logger` directly, while the other new files in this PR (e.g. `utils/config.ts`, `mcp/requestState.ts`) use `createLogger('ModuleName')` for a prefixed logger 👍

Worth matching the convention this same PR establishes, for consistent log filtering (and because the prefixing does really help sometimes).

apps/mcp-server/src/utils/config.ts
```diff
@@ -4,6 +4,31 @@ import { createLogger } from '#utils/logger.js';
...
+const DEFAULT_MAX_BODY_BYTES = 102_400;
+
+/** Bind addresses that are only reachable from this host, so a Host allowlist is not required. */
+const LOCALHOST_HOSTNAMES = ['127.0.0.1', 'localhost', '::1'];
+
+/**
+ * Hostnames the SDK's own localhost guards allow. `[::1]` is bracketed because both guards compare
+ * against `new URL(...).hostname`, which brackets IPv6 literals.
+ */
+const LOCALHOST_ALLOWED_HOSTNAMES = ['localhost', '127.0.0.1', '[::1]'];
```
justincorrigible, line 20: `LOCALHOST_ALLOWED_HOSTNAMES` hand-rolls the same array `@modelcontextprotocol/server` already exports as `localhostAllowedHostnames()`/`localhostAllowedOrigins()`. Importing instead of reimplementing removes a value that would otherwise silently drift if the SDK's list ever changes.

apps/mcp-server/src/http/requestBody.ts
```diff
@@ -0,0 +1,86 @@
...
+ * "no pre-parsed body", and it is safe to pass after the stream has been drained: the adapter finds
+ * an exhausted stream and reads nothing.
+ */
+export type RequestBodyResult = { body?: unknown; refusal?: RequestBodyRefusal };
+
+/** JSON-RPC parse error, the answer for a body that is not valid JSON. */
+const PARSE_ERROR = -32700;
+
+/** Applied to a body over the configured ceiling. There is no JSON-RPC code for "too large". */
+const PAYLOAD_TOO_LARGE = -32600;
```
justincorrigible, line 21: `PARSE_ERROR` and `PAYLOAD_TOO_LARGE` reimplement the SDK's own `PARSE_ERROR`/`INVALID_REQUEST` JSON-RPC error codes under different names; importing them from `@modelcontextprotocol/server` (already imported elsewhere in this same request path) would be a trivial swap.

justincorrigible (changes requested):
Really strong first pass at the v2 migration, and the design doc's reasoning holds up well against the code. There is **just one thing that needs to be fixed before this can merge**, and a few behavioral changes from the old transport are worth a deliberate decision rather than a silent carry-over.

**Blocking:** `MCP_HOST=::1` 500s on every request (see inline comment on `http/server.ts`). This is a documented supported value, so it's a full outage on a supported configuration, not an edge case.

**Worth a decision before merge**, each noted inline:
- The Content-Type gate that used to come from Express's default JSON parser is gone, with nothing explicit replacing it.
- Elicitation capability now has to be re-declared on every `tools/call` rather than once per session, which is a real behavioural change for any client that only sends capabilities at `initialize`.
- `MCP_REQUEST_STATE_SECRET` being unset degrades to a warning rather than a startup failure, unlike the equivalent `MCP_HOST`/`MCP_ALLOWED_HOSTS` case.

Everything else inline is lower-priority: a few efficiency opportunities from the new per-request server model, some duplicated constants that could import from the SDK instead, and a couple of nits.

apps/mcp-server/src/http/server.ts
mistryrn: Great catch! Reproduced locally, fixing as part of my batch of PR feedback changes 🙏

apps/mcp-server/src/http/requestBody.ts
mistryrn: I hadn't considered this specifically, but it appears that `createMcpHandler`'s `handle()` rejects any POST whose Content-Type isn't `application/json` with a `415` response. I had Claude verify this after reviewing it: both `text/plain` and an absent Content-Type altogether returned `415 Unsupported Media Type`. So I think that gate is still covered fortunately 🙌

apps/mcp-server/src/mcp/executeQueryTool.ts
mistryrn: 1. Yes, this is an accepted consequence of the protocol change. Revision `2026-07-28` has no `initialize` handshake at all, so a client like the one described in your example (whose implementation only populates capabilities at `initialize`) would be a legacy client, refused at the door by the `legacy: 'reject'` setting we've applied to the MCP server.
 
The reason we're rejecting all legacy clients is that under the new spec and elicitation flow, `execute_query` would be unable to ask for confirmation on a legacy client. Since it's imperative that we ask for user confirmation before executing any queries, supporting legacy clients was a non-starter.

2. Looks like this specific example (malformed `elicitation: 0` instead of an object) would be impossible as the SDK parses the `envelope` against `ClientCapabilities2026Schema` before dispatch. I had Claude verify this claim, `elicitation: 0` returned an error `-32602 "expected object, received number" before reaching `clientCanElicit`. That said, we can still add the defensive parse to make it very clear 🫡  I'll add that change to my list for this round of feedback.

apps/mcp-server/src/utils/config.ts
mistryrn: Good catch! Will add to my batch of PR feedback changes 🫡

apps/mcp-server/src/http/requestBody.ts
mistryrn: Another great catch! Will update ASAP 🙏

apps/mcp-server/src/mcp/executeQueryTool.ts
mistryrn: Good call! Will update this to parallelize the requests.

apps/mcp-server/src/http/server.ts
mistryrn: Oooh, good call. Updated to wrap this in an `else {}` clause!

apps/mcp-server/src/http/server.ts
mistryrn: Valid (low priority) concern, but it looks like reversing the order causes its own unintended consequence: `httpServer.close()` doesn't drop existing keep-alive connections, it waits for them to drain, and MCP clients hold those open. So reversing the order likely delays shutdown behind a 30s hard timeout. Given this consequence, I'm leaning towards ignoring this one for now.

apps/mcp-server/src/http/server.ts
mistryrn: Good catch, that was a miss on my part! Correcting in this round of feedback 🙂

apps/mcp-server/src/server.ts
mistryrn: Good catch. `_createRegisteredTool` does convert eagerly, and the `_toolInputSchemaJson` memo is per-instance, so we do pay the full conversion cost once per request.

I had Claude take some measurements after reviewing this bit of feedback. The whole factory (5 tools, 3 resources, 1 prompt, all schema conversion) costs **0.28ms per request**, stable across repeated runs (on my machine, at least). A single `tools/list` call is already a 1ms request, and every other method makes at least one Elasticsearch-backed Arranger call that takes more time. Note that these timings would be different based on hardware and load, but at the moment it does not seem to be causing a significant impact in performance.

The reason I would push back on adding a cross-instance schema cache is that it feels like the cost to build it outweighs the performance cost of leaving the per-request conversion. Adding such a cache is a lot of overhead. The only clean public seam is `fromJsonSchema(schema, validator?)`, which takes a JSON schema plus a **JSON-Schema validator** and offers no way to keep a Zod validate function (the SDK ships `ajv` providers for this validation instead). Adopting it would replace our hand-written Zod messages with `ajv` output, and those messages are how the model corrects itself. Keeping Zod validation **and** caching the conversion means writing into `server._toolInputSchemaJson`, which the SDK marks `@internal`, so any patch release could silently break tool listing.

Maybe this is worth revisiting if we find that the server struggles under load?

mistryrn (commented):
Thanks for the feedback, Justin! I forgot to submit these responses before working on the fixes, so I've attached a summary of the responses/outcomes below 🙂 

## Summary of responses

| # | Feedback | Outcome |
|---|---|---|
| 1 | **Blocking:** `MCP_HOST=::1` 500s on every request | Fixed, with a regression test that binds `::1` |
| 2 | `requestBody.ts` reimplements the SDK's JSON-RPC codes | Done |
| 3 | Content-Type gate lost with Express | No change, see reply: the SDK answers `415` first |
| 4 | `http/server.ts` uses the unprefixed logger | Done |
| 5 | `res.end()` runs unconditionally in the catch | Done |
| 6 | `close()` order lets a request hit a closed handler | No change, see reply |
| 7 | Elicitation re-declared per `tools/call` | No change, see reply |
| 8 | Unchecked casts on peer-controlled input | Done, capabilities are now schema-parsed |
| 9 | Introspection calls awaited sequentially | Done, now concurrent |
| 10 | `MCP_REQUEST_STATE_SECRET` only warns | Done, required on a routable bind |
| 11 | `bind()` comment and `\0` concatenation | ~No change: `verify` is server-wide so method separation is live, and `\0` cannot appear in a method name. Worth revisiting when auth makes `clientId` real~ **UPDATE Sept 23:** Correct, the seam verifies only on `tools/call`, so the tag separates nothing until auth supplies a principal. Comment corrected. The `\0` join is unchanged, with the instruction to encode the parts left at that line for when auth lands |
| 12 | `LOCALHOST_ALLOWED_HOSTNAMES` duplicates an SDK export | Done |

### Beyond the review

- `.env.schema` omitted `[::1]`, so `::1` still failed at the Host guard after the fix for item 1
- A wrong path and an oversized body now answer `-32000` instead of `-32601` and `-32600`, matching what the SDK returns for its own pre-dispatch refusals here
- Added `setup-env` and `generate-secret`, since making the secret mandatory means `.env.schema` can no longer be copied and run

apps/mcp-server/src/mcp/executeQueryTool.ts
```diff
@@ -177,45 +215,96 @@ const validateRequest = ({
...
-	server: McpServer;
-	catalogueId: string;
-	endpoint: string;
-	query: string;
-	variables: Record<string, unknown>;
-}): Promise<boolean> => {
-	if (!server.server.getClientCapabilities()?.elicitation) {
-		return true;
-	}
+const clientCanElicit = (ctx: ServerContext): boolean => {
```
justincorrigible: still requires the capability on every `tools/call` rather than once per session, since revision `2026-07-28` has no session state to fall back on. Not fixable without a client-side change, just worth knowing: **a client that only sends capabilities at `initialize` gets a silent, permanent refusal instead of graceful degradation.**

apps/mcp-server/src/mcp/requestState.ts
```diff
@@ -0,0 +1,88 @@
...
+ * every confirmation fail.
+ */
+export const createConfirmationCodec = (config: ArrangerMcpConfig): RequestStateCodec<ConfirmationState> =>
+	createRequestStateCodec<ConfirmationState>({
+		key: resolveKey(config.mcp.requestStateSecret),
+		ttlSeconds: CONFIRMATION_TTL_SECONDS,
+		// The SDK's documented binding. With authentication out of scope the principal is always
+		// empty, so today this only stops state minted for one method being replayed against another;
+		// it starts separating principals the moment auth lands, with no change needed here.
+		bind: (ctx: ServerContext) => `${ctx.mcpReq.method}\0${ctx.http?.authInfo?.clientId ?? ''}`,
```
justincorrigible: the comment above this line says the method+clientId tag "only stops state minted for one method being replayed against another," but neither varies yet (`clientId` is always empty, `method` is always `tools/call`), so **the check is currently a no-op**... the `\0`-delimited concatenation also isn't collision-free once `clientId` becomes real text. Neither matters today, but it may become an issue in the future.

`#ToDo` 🤭

apps/mcp-server/src/http/server.ts
```diff
@@ -0,0 +1,158 @@
...
+import { createMcpHandler, INTERNAL_ERROR, type McpServerFactory } from '@modelcontextprotocol/server';
+
+import { TRANSPORT_REJECTION } from '#http/errorCodes.js';
+import { readCappedJsonBody } from '#http/requestBody.js';
+import { type ArrangerMcpConfig } from '#utils/config.js';
+import { createLogger } from '#utils/logger.js';
+
+const logger = createLogger('HttpServer');
+
+export type McpHttpServer = {
```
justincorrigible: this is the one newly exported type in this PR with no top-level TSDoc.

`#ToDo`? 🤷

justincorrigible (approved):
### This resolves everything that mattered. 
The `MCP_HOST=::1` outage is fixed and verified, and you caught a related bug in the same fix 💪  (the startup log printing an "undialable" address) that wasn't even flagged. The three declined items all hold up: `createMcpHandler` does 415 the Content-Type case before this app's own code runs, the per-request schema-conversion cost is now a measured 0.28ms against a real trade-off (i.e. losing the Zod-based error messages 👎), and reversing the shutdown order really would trade a narrow fault-response window for blocking on connection drain...

### A few small items left over, none blocking, just documenting for posterity:

